PR #53, opened 2026-06-25
fix(cli): build summary lists the bundled tools, matching dev/start

kid7st:
## Why

A full clean-room E2E (install `@kid7st/fastagent@0.5.0` from npm → `init` → add a custom tool → `dev` → `build` → copy artifact → `npm ci` → `start`, driving a real Anthropic model that called the tool) validated the whole flow. One DX nit surfaced: `fastagent build` printed `model/agents/skills` but **not** `tools:`, while `dev` and `start` both do — so after `build` an author couldn't confirm at a glance that their code tools shipped.

## Change

`build` deliberately does **not** import tools (their deps may not be installed at build time), so it can't reuse `loadTools`' executed names. Added `listToolFiles(dir)` — a pure filesystem listing of `tools/` (names from filenames, the authoritative source) with **no import**. `runBuild` scans the bundled artifact and prints `tools:` when present.

`listToolFiles` is an internal build-summary helper, not public API, so it is imported from its module (not re-exported from `index.ts`).

## Test

Locks that a tool file which would **throw on import** is still listed (proving the listing does not execute tool code), `.d.ts`/non-tool extensions are excluded, and missing `tools/` returns none. Full suite green; smoke: `fastagent build` now prints `tools:  word-count`.

> The other E2E observation (port override) was **not** a bug — `--port` / `PORT` and the precedence are already documented in `fastagent --help`.

### DIFF
--- a/core/src/engines/pi/tool.ts
+++ b/core/src/engines/pi/tool.ts
@@ -146,16 +146,23 @@ export async function loadTools(dir: string): Promise<{ tools: AgentTool[]; coll
  * deps may not be installed at build time). Missing `tools/` returns none.
  */
 export async function listToolFiles(dir: string): Promise<string[]> {
+  const toolsDir = join(dir, "tools");
   let entries: Dirent[];
   try {
-    entries = await readdir(join(dir, "tools"), { withFileTypes: true });
-  } catch {
-    return [];
+    entries = await readdir(toolsDir, { withFileTypes: true });
+  } catch (error) {
+    // Mirror loadTools: a missing tools/ is normal (none); any other failure is real — surface it
+    // rather than swallowing an unreadable dir into a false "no tools".
+    if ((error as NodeJS.ErrnoException).code === "not_found" || (error as NodeJS.ErrnoException).code === "ENOENT") {
+      return [];
+    }
+    throw new Error(`cannot read ${toolsDir}: ${(error as Error).message}`);
   }
-  return entries
+  // Dedup by basename, mirroring loadTools' "names that actually load" (foo.ts + foo.js → one foo).
+  const names = entries
     .filter((e) => e.isFile() && TOOL_EXTS.has(extname(e.name)) && !e.name.endsWith(".d.ts"))
-    .map((e) => basename(e.name, extname(e.name)))
-    .sort((a, b) => a.localeCompare(b));
+    .map((e) => basename(e.name, extname(e.name)));
+  return [...new Set(names)].sort((a, b) => a.localeCompare(b));
 }
 
 /**
--- a/core/src/engines/pi/tool.ts
+++ b/core/src/engines/pi/tool.ts
@@ -141,6 +141,24 @@ export async function loadTools(dir: string): Promise<{ tools: AgentTool[]; coll
 }
 
 /**
+ * List the code-tool names in `<dir>/tools/` WITHOUT importing them — the tool name is the filename
+ * (authoritative), so the build summary can report what it bundled without executing tool code (whose
+ * deps may not be installed at build time). Missing `tools/` returns none.
+ */
+export async function listToolFiles(dir: string): Promise<string[]> {
+  let entries: Dirent[];
+  try {
+    entries = await readdir(join(dir, "tools"), { withFileTypes: true });
+  } catch {
+    return [];
+  }
+  return entries
+    .filter((e) => e.isFile() && TOOL_EXTS.has(extname(e.name)) && !e.name.endsWith(".d.ts"))
+    .map((e) => basename(e.name, extname(e.name)))
+    .sort((a, b) => a.localeCompare(b));
+}
+
+/**
  * Merge already-resolved tools (pi defaults + `config.tools`) with discovered `tools/` tools,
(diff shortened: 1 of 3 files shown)


core/src/engines/pi/tool.ts
```diff
@@ -140,6 +140,24 @@ export async function loadTools(dir: string): Promise<{ tools: AgentTool[]; coll
...
+/**
+ * List the code-tool names in `<dir>/tools/` WITHOUT importing them — the tool name is the filename
+ * (authoritative), so the build summary can report what it bundled without executing tool code (whose
+ * deps may not be installed at build time). Missing `tools/` returns none.
+ */
+export async function listToolFiles(dir: string): Promise<string[]> {
+  let entries: Dirent[];
+  try {
+    entries = await readdir(join(dir, "tools"), { withFileTypes: true });
+  } catch {
```
kid7st: 宽 catch 吞掉所有 readdir 失败，把"不可读"误报成"无工具"。`catch { return [] }` 不区分 ENOENT（tools/ 不存在，正常）与其它错误（如 EACCES：目录存在但不可读）。同文件的 `loadTools` 明确区分：ENOENT 返回空，其它 `throw "cannot read …"`。这里一律吞成 `[]` → build 摘要静默不打印 `tools:` 行 → 作者得到"没有工具"的假信号。build 刚写完 outDir，紧接着读 outDir/tools 失败本身就是异常，正撞 AGENTS.md "fail visibly, no silent fallback"。

建议：照 loadTools 区分 ENOENT（静默 `[]`）与其它错误；后者用 `console.error` 警告（不 throw —— 此时 build 已成功，为一行摘要而让整个 build 失败是错的责任分配）。

core/src/engines/pi/tool.ts
```diff
@@ -140,6 +140,24 @@ export async function loadTools(dir: string): Promise<{ tools: AgentTool[]; coll
...
+    return [];
+  }
+  return entries
+    .filter((e) => e.isFile() && TOOL_EXTS.has(extname(e.name)) && !e.name.endsWith(".d.ts"))
+    .map((e) => basename(e.name, extname(e.name)))
```
kid7st: 不按 basename 去重，与 loadTools 在它唯一特殊处理的点上分叉。本函数目标是"对齐 dev/start"，但 `loadTools` 对同名文件（`foo.ts` + `foo.js`）按 basename 去重并记一条 `ToolCollision`，dev/start 据此只列一个 `foo` 且额外打印 collision 警告。这里把两者都 map 成 `"foo"` 不去重 → build 打印 `tools:  foo, foo`，与 dev/start 的输出和实际加载结果都不符；该 collision 是 loadTools 专门建模过的领域情况。此分叉无测试覆盖。

建议：map 后 `[...new Set(...)]` 去重 basename，镜像 loadTools 的"实际会加载的名字"；加一条 `foo.ts`+`foo.js` 测试锁住。

kid7st (commented):
Thesis: add `listToolFiles(dir)` — a pure filesystem listing of `tools/` (names from filenames, no import) — so `fastagent build`'s summary can print a `tools:` line matching dev/start without importing tool modules (whose deps may be absent at build time). The concept is honest (it lists files, not validated tools) and well placed (next to `loadTools`). Two findings, both where it diverges from `loadTools` — inline.

core/src/engines/pi/tool.ts
kid7st: Fixed. listToolFiles now mirrors loadTools: ENOENT/not_found → [] (no tools/ is normal); any other readdir error → throw 'cannot read <toolsDir>: …'. runBuild catches that and prints '[fastagent] warn: …' without failing the build — the build already succeeded, so a summary line must not sink it, but the failure is now visible instead of swallowed into a false 'no tools'. (Kept the throw in the data helper + warn in the cli layer rather than console.error inside listToolFiles, so it stays a pure sibling of loadTools and the warn lives where the other [fastagent] warn: lines are.) Test: tools/ as a FILE → readdir ENOTDIR → listToolFiles throws; mutation-verified (swallow non-ENOENT → red).

core/src/engines/pi/tool.ts
kid7st: Fixed. map → [...new Set(names)], so foo.ts + foo.js collapse to a single 'foo' — the name that actually loads, matching dev/start. Test added (foo.ts + foo.js → ['foo']); mutation-verified (drop the Set → 'foo','foo' → red).

kid7st (commented):
Re-review of 7971f16. Both prior findings resolved, no new findings.

- Error distinction — `listToolFiles` now mirrors `loadTools`: ENOENT/not_found → `[]`, any other error → `throw 'cannot read …'`; `runBuild` catches that and warns visibly without failing the build (the build already succeeded — correct that a summary line cannot sink it). A `tools/`-as-a-file (ENOTDIR) failure-path test locks the no-silent-`[]` behavior.
- Dedup — `[...new Set(names)]`, with a `foo.ts` + `foo.js` → `['foo']` test.

Both mutation-verified per the commit message. The `runBuild` warn-not-fail wrapper itself is not unit-tested (cli.ts terminal behavior, integration-level — consistent with the accepted boundary), not blocking.

PR #54, opened 2026-06-25
test: deflake CLI e2e subprocess tests (timeout + explicit forks pool)

kid7st:
## Why

The CLI e2e tests spawn `node src/cli.ts <cmd>` subprocesses that each **cold-start the full engine import graph** (~0.7s+, measured — `dist` and `src` are identical, so it's the pi packages, not TS stripping). Under vitest's default file-parallelism a many-core machine fans out enough of these at once to saturate CPU (load → 30+) and blow the **5s default per-test timeout**. It looks like 'random machine load,' but it's deterministic: heavy subprocess tests × unbounded parallelism × a tight timeout. Ironically CI passes *because* its runners have fewer cores → fewer parallel spawns.

Repro: `npm test` intermittently failed ~13 subprocess tests (cli/init/start) with ~5000ms timeouts under load; `--no-file-parallelism` or `--test-timeout=20000` made it green — pinpointing both the parallelism and the timeout.

## Fix

New `core/vitest.config.ts`, two stacked guards:
- `poolOptions.forks.maxForks: 4` — bounds concurrent heavy spawns (the load-spike root cause). A no-op on a CI runner that already has ≤4 cores; on an 8+ core dev machine it halves the fan-out.
- `testTimeout / hookTimeout: 20000` — absorbs a cold-start that's still CPU-starved.

## Verification

Full suite deterministic green (216/216, ~17.5s) with the config. The measured per-spawn cost and the parallelism mechanism are documented in the config comment for the next contributor.

### DIFF
--- a/core/vitest.config.ts
+++ b/core/vitest.config.ts
@@ -1,0 +1,22 @@
+import { defineConfig } from "vitest/config";
+
+// The CLI e2e tests spawn `node src/cli.ts <cmd>` subprocesses that each cold-start the full engine
+// import graph (~0.7s+, measured; the pi packages, not TS stripping). Under vitest's default
+// file-parallelism a contended machine can push a cold-start past the 5s default per-test timeout —
+// flaky under load, while a small CI runner squeaks by because it spawns fewer.
+//
+// The fix is the timeout, not a parallelism cap. A 20s ceiling absorbs a slow cold-start from any
+// cause at near-zero cost — it only bites when a test is genuinely slow. Capping vitest's own workers
+// (maxWorkers) was measured and rejected: it slows every run several-fold (17s -> ~60s here), is a
+// no-op on a CI runner with few cores, and can't touch the real culprit on a dev machine — external
+// CPU load (browsers, editors), which no test-runner setting controls.
+//
+// `pool: "forks"` is explicit, not left to the default: it is the process isolation these
+// subprocess / process.env tests assume.
+export default defineConfig({
+  test: {
+    pool: "forks",
+    testTimeout: 20000,
+    hookTimeout: 20000,
+  },
+});


core/vitest.config.ts
```diff
@@ -0,0 +1,17 @@
...
+// blow the 5s default per-test timeout — flaky under load, while a small CI runner squeaks by only
+// because it spawns fewer. Two stacked guards:
+//   - maxForks caps concurrent heavy spawns (bounds the load spike on big machines; a no-op on a CI
+//     runner that already has fewer cores than the cap).
+//   - a generous timeout absorbs a cold-start that is still CPU-starved.
+export default defineConfig({
+  test: {
+    testTimeout: 20000,
+    hookTimeout: 20000,
+    poolOptions: { forks: { maxForks: 4 } },
```
kid7st: 并发上限（PR 标称的 root-cause 修复）静默依赖未声明的默认 `pool: "forks"`。`poolOptions.forks.maxForks` 只在 pool 为 `forks` 时生效；vitest 4 当前默认正是 forks，所以现在能用 —— 但这条 root-cause 防线挂在一个隐式默认上：一旦 pool 变成 `threads`（未来默认变更，或有人 `vitest --pool=threads`），`poolOptions.forks` 整段被忽略，cap 静默失效，只剩 timeout band-aid，PR 的主修复机制就没了，且无任何信号。

建议：在 `test` 里显式加 `pool: "forks"`，让 cap 自洽、不依赖默认值，与注释 "maxForks caps concurrent heavy spawns" 的承诺对齐。

kid7st (commented):
Thesis: add `core/vitest.config.ts` with two stacked guards — `maxForks=4` (cap concurrent heavy subprocess spawns, the PR's stated root cause) + `testTimeout/hookTimeout=20000` (cushion a cold-start that's still CPU-starved) — to deflake the CLI e2e subprocess tests under CPU saturation. One finding, inline.

(Not a finding: the global `testTimeout=20000` applies the loose timeout to all 20 test files when only 4 spawn subprocesses, so a hang in a pure unit test now takes 20s to surface instead of 5s — but that cost is paid only on a (rare) hang, is documented in the config comment, and the global form fits the repo's simplicity-first stance. Naming the trade-off, not blocking it.)

core/vitest.config.ts
kid7st: Good catch — and chasing it surfaced a deeper bug: poolOptions.forks.maxForks is the vitest 3 API; vitest 4 (4.1.9 here) removed poolOptions, so the cap was being SILENTLY IGNORED from the start — the timeout was the only active guard, and the green run was that plus low load, not the cap. Your point (a root-cause defense hanging on an implicit default) was exactly right; it just bottomed out one layer further down.

When I fixed the API (top-level maxWorkers) so the cap actually applied, it measured as a net negative: 17s -> ~60s per run, a no-op on a few-core CI runner, and powerless against the real culprit on a dev machine (external CPU load — browsers/editors). So the cap is dropped. The config now keeps the timeout (the symptom-level fix that costs nothing when tests are fast) and makes pool: 'forks' explicit — both the isolation these subprocess/process.env tests assume and the self-contained declaration your comment asked for. No more dependence on an implicit default.

kid7st (commented):
Re-review of 6ba1eb1 (amended). Prior finding resolved, no new findings.

- The `pool: "forks"` silent-default dependency I flagged is now explicit — and it is load-bearing, not cosmetic: start/build/chat tests mutate `process.env.HOME` / `FASTAGENT_SESSIONS_DIR` / `PI_CODING_AGENT_DIR` with per-file save/restore, which only holds under per-file process isolation (forks); under `threads` those would race/leak across files. Good pin.
- The author went beyond the suggestion and re-derived the approach: dropped `maxForks` entirely with a measured rationale (caps every run 17s→~60s, no-op on a few-core CI runner, and cannot address the real culprit — external CPU load on a dev machine), leaving the 20s timeout as the symptom treatment. The trade-off is stated in the commit and the config comment.

Still-accepted, not a finding: the 20s `testTimeout`/`hookTimeout` is global, so the 16 non-subprocess test files also get the loose ceiling — a genuinely hung unit test now surfaces in 20s instead of 5s. Cost is paid only on a (rare) hang, documented, and fits simplicity-first.

PR #55, opened 2026-06-26
fix(config): a config syntax error names the file, not a raw SyntaxError

kid7st:
## Why

A manual DX validation pass (building several agents across different paths) found one rough edge in an otherwise excellent error surface: `loadConfig` wraps every config **shape** error with the file path —

```
/path/fastagent.config.mjs: must default-export defineConfig({...})
/path/fastagent.config.mjs: unknown key "htttp" (valid keys: model, tools, http)
```

— but the dynamic `import()` of the config file was unwrapped, so a **syntax** error surfaced as:

```
SyntaxError: Unexpected end of input
    at compileSourceTextModule (node:internal/modules/esm/utils:354:16)
    ...node ESM internal stack...
```

No filename, no `[fastagent]` framing — a stack-trace outlier exactly where a vibe/AI-coding author most needs a located, actionable message.

## Change

Wrap the config `import()` in try/catch and rethrow as ```<path>: <message>```, matching the shape errors. A syntax error now reads ``/path/fastagent.config.mjs: Unexpected end of input``.

## Test

`config.test.ts`: a truncated config (parse error) → the thrown message names the file. Full suite green (217).

---

### Note on scope (honest retraction)
The validation pass *initially* flagged a more serious 'dev orphans its worker on exit → port stuck in use' bug. On clean re-investigation that was **my own test-harness pollution** (a wrong `pkill` pattern leaked workers across runs); `fastagent dev` cleanup (SIGINT/SIGTERM) and watch-reload are correct — no orphan, no port leak. So this PR fixes only the real, small issue.

### DIFF
--- a/core/src/engines/pi/config.ts
+++ b/core/src/engines/pi/config.ts
@@ -69,7 +69,14 @@ export async function loadConfig(dir: string): Promise<LoadedConfig> {
   // Exactly one config file at this point (0 and >1 handled above).
   // biome-ignore lint/style/noNonNullAssertion: length checked above — exactly one element here
   const path = found[0]!;
-  const mod = (await import(pathToFileURL(path).href)) as { default?: unknown };
+  let mod: { default?: unknown };
+  try {
+    mod = (await import(pathToFileURL(path).href)) as { default?: unknown };
+  } catch (error) {
+    // A syntax/load error in the config would otherwise surface as a raw SyntaxError with a Node ESM
+    // internal stack; name the file and keep the message, matching the shape errors below.
+    throw new Error(`${path}: ${(error as Error).message}`);
+  }
   const config = mod.default;
   if (!config || typeof config !== "object") {
     throw new Error(`${path}: must default-export defineConfig({...})`);
--- a/core/src/engines/pi/config.ts
+++ b/core/src/engines/pi/config.ts
@@ -31,6 +31,7 @@ import { pathToFileURL } from "node:url";
 import type { AgentTool } from "@earendil-works/pi-agent-core";
 import type { Models } from "@earendil-works/pi-ai";
 import type { AnyModel } from "./harness.ts";
+import { moduleLoadHint } from "./loader.ts";
 
 export interface FastagentConfig {
   /** "provider/modelId", e.g. "openai-codex/gpt-5.5". Precedence: CLI --model > FASTAGENT_MODEL > config. */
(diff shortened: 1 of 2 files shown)


core/src/engines/pi/config.ts
```diff
@@ -69,7 +69,14 @@ export async function loadConfig(dir: string): Promise<LoadedConfig> {
...
   // biome-ignore lint/style/noNonNullAssertion: length checked above — exactly one element here
   const path = found[0]!;
-  const mod = (await import(pathToFileURL(path).href)) as { default?: unknown };
+  let mod: { default?: unknown };
+  try {
+    mod = (await import(pathToFileURL(path).href)) as { default?: unknown };
+  } catch (error) {
+    // A syntax/load error in the config would otherwise surface as a raw SyntaxError with a Node ESM
+    // internal stack; name the file and keep the message, matching the shape errors below.
+    throw new Error(`${path}: ${(error as Error).message}`);
```
kid7st: 新 catch 给 import 失败加了路径，却丢掉了项目既有的 `moduleLoadHint`，与 loadTools/loadChannels 不一致。`moduleLoadHint`（loader.ts）的存在目的就是覆盖 dynamic import 的两类常见 workspace 失败：依赖未装（`ERR_MODULE_NOT_FOUND` → "run `npm install`"）、非 ESM 包（`Unexpected token 'export'` → "set `type: module`"）。`loadTools`（tool.ts）和 `loadChannels`（channel.ts:74）都对 import 失败追加它；这里只贴了路径。而 config 的 import 会因完全相同的原因失败：用户写 `fastagent.config.js` 但 package.json 非 ESM → "Unexpected token 'export'"；config 里 import 了未安装依赖 → `ERR_MODULE_NOT_FOUND`。这两类失败在 tools/channels 处有可操作指引，config 处却没有，而 config 是作者最先加载的模块；commit 标题自己写的是 "syntax/**load** error"，load 失败本在范围内。

建议：rethrow 时追加 `moduleLoadHint(error as NodeJS.ErrnoException)`（并 import 它），与 loadTools/loadChannels 对齐：`` `${path}: ${(error as Error).message}${moduleLoadHint(error as NodeJS.ErrnoException)}` ``。

kid7st (commented):
Thesis: wrap the config `import()` in try/catch and rethrow as `<path>: <message>`, so a config syntax/load error is located + actionable like the shape errors below it, instead of a raw `SyntaxError` + Node ESM internal stack. Direction is right and matches the existing `<path>: …` style. One finding, inline.

core/src/engines/pi/config.ts
kid7st: Fixed. Appended moduleLoadHint(error) to the rethrow, so config now matches loadTools/loadChannels exactly: ```${path}: ${message}${moduleLoadHint(error)}```. The two workspace failure modes the hint covers both apply to config and now get guidance — a non-ESM package.json (`Unexpected token 'export'` → 'set type: module') and a dep imported by the config that isn't installed (ERR_MODULE_NOT_FOUND → 'run npm install'). A pure syntax error still yields no extra hint (message unchanged). Test added: a config importing a missing module → the error ends with the npm-install hint. Same theme as before — I mirrored the `<path>:` prefix but had missed the sibling's hint; now mirrored completely.

kid7st (commented):
Re-review of bf2e287. Prior finding resolved, no new findings.

- `loadConfig`'s import-failure rethrow now appends `moduleLoadHint(error)` (import added), aligning config with `loadTools`/`loadChannels` — a non-ESM package.json or a missing dep imported by the config now ends with the same `set type: module` / `run npm install` hint.
- A real failure-path test was added (config `import`s a missing package → asserts the file path + `npm install`, `/s` for the hint's newline), exercising the actual ERR_MODULE_NOT_FOUND path rather than a stub; the pure-syntax test still asserts path-only, so both branches are pinned. config suite green (20/20).

PR #52, opened 2026-06-25
chore(publish): publish @kid7st/fastagent to npm, not GitHub Packages

kid7st:
## Why

GitHub Packages gates every install behind a `read:packages` token + a per-user scope-registry `.npmrc` — friction the framework's core promise (existing definition → a service, fast) can't afford. The blocker to an open registry is gone: `@earendil-works/pi-*` already resolve from **public npm**, so a public `@kid7st/fastagent` installs with zero auth.

## Changes

| File | Change |
|---|---|
| `.github/workflows/publish.yml` | `registry-url` → `registry.npmjs.org`; `NODE_AUTH_TOKEN` → `secrets.NPM_TOKEN`; drop `packages:write`; rename job |
| `core/package.json` | `publishConfig` → `{ registry: registry.npmjs.org, access: public }` |
| `core/src/engines/pi/init.ts` | drop the scaffolded `.npmrc` (public npm needs none): removed the `NPMRC` template, scaffold entry, doc refs |
| `core/test/init.test.ts` | drop the `.npmrc` expectations |
| `README.md` / `docs/quickstart.md` | Install drops the GitHub Packages token dance; `npm i -g @kid7st/fastagent` directly |

## Prerequisites (maintainer, before the release)

1. The `@kid7st` **npm** scope owned by the publishing account.
2. An **`NPM_TOKEN`** (Automation) secret in the repo Actions secrets.
3. Drop `@kid7st:registry=https://npm.pkg.github.com` from your local `~/.npmrc`.

Then cut the `v0.5.0` release → publish.yml → first **npm** publish (0.5.0 is unpublished on any registry, so the first publish is clean).

## Verification

format / typecheck / lint clean; **214 tests**; build clean. Smoke: `fastagent init` no longer writes `.npmrc`. Only residual `npm.pkg.github` ref is a stale build artifact under `apps/github-reviewer/.fastagent/build/` (closed app, off the publish path) — flagged for separate cleanup.

PR #56, opened 2026-06-26
chore(release): 0.5.1 + publish via Trusted Publishing (OIDC)

kid7st:
Two release-enabling changes that ship together:

**1. Bump `core/package.json` to 0.5.1** — rolls up the DX fixes since 0.5.0 (all author-facing, no runtime/API change):
- #53 build summary lists the bundled `tools:`; `listToolFiles` mirrors `loadTools`'s edges.
- #54 vitest: longer timeout + explicit `pool: "forks"` (deflake CLI e2e under load).
- #55 config syntax/load errors name the file + carry the `npm install` / `type: module` hint.

**2. Switch `publish.yml` to npm Trusted Publishing (OIDC)** — no `NPM_TOKEN` secret, provenance attached automatically. Adds `id-token: write`, drops `NODE_AUTH_TOKEN`. Node 26 ships npm 11.12.1 (≥ 11.5.1), so no upgrade step.

## Before cutting the release (maintainer, one-time)
On npmjs.com → `@kid7st/fastagent` → Settings → **Trusted Publisher** → GitHub Actions:
- Organization or user: `kid7st` · Repository: `fastagent` · Workflow filename: `publish.yml`

Then `gh release create v0.5.1` triggers `publish.yml` → tokenless OIDC publish with provenance.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,15 +1,18 @@
 name: Publish
 
-# Publish @kid7st/fastagent to npm when a GitHub Release is published.
-# The release tag is the manual gate; the job re-verifies (typecheck + test) before
-# publishing so a non-green tag never ships. Version comes from core/package.json —
-# bump it before tagging (npm rejects re-publishing an existing version).
+# Publish @kid7st/fastagent to npm when a GitHub Release is published, via npm Trusted Publishing
+# (OIDC) — no NPM_TOKEN secret, and provenance is attached automatically. Requires a trusted publisher
+# configured on npmjs (org/user kid7st, repo fastagent, workflow publish.yml) and npm >= 11.5.1 (Node 26
+# ships 11.12.1). The release tag is the manual gate; the job re-verifies (typecheck + test) before
+# publishing so a non-green tag never ships. Version comes from core/package.json — bump it before
+# tagging (npm rejects re-publishing an existing version).
 on:
   release:
     types: [published]
 
 permissions:
   contents: read
+  id-token: write # OIDC token for Trusted Publishing
 
 jobs:
   publish:
@@ -41,6 +44,4 @@ jobs:
         run: npm test
 
       - name: Publish
-        run: npm publish
-        env:
-          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: npm publish # Trusted Publishing (OIDC) authenticates from the id-token; access/registry come from publishConfig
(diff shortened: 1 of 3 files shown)


kid7st (commented):
Thesis: ship 0.5.1 (rolling up #53/#54/#55) + switch publish.yml to npm Trusted Publishing (OIDC) — add `id-token: write`, drop `NODE_AUTH_TOKEN`, keep `registry-url`, provenance automatic. The OIDC wiring is correct (`id-token: write` present; token env removed; `registry-url` kept, which is the canonical setup-node + Trusted Publishing setup — an empty authToken makes npm fall back to OIDC; provenance auto). One finding, in body (the file is not in this PR's diff).

**`core/package-lock.json` — the version bump didn't bring the lock with it.** package.json is now 0.5.1 but the lock's root `version` is still `0.4.0` (three versions behind — it never tracked 0.5.0 either). This does NOT block the publish: `npm ci` checks the dependency tree, not the root version (proven by #52–#55 merging green with pkg 0.5.0 / lock 0.4.0), and package-lock.json is not in the published tarball, so the artifact version is correct. But it is the missing companion change to exactly what this release PR does, and a three-version drift signals the bump is hand-edited rather than via `npm version` (which updates both). If a future tool — or a stricter npm — starts enforcing root-version sync, this bites silently. Fix: regenerate the lock at 0.5.1; use `npm version patch` for future bumps so both stay in sync.

Not a finding: the first Trusted Publishing run is exercised only at release time and a failure lands after the tag/Release are created — but the one-time npmjs trusted-publisher config is documented in the PR's prereqs, and 0.5.1 is unpublished so a retry is clean. Documented and recoverable.

kid7st:
Fixed: synced `core/package-lock.json` root version 0.4.0 → 0.5.1 via `npm install --package-lock-only` — a clean 2-line diff (root `version` + `packages[""].version`), no dependency-tree change, `npm ci` consistent. Agreed it doesn't block the publish (npm ci checks the tree, lock isn't in the tarball) but it's the missing companion change. Will use `npm version patch` for future bumps so package.json + lock stay in sync.

kid7st (commented):
Re-review of 0b7785b. Prior finding resolved, no new findings.

`core/package-lock.json` root version is now 0.5.1 (both `version` and `packages[""].version`), regenerated via `npm install --package-lock-only` — a 2-line diff, no dependency-tree change, so the release is self-consistent. The commit records the root-cause fix (`npm version patch` to update both atomically on future bumps). The Trusted Publishing (OIDC) wiring was already correct. Nothing else outstanding.

PR #57, opened 2026-06-26
ci: add --provenance to npm publish

kid7st:
## Why

0.5.1 published cleanly via Trusted Publishing (tokenless OIDC) but shipped **without a provenance attestation**. npm's docs say provenance is automatic under trusted publishing, but npm 11.12.1 did **not** auto-attach it: the publish log had no provenance step and `dist.attestations` is empty on the registry (matches widely-reported behavior — the `--provenance` flag is what actually triggers it).

## Change

Add `--provenance` to the publish command. Everything the flag needs is already in place: `id-token: write`, a GitHub-hosted runner, and a public package (`publishConfig.access: public`). Future releases (0.5.2+) will carry the supply-chain provenance attestation — completing the OIDC trusted-publishing story.

No effect on 0.5.1 (already published); takes effect on the next release.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,7 +1,8 @@
 name: Publish
 
 # Publish @kid7st/fastagent to npm when a GitHub Release is published, via npm Trusted Publishing
-# (OIDC) — no NPM_TOKEN secret, and provenance is attached automatically. Requires a trusted publisher
+# (OIDC) — no NPM_TOKEN secret; provenance comes from the explicit --provenance flag below (npm did NOT
+# auto-attach it under OIDC), which needs the id-token: write permission. Requires a trusted publisher
 # configured on npmjs (org/user kid7st, repo fastagent, workflow publish.yml) and npm >= 11.5.1 (Node 26
 # ships 11.12.1). The release tag is the manual gate; the job re-verifies (typecheck + test) before
 # publishing so a non-green tag never ships. Version comes from core/package.json — bump it before
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -44,4 +44,7 @@ jobs:
         run: npm test
 
       - name: Publish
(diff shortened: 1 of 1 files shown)


kid7st (commented):
Thesis: add `--provenance` to `npm publish` because npm 11.12.1 under OIDC Trusted Publishing did NOT auto-attach provenance (verified: 0.5.1's `dist.attestations` empty), contradicting the docs and PR #56's "automatic" claim. The fix is correct — `id-token: write`, a GitHub-hosted runner, and a public package are all in place, which is everything `--provenance` needs; the PR also honestly records that 0.5.1 shipped without provenance and this takes effect on 0.5.2+. One finding, in body (the line is outside this PR's diff hunk).

**`.github/workflows/publish.yml:4` — the file's header comment still claims auto-provenance, directly contradicted by this PR's own fix.** Line 4 still reads "and provenance is attached automatically", while lines 48–50 added here say "npm 11.12.1 did NOT auto-attach it under OIDC (despite the docs)" and add `--provenance`. One file now carries two opposite claims — a future reader gets the wrong one from the header. This is the companion comment change the behavior fix should have brought along. Fix: rewrite line 4 to state provenance comes from the explicit `--provenance` flag (requires `id-token: write`), not automatically.

kid7st:
Fixed: rewrote the publish.yml header — it no longer claims 'provenance is attached automatically' (the stale #56 wording) and now states provenance comes from the explicit --provenance flag below, which needs id-token: write. The file's header and the publish step now agree; no contradicting claims.

kid7st (commented):
Re-review of c006bf1. Prior finding resolved, no new findings.

The `publish.yml` header no longer claims auto-provenance — it now states provenance comes from the explicit `--provenance` flag (which needs `id-token: write`), matching the publish-step comment and the flag itself. The file is self-consistent. Nothing else outstanding.

PR #58, opened 2026-06-26
redesign(login): persist via fastagent's own CredentialStore, not pi's AuthStorage

kid7st:
## The structural finding (from a post-iteration refactor scan)

`fastagent login` wrote credentials through pi-coding-agent's **`AuthStorage`**, while the runtime read/refreshed the **same** `~/.fastagent/auth.json` through fastagent's own **`fastagentCredentialStore`** (auth.ts). Two writer abstractions over one file, with **divergent semantics**: AuthStorage records-don't-throw; the store refuses-corrupt-**throws**. The entire #51 saga's residue — `drainErrors` plumbing, `assertPersisted`, `preflightAuthFile` — existed only to bolt fail-visibly onto AuthStorage's record-don't-throw persistence. The repeated #51 review bleeding was the symptom of this split.

## Why this direction (not preference — port shapes force it)

- The runtime resolves auth through **pi-ai's `CredentialStore` port** (`Models` consumes it), and pi ships **no file-backed `CredentialStore`** (only `InMemoryCredentialStore`). So `fastagentCredentialStore` is **mandatory** and already owns the file.
- `AuthStorage` is pi-coding-agent's **CLI-world** manager — a different port (no `modify(id, fn)` for OAuth-refresh-in-lock) the runtime cannot consume.
- ∴ "all pi's AuthStorage" is **impossible**; the only coherent unification is **login joining the engine-world store**.

## Change

- `login.ts`: OAuth flow via `pi-ai/oauth` `getOAuthProvider().login` (injectable `OAuthFlow` seam) + provider list via `getOAuthProviderInfoList()`; persist via `store.modify` (refuses to clobber a corrupt file → fails visibly on its own). A no-op `modify` up front is the preflight. **Deleted** the `AuthStore` interface, `drainErrors` plumbing, `assertPersisted`, `preflightAuthFile`, and the `AuthStorage` import — one writer, one semantics.
- `login.test.ts`: runs against the **real** `fastagentCredentialStore` over a temp file (not a mock of the binding); only the external OAuth flow is injected. Both routes, select, empty-selection/empty-key, corrupt-file preflight (asserts the flow did **not** run), manual-prompt abort.
- `auth.ts`: fix the `read` docstring that claimed "Read under the lock" — it is the deliberately **unlocked** hot path with resilient re-read (the L1 nit).

## Verify

typecheck / lint clean; **217 tests**; build clean; `fastagent login` lists the real providers. Net: deletes the #51 reconciliation machinery and ends the two-stores-one-file split.

### DIFF
--- a/core/test/login.test.ts
+++ b/core/test/login.test.ts
@@ -1,4 +1,4 @@
-import { mkdtemp, readFile, writeFile } from "node:fs/promises";
+import { chmod, mkdtemp, readFile, writeFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { getOAuthProviderInfoList } from "@earendil-works/pi-ai/oauth";
@@ -95,6 +95,29 @@ describe("loginFlow", () => {
     ).rejects.toThrow(/corrupt auth file/);
     expect(flowRan).toBe(false); // preflight threw before the flow — no wasted round-trip
   });
+
+  // The #51 invariant: the SECOND modify (the real in-place write, AFTER a successful flow) must fail
+  // visibly — never a false success. The corrupt-preflight test above only exercises the parse throw on
+  // the FIRST (no-op) modify; this one drives the write-IO path the preflight never reaches. A read-only
+  // (0444) file lets the preflight pass (read + lock, no write) and the flow run, then the persist write
+  // hits EACCES. (Skipped under root, which bypasses the permission check.)
+  it.skipIf(process.getuid?.() === 0)(
+    "a persist write failure after a successful OAuth flow rejects, persisting nothing",
+    async () => {
+      const path = await tmpAuth(JSON.stringify({ existing: { type: "api_key", key: "x" } })); // valid JSON
+      await chmod(path, 0o444); // read-only file; parent dir stays writable so the lock can still be taken
+      let flowRan = false;
+      const oauthFlow: OAuthFlow = async () => {
+        flowRan = true;
+        return fakeCreds;
+      };
+      await expect(
+        loginFlow(fakeIO(), { provider: "anthropic", store: fastagentCredentialStore(path), oauthFlow }),
+      ).rejects.toThrow(/EACCES|permission/i);
+      expect(flowRan).toBe(true); // got past preflight + flow; only the persist write failed
+      expect((await readAuth(path)).anthropic).toBeUndefined(); // no false success — nothing landed
+    },
+  );
 
   it("c1: a server win that leaves the manual-paste prompt pending is aborted, so the CLI does not hang", async () => {
     let manualAborted = false;
--- a/core/test/login.test.ts
+++ b/core/test/login.test.ts
@@ -1,115 +1,105 @@
-import { AuthStorage, InMemoryAuthStorageBackend } from "@earendil-works/pi-coding-agent";
-import type { AuthStorageBackend } from "@earendil-works/pi-coding-agent";
+import { mkdtemp, readFile, writeFile } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { getOAuthProviderInfoList } from "@earendil-works/pi-ai/oauth";
+import type { OAuthCredentials } from "@earendil-works/pi-ai";
 import { describe, expect, it } from "vitest";
-import type { AuthStore, LoginIO } from "../src/engines/pi/login.ts";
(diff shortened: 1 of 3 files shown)


core/test/login.test.ts
```diff
@@ -1,115 +1,105 @@
...
-        const { result, next } = fn(undefined);
-        if (next !== undefined) throw new Error("EROFS: read-only file system");
-        return result;
-      },
-      withLockAsync: async (fn) => {
-        const { result, next } = await fn(undefined);
-        if (next !== undefined) throw new Error("EROFS: read-only file system");
-        return result;
-      },
+  it("a corrupt auth file fails up front (preflight), before the OAuth round-trip", async () => {
```
kid7st, line 86: 测了 preflight 的 corrupt（第一个 `modify`，flow 之前），却没测 persist 时的 save 失败（第二个 `modify` 在 flow 成功之后抛）—— 而后者正是整个 #51 saga 存在的理由。

这个 corrupt-file 测试在 **preflight**（`modify(provider, () => undefined)`，不写）处抛，证明的是 parseForWrite 的 throw。但真正承载 "save 失败必须 fail-visibly、绝不假报成功" 的是**第二个** `modify`（`fn` 返回 credential、真正写盘）——它额外走了 preflight 不走的 write-IO 路径，且只有它能复现 "OAuth 跑完了、保存却失败"。redesign 删掉了旧套件里的 "write-throwing backend → throws" 测试，只用一个 preflight 测试顶替。后果：若有人日后把 persist 的 `modify` 包进 `.catch`（看似合理的 "别为一次保存抖动让 login 失败"），corrupt-preflight 测试依旧绿，而 #51 的原始 bug 静默复活、无人察觉。

建议：补一个 persist-时-失败的真实-store 测试 —— 预置一个合法 JSON 但 `chmod 0444` 的 auth 文件（父目录可写）：preflight 通过（只读+加锁不写），OAuth flow 跑完，第二个 `modify` 的 in-place 写触发 EACCES → loginFlow reject、断言无凭据落盘（无假成功）。in-place 写已由 auth.ts 注明，0444 文件能可靠触发该路径。

kid7st (commented):
Thesis: route `fastagent login`'s write through the runtime's own `fastagentCredentialStore` (`store.modify`) instead of pi-coding-agent's `AuthStorage` — one writer, one corruption/lock semantics — deleting the #51 residue (`drainErrors`/`assertPersisted`/`preflightAuthFile`/`AuthStore`). The direction is forced by port shapes (the runtime resolves auth only through pi-ai's `CredentialStore` port, pi ships no file-backed impl, so auth.ts already owns the file; `AuthStorage` is a CLI-world port the runtime can't consume) — a first-principles convergence, not preference. The redesign is clean: no dangling refs, OAuth behind an injectable `OAuthFlow` seam, the provider list and the flow both derive from `getOAuthProviders()` (no routing mismatch), the preflight reuses the store's refuse-corrupt via a no-op `modify`, and the `auth.ts` `read` docstring is corrected from the wrong "under the lock" to UNLOCKED (it previously contradicted the impl's own inline comment). One finding, inline.

core/test/login.test.ts
kid7st: Fixed — and you're right that this was the one invariant the redesign left unlocked. Added a real-store test driving the SECOND modify (the actual write, after the flow): a valid-JSON auth file chmod 0444 lets the preflight pass (read + lock, no write) and the injected OAuth flow run, then the in-place persist write hits EACCES → loginFlow rejects, the flow is asserted to have run (so we got past preflight to the real persist), and no credential lands (no false success). Skipped under root (bypasses the check). Mutation-verified: wrapping the persist modify in .catch(() => {}) — the exact #51 revival you described — turns only this test red.

kid7st (commented):
Re-review of 1af2df8. Prior finding resolved, no new findings.

The persist-time save failure is now locked by a real-store test: a valid-JSON `chmod 0444` file lets the preflight pass (read + lock, no write) and the injected OAuth flow run, then the SECOND modify's in-place write hits EACCES → `loginFlow` rejects, with both assertions that matter — the flow ran (got past preflight) and no credential landed (no false success). It is `skipIf` root (which bypasses the permission check) and mutation-verified (wrapping the persist modify in `.catch` turns exactly this test red). Confirmed it runs here (uid 501, not skipped): login suite 8/8 green. This closes the #51 invariant — the redesign now has the post-flow failure path it was built to guarantee.

PR #60, opened 2026-06-26
refactor(cli): extract the dev process supervisor to its own module

kid7st:
**Stacked on #59** (base is `refactor/shared-opener-tools`; retarget to main after #59 merges).

## Why

`runDevSupervisor` — the `fastagent dev` watch-and-restart mechanism (~75 lines: spawn the CLI as a worker, IPC ready signal, chokidar watch, debounced reload, SIGINT/SIGTERM shutdown) — lived inline in the 690-line `cli.ts` dispatch file. It is **process orchestration**, a distinct concern from command dispatch, and it closes over only the workspace `dir`.

## Change

Moved **verbatim** to `src/dev-supervisor.ts` as `runDevSupervisor(dir)`; `cli.ts` imports it and drops the now-unused `chokidar` import. The layer is unchanged (process side-effects still at the CLI tier, not the engine) — this just gives the dev process lifecycle its own readable module. **No behavior change** (`cli.ts` 690 → 606 lines).

## Verify

typecheck / lint clean; **218 tests**; build clean. Smoke: `fastagent dev` binds, an edit triggers `change detected → restart → re-bind`, and Ctrl-C cleanly kills the worker + frees the port.

### DIFF
--- a/core/src/cli.ts
+++ b/core/src/cli.ts
@@ -14,7 +14,6 @@
  * is the application entry point.
  */
 import { spawn } from "node:child_process";
-import { watch as watchTree } from "chokidar";
 import { existsSync } from "node:fs";
 import { isAbsolute, join, relative, resolve } from "node:path";
 import { createInterface } from "node:readline/promises";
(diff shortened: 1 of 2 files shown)


kid7st (commented):
No findings.

Scoped to this PR's delta (commit 592caa9 only — base is #59's branch, so the `resolveWorkspaceTools` commit belongs to PR #59). The dev-supervisor extraction is a behavior-preserving verbatim move: the single closed-over `dir` is promoted to a parameter, the module depends only on `node:child_process` + `chokidar` (no engine import — engine-neutral, CLI-tier, unchanged layer), the `chokidar`/`watchTree` import is fully dropped from cli.ts (no dangling reference), `dir` is in scope at the `runDevSupervisor(dir)` call site, and `runDevSupervisor` is not re-exported from `index.ts` (correctly internal). typecheck clean. The module boundary (process orchestration vs command dispatch) is honest.

PR #64, opened 2026-06-26
refactor: the directory is the agent — remove build/artifact and global skills

kid7st:
**Milestone 1 of the runtime-model redesign** (deploy is milestone 2). Net deletion: **-1869 / +115**.

## Why

A design pass concluded fastagent's `build` step was solving a problem agents don't have. Unlike Next.js (`next build` is mandatory — it compiles JS bundles), an **agent has no compile output**: instructions/skills are markdown, code tools are imported directly in dev and start alike. So `build` was *packaging*, not compiling — and it forced the artifact dir (`.fastagent/build`, **ephemeral**, replaced wholesale on rebuild) to double as the agent's **cwd**, a real footgun (tools writing to a dir the next build wipes).

Separately, **global skills** (`--global-skills`, `~/.pi/agent/skills`) let machine state leak into agent behavior — the exact "works on my machine, breaks deployed" trap fastagent exists to kill, and code you can't see in the directory. Both violate *the directory is the agent*.

## What changed

| Removed | Now |
|---|---|
| `build.ts`, `runBuild`, manifest, `bundleAgentDefinition` + ship helpers, `listToolFiles`, 6 build/artifact exports | — |
| `start` ran a built artifact (manifest model, sessions outside) | `start` runs the **definition directory** directly — the SAME opener `dev` drives (single assembly source, no drift); model/http from `fastagent.config.ts` |
| cwd = ephemeral artifact dir | cwd = the definition dir (**persistent, writable**; self-modification coherent) |
| global skills (`--global-skills`, `defaultGlobalSkillPaths`, the `skillPaths` mount) | skills come **only** from the definition's own `skills/` — dev == deployed, no exceptions |

Sessions default `<dir>/.fastagent/sessions`, overridable via `--sessions-dir` / `FASTAGENT_SESSIONS_DIR` to a volume (`start` reminds the operator on the default). Packaging (exclude/freeze) returns later as an **internal step of `deploy`** (milestone 2: multi-target + hosted), never a user-facing build.

## Public surface

`index.ts` drops `buildPiArtifact`, `ArtifactManifest`, `BuildPiArtifactOptions`, `createPiAgentFromArtifact`, `loadManifest`, `CreatePiAgentFromArtifactOptions`, `defaultGlobalSkillPaths`. `createPiAgentFromWorkspace` gains a `sessionsDir` option (start's deploy override). **Review-tier change** (public API + product semantics).

## Verify

[...]

PR #65, opened 2026-06-26
feat: fastagent add skill <source> — vendor an Agent Skills skill

kid7st:
## Why

Reusing a skill = **vendoring** (copy into `skills/`), the way the cross-tool [Agent Skills](https://agentskills.io) standard is meant to be consumed (verified compatible in #64). "Adding a skill" is still an unfamiliar operation, so this command doubles as **discoverability** — and its no-arg usage teaches **both paths, vibe first**:

```
add a skill — two ways:
  1. write your own (vibe): create skills/<name>/SKILL.md (auto-discovered). No command — common path.
  2. vendor an existing Agent Skills skill: fastagent add skill <source> [dir]
```

## Sources (all explicit, registry-neutral)

The filesystem is the registry, git is the distribution, the user owns trust — no proprietary marketplace binding.

| source | example | mechanism |
|---|---|---|
| **git ref** | `anthropics/skills/skills/pdf` | giget tar API (github default; `gitlab:`/`bitbucket:` kept) |
| **local path** | `./x`, `../x`, `/abs` | copy |
| **bare name** | `pdf` | found in local global dirs (`~/.agents/skills`, `~/.pi/agent/skills`) |

The **bare-name search is an add-time COPY** (git-tracked result), **NOT** the removed runtime global-skill scan: nothing is loaded from those dirs at run time; the definition stays self-contained. This is the explicit-vendor successor to the implicit global-skills mount #64 removed.

## Behavior

- copy-in, **validated with the same runtime loader** (a non-spec skill surfaces immediately)
- **refuses to overwrite** an existing skill (side-effect-free check first)
- **warns when a skill ships `scripts/`** (executable code that runs in your agent — review before deploy)
- **skill only, not tool**: skills are a cross-framework markdown standard (vendor-and-go); tools are framework-specific `defineTool` + carry npm deps, so an external tool wouldn't drop in

## Dependency

Adds `giget` (unjs, ^3.3.0) — download git repos/subdirs via tar API, no git binary.

## Verify

typecheck/lint/build green. Tests: local-path + global-dir vendor (copy / validate / scripts flag / refuse-overwrite), no-SKILL.md reject (no half-vendor), bare-name-not-found guidance. **Real-network smoke** confirmed all three sources vendor correctly.

> Note: this command's necessity is a DX judgment (discoverability + teaching), not a hard requirement. It is deliberately scoped (skill only, registry-neutral, vibe-first usage) so it informs without coercing.

### DIFF
--- a/core/src/engines/pi/init.ts
+++ b/core/src/engines/pi/init.ts
@@ -27,9 +27,9 @@
  * It does NOT defend against every pathological pre-existing target state (TOCTOU, read-only
  * dirs, FIFOs, mid-write disk-full, …): a local scaffolding command recovered by delete-and-retry.
  */
-import { access, cp, lstat, mkdir, readFile, readdir, rm, writeFile } from "node:fs/promises";
+import { access, cp, lstat, mkdir, readFile, readdir, rename, rm, writeFile } from "node:fs/promises";
 import { existsSync } from "node:fs";
-import { basename, dirname, isAbsolute, join, resolve } from "node:path";
+import { basename, dirname, isAbsolute, join, relative, resolve } from "node:path";
 import { homedir } from "node:os";
 import { type LoadedDefinition, loadAgentDefinition, loadRootIgnore } from "./definition.ts";
 import { fastagentVersion } from "./version.ts";
@@ -320,8 +320,9 @@ export interface VendoredSkill {
  * Copy-in, not link: the skill becomes a git-tracked part of the definition ("your folder is the
  * agent" — there is no registry; the filesystem is the registry, git is the distribution, the user
  * owns trust). Refuses to overwrite an existing skill UNLESS `options.update` — and then it is a plain
- * git-tracked overwrite (review/rollback via your repo's git diff/checkout), never a merge. Validates
- * the result with the SAME loader the runtime uses, so a non-spec skill surfaces immediately.
+ * git-tracked overwrite (review/rollback via your repo's git diff/checkout), never a merge. Fetches
+ * into a staging dir and validates BEFORE replacing, so a failed/invalid fetch never destroys an
+ * existing skill. Validation uses the SAME loader the runtime uses, so a non-spec skill surfaces now.
  */
 export async function vendorSkill(
   workspaceDir: string,
@@ -332,7 +333,8 @@ export async function vendorSkill(
   if (name === "" || name === "." || name === "..") {
     throw new Error(`cannot derive a skill name from "${source}" — point at a skill directory (…/skills/<name>)`);
   }
-  const dest = join(workspaceDir, "skills", name);
+  const skillsDir = join(workspaceDir, "skills");
+  const dest = join(skillsDir, name);
   // Refuse to clobber unless --update; the check is side-effect-free, and a git-tracked overwrite is
   // safe (review with `git diff`, undo with `git checkout`).
   const overwritten = existsSync(dest);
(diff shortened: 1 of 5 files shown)


core/src/engines/pi/init.ts
```diff
@@ -267,6 +270,108 @@ export async function assertChannelReady(dir: string): Promise<void> {
...
+  // Validate with the runtime loader: name/description + any spec diagnostics for THIS skill.
+  const def = await loadAgentDefinition(workspaceDir);
+  const skill =
+    def.skills.find((s) => s.filePath.includes(`skills/${name}/`)) ?? def.skills.find((s) => s.name === name);
+  return {
+    name: skill?.name ?? name,
+    description: skill?.description,
+    dest: `skills/${name}`,
+    hasScripts: existsSync(join(dest, "scripts")),
+    diagnostics: def.diagnostics.filter((d) => d.path?.includes(`skills/${name}`)),
```
kid7st: post-vendor 的 lookup/diagnostics 用了松散的正斜杠子串 `includes`，既跨同名前缀误attribute，又在 Windows 上静默丢弃全部。

- 本行 `def.diagnostics.filter((d) => d.path?.includes(\`skills/${name}\`))` **没有结尾分隔符**，于是前缀匹配：vendor `pdf` 会把同级 `pdf-tools` 的 diagnostics 也吸进来（`skills/pdf` ⊂ `skills/pdf-tools`）—— 在**所有**平台都误attribute。
- 本行与 365 行（`s.filePath.includes(\`skills/${name}/\`)`）都硬编码 `/`。loader 的 `filePath`/`d.path` 由绝对的 `workspaceDir` 拼出，Windows 上是反斜杠 → 两个 `includes` 都不匹配 → 报告静默丢掉该 skill 的 **description 与 diagnostics**，使 docstring 承诺的 "validated with the SAME runtime loader, so a non-spec skill surfaces immediately" 在 Windows 上落空（项目有 win32 分支，Windows 是考虑内的平台）。

依据：概念声称"立即 surface 非 spec skill"，实现只在 POSIX 且无同名前缀时成立；本该 warn 的 diagnostic 被静默吞掉。

建议：用精确目录比较替代子串 `includes` —— 例如 `relative(workspaceDir, dirname(s.filePath)) === join("skills", name)`，diagnostics 按各自的 skill 目录精确匹配；既消除前缀污染又跨平台正确。

kid7st (commented):
Thesis: `fastagent add skill <source>` vendors an Agent Skills skill (copy into `skills/<name>/`) from a git ref (giget, github default) / local path / bare name (local global dirs), validated by the runtime loader, refusing overwrite, warning on `scripts/`. Registry-neutral; the explicit-vendor successor to the runtime global-skill mount #64 removed (add-time copy → git-tracked, not a runtime scan). Concept clean, source routing correct, failure-path tests solid (no-SKILL reject + no half-vendor, bare-name-not-found guidance, refuse-overwrite). One finding, inline.

(Not flagged: a failed `giget` download can leave a partial `skills/<name>/` that blocks retry with an "already exists" message — but init.ts's header pre-declares mid-write failures as out of scope, "recovered by delete-and-retry," and the message points there. Noting, not re-litigating.)

core/src/engines/pi/init.ts
```diff
@@ -267,6 +269,124 @@ export async function assertChannelReady(dir: string): Promise<void> {
...
+  // Refuse to clobber unless --update; the check is side-effect-free, and a git-tracked overwrite is
+  // safe (review with `git diff`, undo with `git checkout`).
+  const overwritten = existsSync(dest);
+  if (overwritten && !options.update) {
+    throw new Error(
+      `skills/${name} already exists — re-run with --update to overwrite it (git tracks the change), or remove it`,
+    );
+  }
+  await mkdir(join(workspaceDir, "skills"), { recursive: true });
+  if (overwritten) await rm(dest, { recursive: true, force: true }); // --update: replace wholesale, then re-fetch
```
kid7st: `--update` 先删后取，failed re-fetch 会摧毁一个本来好用的 skill —— 这是数据丢失，不只是 half-vendor。

顺序是：本行 `if (overwritten) await rm(dest)`（345）→ 然后 `cp`/`downloadTemplate`（370）→ 再做 `SKILL.md` 有效性检查（373）。对 git-ref 源，370 行一次瞬时网络失败就会 throw，而旧 skill 已被删、新的没落地 —— `skills/<name>/` 没了。re-fetch 出来不是 skill 时同理：373 行的清理 `rm` 叠加在已删除的原件上。`--update` 的语义是"换成新的有效版本"，实现却 destructive-first —— 一次常规网络抖动让你两手空空。git-tracked 安全网（`git checkout`）只能救已 commit 的 skill，且 cli 提示把 `git checkout` 框定为"undo the update"，不是"找回丢失的 skill"。新增的 --update 测试只跑了 local-cp 的 happy path（v1→v2），从未覆盖 failed-refetch 丢失。

依据：概念是"换成新的有效版本"，实现先删后取，把可恢复的失败变成不可恢复的数据丢失。

建议：fetch+validate 到 staging 目录（temp），确认 `SKILL.md` 通过后再 `rm` 旧的并把新的 move 到位 —— 失败时旧 skill 原样保留。

kid7st (commented):
Re-review at e73f162 (--update flag + lazy-load giget). One new finding (inline) + a prior finding still open.

Lazy `import("giget")` is clean (matches the existing chat.ts lazy-load; the serve path no longer pulls giget) — no finding there.

**Still open from the prior review (https://github.com/kid7st/fastagent/pull/65#pullrequestreview-4582232688), now at init.ts:379/385, untouched by this commit:** the post-vendor skill lookup (379) and diagnostics filter (385) still use loose forward-slash substring `includes`. Line 385 has no trailing separator, so `pdf` prefix-pollutes a sibling `pdf-tools` skill's diagnostics on ALL platforms; and both hardcode `/`, so on Windows (loader paths are backslash) neither matches → the report silently drops the skill's description and diagnostics, defeating "validated … so a non-spec skill surfaces immediately." Fix: compare the skill directory precisely (`relative(workspaceDir, dirname(s.filePath)) === join("skills", name)`), not a substring.

core/src/engines/pi/init.ts
kid7st: Fixed in 1d4312b. Replaced both loose substring `includes` with an exact directory compare: `relative(workspaceDir, dirname(s.filePath)) === join("skills", name)` for the lookup, and the same on `dirname(d.path)` for diagnostics. Kills the prefix pollution (`skills/pdf` no longer matches `skills/pdf-tools`) on all platforms, and is cross-platform (no hardcoded `/`), so Windows no longer silently drops description/diagnostics. New mutation-verified test: a clean `pdf` vendored next to a `pdf-tools` carrying a name-not-equal-dir diagnostic asserts pdf diagnostics stay empty (revert to the loose prefix -> red). Thanks for flagging this twice — I missed it the first round.

core/src/engines/pi/init.ts
kid7st: Fixed in 1d4312b. `--update` no longer deletes first: it fetches into a staging dir (`skills/.<name>.vendoring`, same filesystem -> atomic rename), validates SKILL.md there, and only THEN `rm`s the old dest and `rename`s staging into place. Every failure path before the swap — a transient git-ref network blip, a source with no SKILL.md — leaves the old skill fully intact (the staging dir is cleaned in `catch`); a failed first-time vendor leaves no half-written skill either. New mutation-verified test: `--update` from an invalid (no-SKILL.md) source asserts the old skill is still v1 on disk and no staging dir leaks (revert to destructive-first -> red). Real-network smoke confirmed staging works for a git-ref update too.

kid7st (commented):
Re-review of 1d4312b. Both findings resolved, no new findings.

- `--update` data loss → now fetches into a staging dir (`.${name}.vendoring`), validates SKILL.md, and only then `rm`s the old dest + atomic `rename(staging, dest)`; a `try/catch` rm's staging on any failure so the existing skill survives a network blip / invalid source. The `--update`-from-an-invalid-source test proves v1 stays intact and no staging leftover remains — validate-before-replace, not destructive-first.
- substring `includes` → both the skill lookup and the diagnostics filter now use `relative(workspaceDir, dirname(...)) === join("skills", name)`: prefix-safe (the pdf/pdf-tools test proves pdf doesn't absorb pdf-tools' diagnostic) and cross-platform (separators normalized, no hardcoded `/`).

typecheck clean; init suite 36/36 green (two new failure-path tests). The earlier lazy-`import("giget")` was already clean. Nothing outstanding.

PR #66, opened 2026-06-26
refactor: drop vestigial symlinked-channels/ rejection + scrub stale build refs

kid7st:
A two-commit refactor from a post-iteration structural scan (after #64 removed build/global-skills and #65 added `add skill`). Structural pass found no major mislayering — the one structural finding (S2) + a surface cleanup (L1).

## Commit 1 — drop the vestigial symlinked-`channels/` rejection (S2)

`loadChannels` and `scaffoldChannel` rejected a symlinked `channels/` for **one** stated reason: *"the build skips symlinks, so a symlinked `channels/` serves in dev but is omitted from the artifact — dev and deployed diverge."* #64 removed the build/artifact, and dev/start now read the directory through the **same opener**, so a symlink can no longer make them diverge. The constraint the check enforced no longer exists → the check is removed; a symlinked `channels/` is simply followed (`readdir`/`mkdir` follow it), consistent with *the directory is the agent*.

- Kept: the unrelated scaffold-parent symlink guard (`init.ts`, a write-safety check against scaffolding *out of* the workspace).
- Tests rewritten "rejects" → "follows": `loadChannels` discovers a channel through a symlinked `channels/`; `add github` scaffolds through one.

## Commit 2 — scrub stale build/artifact/manifest references (L1)

#64 left 9 comments — and one **user-facing template** — referencing the removed build. Updated to the current model. The notable one: the `GITIGNORE` template scaffolded into every new agent said *"ship in the build artifact"* / *"dev sessions + build output"* — now phrased as secrets kept out of git and any deploy copy. Comments-and-template only; no behavior change.

## Not done (deliberately)

The scan's third candidate — splitting `init.ts` (507 lines: scaffold + channel-scaffold + skill-vendor) into per-concern modules — was **declined**: same category as the earlier dev-supervisor extraction (organizational, complexity-relocating not -reducing). Left for if/when the coupling actually bites.

## Verify

typecheck/lint green; 67 in-process tests across init/channel/config/definition pass; rewritten symlink tests lock the new follow behavior.

### DIFF
--- a/core/src/engines/pi/channel.ts
+++ b/core/src/engines/pi/channel.ts
@@ -7,7 +7,7 @@
  * may list inline).
  */
 import type { Dirent } from "node:fs";
-import { lstat, readdir } from "node:fs/promises";
+import { readdir } from "node:fs/promises";
 import { extname, join } from "node:path";
 import { pathToFileURL } from "node:url";
 import type { Agent } from "../../agent.ts";
@@ -36,27 +36,15 @@ export async function loadChannels(
   agent: Agent,
 ): Promise<{ routes: Routes; collisions: ChannelCollision[] }> {
   const channelsDir = join(dir, "channels");
-  // lstat (does not follow the link) first: the build skips symlinks (definition.ts), so a symlinked
-  // channels/ would serve in dev but be omitted from the artifact — dev and deployed would diverge.
-  // Reject it rather than silently follow it. (Symlinked FILES inside channels/ are skipped below by
-  // the isFile() check, matching the build, so they need no special case here.)
-  let stat: Awaited<ReturnType<typeof lstat>>;
-  try {
-    stat = await lstat(channelsDir);
-  } catch (error) {
-    const e = error as NodeJS.ErrnoException;
-    if (e.code === "ENOENT" || e.code === "not_found") return { routes: {}, collisions: [] };
-    throw new Error(`cannot read ${channelsDir}: ${(error as Error).message}`);
-  }
-  if (stat.isSymbolicLink()) {
-    throw new Error(
-      `${channelsDir} is a symlink — the build does not follow it, so dev and deployed would diverge; use a real directory`,
-    );
-  }
+  // A symlinked channels/ is simply followed: with no build/artifact, dev and start read the directory
+  // through the same opener, so a symlink cannot make them diverge. (Top-level non-file entries —
+  // including symlinks-to-files — are skipped by the isFile() filter below; only real channel files load.)
   let entries: Dirent[];
   try {
     entries = await readdir(channelsDir, { withFileTypes: true });
   } catch (error) {
+    const e = error as NodeJS.ErrnoException;
+    if (e.code === "ENOENT" || e.code === "not_found") return { routes: {}, collisions: [] }; // no channels/ is normal
     throw new Error(`cannot read ${channelsDir}: ${(error as Error).message}`);
   }
   const routes: Routes = {};
--- a/core/src/engines/pi/channel.ts
+++ b/core/src/engines/pi/channel.ts
@@ -12,6 +12,7 @@ import { extname, join } from "node:path";
 import { pathToFileURL } from "node:url";
 import type { Agent } from "../../agent.ts";
 import { parseRouteKey, type Routes } from "../../host/node.ts";
+import { assertInsideWorkspace } from "./definition.ts";
 import { moduleLoadHint } from "./loader.ts";
 
 /** A `channels/<name>.ts` default export: receives the assembled agent, returns the routes it mounts. */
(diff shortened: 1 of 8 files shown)


core/src/engines/pi/channel.ts
```diff
@@ -36,27 +36,15 @@ export async function loadChannels(
...
-    const e = error as NodeJS.ErrnoException;
-    if (e.code === "ENOENT" || e.code === "not_found") return { routes: {}, collisions: [] };
-    throw new Error(`cannot read ${channelsDir}: ${(error as Error).message}`);
-  }
-  if (stat.isSymbolicLink()) {
-    throw new Error(
-      `${channelsDir} is a symlink — the build does not follow it, so dev and deployed would diverge; use a real directory`,
-    );
-  }
+  // A symlinked channels/ is simply followed: with no build/artifact, dev and start read the directory
```
kid7st: 这条 rejection 的理由没有完全死亡 —— 它从"build skips symlinks"转移到了"deploy copies the directory"，而本次移除比那个死掉的理由所允许的更宽、更不一致。

- 它防的 divergence 在**当前已陈述的 deploy 模型**下复活：`cli.ts:221` 已把 deploy 表述为 "a deploy that copies the directory"。一个**逃出 workspace** 的 `channels/` symlink 不在这份拷贝里 → 部署后的 agent **没有 channels**，而本地 dev/start（同 opener、symlink 在场）能加载 —— 正是 dev/deployed divergence，只是"build"换成了"copy the directory"。
- 跟随一个逃逸的 `channels/` symlink 意味着 agent 的 channels 活在它的目录**之外** —— 这与移除所引用的"the directory is the agent"自相矛盾。这个检查本就是在**执行**该原则，而非 build 的副产品。
- 与 commit 2 自己的措辞不一致：commit 2 在 `scaffoldWorkspace` 里**保留**了 symlink 守卫，并把理由重述为 "a symlink here would write outside the workspace"（`init.ts:442`）—— 而 commit 1 却把 `channels/` 的同一守卫当作 build-specific 删掉。
- `scaffoldChannel` 的移除**无守卫**：保留的守卫在 `scaffoldWorkspace`（`init`），不在 `scaffoldChannel`（`add`）。于是 `add github`（`init.ts:228` 的 `mkdir`+`writeFile`）现在会**穿过**逃逸的 `channels/` symlink 把 `github.ts` 写到 workspace 之外，无任何检查。

依据："目录即 agent"与"跟随逃逸 symlink"自相矛盾；移除守卫的理由只部分死亡，deploy-copy 下 divergence 复活。

建议：narrow，别 drop —— 只拒绝 realpath **逃出 workspace** 的 `channels/` symlink（指向 dir 内部的 symlink 自包含、无害，旧检查对它过度拒绝了），并在 `loadChannels` 与 `scaffoldChannel` 两处都加这条收窄后的守卫，与你保留的 `scaffoldWorkspace` 守卫一致。

kid7st (commented):
Thesis: post-#64/#65 structural scan. (1) drop the symlinked-`channels/` rejection whose stated rationale ("the build skips symlinks → dev/deployed diverge") died when #64 removed the build; a symlinked `channels/` is now simply followed. (2) scrub 9 stale build/artifact/manifest comments + one user-facing `GITIGNORE` template. Commit 2 is accurate and behavior-free — no finding there. One finding on commit 1, inline.

core/src/engines/pi/channel.ts
kid7st: Fixed in c243b08 — you were right, dropping it went too wide. The rationale only partially died: it moved from "the build omits symlinks" to "a deploy that COPIES the dir omits an escaping symlink", and an escaping channels/ also contradicts "the directory is the agent" (the check was enforcing it, not a build artifact). And — embarrassingly — I dropped the scaffoldChannel write-outside guard in the same PR that KEPT the identical scaffoldWorkspace guard.

So: narrowed, not dropped. New `assertInsideWorkspace(workspaceDir, name)` (definition.ts) rejects a subdir whose realpath escapes the workspace, allows an in-workspace symlink (self-contained — the old check over-rejected it), applied in BOTH loadChannels (read) and scaffoldChannel (write), consistent with scaffoldWorkspace. Both ends realpath'd (so /tmp → /private/tmp isn't a false escape). Tests rewritten to two cases each (in-workspace → followed/scaffolded; escaping → rejected), mutation-verified by removing either guard. Smoke: `add github` through an escaping symlink is refused and writes nothing outside.

Third time you've caught a "change introduced a safety regression" (self-ignore over-reach, --update data loss, now this write-outside hole) — the pattern I need to internalize is to ask, when removing a guard, what invariant it was actually holding.

kid7st (commented):
Re-review of c243b08. Finding resolved, no new findings.

Narrowed, not dropped, exactly as suggested: `assertInsideWorkspace(workspaceDir, name)` (definition.ts) rejects only a symlink whose realpath ESCAPES the workspace, allows an in-workspace symlink, and no-ops a missing subdir. The realpath-on-both-ends correctly avoids a false escape on a symlinked workspace root (macOS /tmp→/private/tmp), and the ENOENT branch handles a not-yet-created channels/. It's applied in BOTH `loadChannels` and `scaffoldChannel` (shared helper), consistent with the kept `scaffoldWorkspace` guard. Tests rewritten for both paths with real fs symlinks: in-workspace symlink followed; escaping symlink rejected (`loadChannels` throws /outside the workspace/, and `add github` is rejected with github.ts asserted NOT written outside the workspace) — distinct, real failure-path coverage. typecheck clean; channel+init 46/46 green.

PR #87, opened 2026-06-29
feat(telegram): per-session serial queue (model B: one shared turn at a time)

kid7st:
In a shared group (model B: session = chat[:thread]), two people @-ing the bot at once
both resolved to the same session and the second collided on the engine's per-session
lease — dropped with a neutral "session busy, try again". Wrong UX for a collaborative
group.

The channel now serializes per session: a second update for the same session is queued
behind the first (FIFO) and runs when it finishes, rather than colliding. Different
sessions still run concurrently (the engine's stateless multi-session invoke), so the
queue never blocks unrelated chats. The engine lease stays as the corruption floor
beneath it. SPEC §8 explicitly leaves this concurrency policy to the channel.

In-process (a restart loses anything still queued — Telegram already got its 200); the
queue is a per-session promise chain whose map entry is dropped when it drains. The
draft id / lifecycle log / engine turn now run at DEQUEUE time, so they reflect the
actual execution order, not arrival.

Tests: same-session turns run one-at-a-time in FIFO order (maxInFlight === 1, the second
not dropped as busy); different sessions run concurrently (maxInFlight === 2).

### DIFF
--- a/core/src/channels/telegram.ts
+++ b/core/src/channels/telegram.ts
@@ -441,6 +441,20 @@ export function telegramChannel(
     );
   }
   const decide = route ?? ((update: TelegramUpdate) => defaultTelegramRoute(update, { botUsername: mentionName }));
+  // Per-session serial queue: model B answers a shared chat[:thread] with ONE turn at a time, so a
+  // second update for the same session waits its turn (FIFO) instead of colliding on the engine lease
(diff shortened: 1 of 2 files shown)


kid7st (commented):
No findings.

PR #88, opened 2026-06-29
feat(telegram): flag a group chat in the envelope (multi-person context)

kid7st:
In a shared group session (model B) the agent sees turns from several people, each
`from`-tagged. A 1:1 DM is self-evident, but a group is not — without a cue the model
may read consecutive turns as one continuous interlocutor. The envelope now adds, for a
group/supergroup, "[group chat — multiple people; each message is prefixed with its
sender]" so the agent addresses participants by name and knows it is not a 1:1. A
private chat gets no note.


kid7st (commented):
No findings.

PR #91, opened 2026-06-30
feat(log): leveled logging — close the observe.ts design gap

kid7st:
Every runtime line — lifecycle, warnings, errors, and the turn trace — went straight to
console.error at one implicit severity: no levels, no per-posture gating, no way to tell a
debug trace from a real error, and the trace (which logs end-user content) could only be
kept out of production by soldering it to the dev command path. observe.ts carried a long
DESIGN GAP comment marking exactly this.

Introduce one process logger (src/log.ts): debug/info/warn/error gated by a single level,
to stderr. It is a module singleton — not injected — because most runtime logs originate
inside author-constructed channels (channels/*.ts call the factory themselves) and deep
engine code the CLI cannot thread a logger into; the leaf sinks that ARE constructed by the
caller (auth, the turn trace) keep their injectable param, defaulting to the singleton.

- Posture sets the level: dev → debug (verbose, trace shown), start → info (trace gated).
- FASTAGENT_LOG_LEVEL is parsed to three states: a valid value locks the level (overrides
  posture); a present-but-invalid value warns at load and is treated as absent (a typo meant
  to make logs louder must never silently pin the level to info nor disable the posture
  default — fail visibly); absent leaves the posture default.
- The turn trace (observe.ts) becomes a DEBUG emitter wired in BOTH dev and start. Production
  runs at info, so it is gated out entirely — end-user content (prompt, tool args/results,
  reply) and its volume never reach production logs. That replaces the old dev-only switch
  and needs no separate redaction pass.
- Runtime console.* across telegram/github/http/invoke/tunnel/dev-supervisor/sessions/auth/
  report move to log.{info,warn,error} at the right level. The CLI's user-facing output
  (help, command results, fatal bind errors) stays on console — it is program UI, not logs.

Structured-JSON output and content redaction beyond level-gating are deferred (YAGNI).

Tests: createLogger level gating; the singleton + setLogLevel; and (via resetModules +
stubEnv) the env override — a valid value makes setLogLevel a no-op, an invalid value warns
and falls back to posture. auth tests inject the warn sink; invoke/sessions tests assert the
stderr the logger writes.

### DIFF
--- a/core/src/log.ts
+++ b/core/src/log.ts
@@ -1,0 +1,66 @@
+/**
+ * Leveled logging. Runtime logs (lifecycle, warnings, errors, the debug turn trace) flow through ONE
+ * process-level logger, gated by a single level. It is a module singleton — not an injected dependency —
+ * because most runtime logs originate inside author-constructed channels (`channels/*.ts` call the
+ * channel factory themselves) and deep engine code, which the CLI cannot thread a logger into. The CLI
+ * sets the level by posture (dev → debug, start → info); `FASTAGENT_LOG_LEVEL` overrides it.
+ *
+ * This is NOT the CLI's user-facing output (help text, command results): that is the program talking to
+ * its operator and stays on plain `console`. Everything here is operational logging to stderr.
+ */
+
+export type LogLevel = "debug" | "info" | "warn" | "error";
+
+export interface Logger {
+  debug(msg: string): void;
+  info(msg: string): void;
+  warn(msg: string): void;
+  error(msg: string): void;
+}
+
+const ORDER: Record<LogLevel, number> = { debug: 0, info: 1, warn: 2, error: 3 };
+const isLevel = (s: string): s is LogLevel => s in ORDER;
+const format = (level: LogLevel, msg: string): string => `${level.toUpperCase().padEnd(5)} ${msg}`;
+
+/** A standalone logger over an explicit sink — used in tests to assert level gating without the singleton. */
+export function createLogger(opts: { level: LogLevel; sink?: (line: string) => void }): Logger {
+  const sink = opts.sink ?? ((line) => console.error(line));
+  const make =
+    (level: LogLevel) =>
+    (msg: string): void => {
+      if (ORDER[level] >= ORDER[opts.level]) sink(format(level, msg));
+    };
+  return { debug: make("debug"), info: make("info"), warn: make("warn"), error: make("error") };
+}
+
+/**
+ * `FASTAGENT_LOG_LEVEL` parsed to three states: a valid value locks the level (overrides posture); a
+ * present-but-invalid value warns and is treated as absent, so a typo (meant to make logs louder) can
+ * never silently pin the level to info nor kill the posture default; absent returns undefined. The
+ * warning is raw — the singleton below is not built yet — but reuses `format` for the same shape.
+ */
+function parseEnvOverride(): LogLevel | undefined {
+  const raw = process.env.FASTAGENT_LOG_LEVEL;
+  if (raw === undefined) return undefined;
+  const value = raw.toLowerCase();
+  if (isLevel(value)) return value;
+  console.error(format("warn", `[fastagent] unknown FASTAGENT_LOG_LEVEL "${raw}"; using the posture default`));
+  return undefined;
+}
+
+const override = parseEnvOverride();
+let currentLevel: LogLevel = override ?? "info";
+
+/** Set the posture default. A valid `FASTAGENT_LOG_LEVEL` override, if present, wins and is not changed. */
+export function setLogLevel(level: LogLevel): void {
+  if (override === undefined) currentLevel = level;
+}
+
+const emit =
+  (level: LogLevel) =>
+  (msg: string): void => {
+    if (ORDER[level] >= ORDER[currentLevel]) console.error(format(level, msg));
+  };
+
+/** The process logger. Runtime code imports this and calls `log.info(...)` etc. */
+export const log: Logger = { debug: emit("debug"), info: emit("info"), warn: emit("warn"), error: emit("error") };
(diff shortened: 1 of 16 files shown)


kid7st (commented):
本次改动新增一个进程级 leveled logger（src/log.ts）：用单一 level 网关 debug/info/warn/error 到 stderr，替换散落的 console.error 单一隐含 severity；posture（dev→debug / start→info）经 setLogLevel 设定，FASTAGENT_LOG_LEVEL 覆盖；turn trace 降为 debug emitter，在 dev/start 都接线、靠级别决定可见性。它位于 CLI posture → log 单例 → 运行时各 emitter 的边界。

core/src/log.ts
```diff
@@ -0,0 +1,51 @@
...
+    };
+  return { debug: make("debug"), info: make("info"), warn: make("warn"), error: make("error") };
+}
+
+const envLevel = process.env.FASTAGENT_LOG_LEVEL?.toLowerCase();
+let currentLevel: LogLevel = envLevel && isLevel(envLevel) ? envLevel : "info";
+
+/** Set the posture default. The `FASTAGENT_LOG_LEVEL` override, if present, wins and is not overridden. */
+export function setLogLevel(level: LogLevel): void {
+  if (envLevel === undefined) currentLevel = level;
```
kid7st: 问题: 无效或空的 `FASTAGENT_LOG_LEVEL` 会静默把级别钉死在 info，并永久使 `setLogLevel` 失效。`envLevel` 只要 env 被设置就非 undefined（哪怕 `"trace"`、`"DEBUG "`、`""`）；init 时 `envLevel && isLevel(envLevel)` 为假 → currentLevel="info"，而 setLogLevel 守卫是 `envLevel === undefined` → 对任何“已设置”的 env 都 no-op。于是用户把 FASTAGENT_LOG_LEVEL 写错一个字（本想更啰嗦）反而得到更安静的 info，且 dev 的 setLogLevel("debug")＋turn trace 被静默吞掉，无任何诊断。这是 fail-visibly 违例 + 责任错放：日志级别是用户可改的输入，写错应让用户看见（warn + fallback），而不是吞掉坏值还顺带废掉 posture 默认。

建议方向: 把 env 解析为三态——present-and-valid → 锁定 override；present-and-invalid → log.warn/stderr 提示 “unknown FASTAGENT_LOG_LEVEL '<x>', using <posture default>” 并当作未设置；absent → 未设置。setLogLevel 守卫改判“有效 override 是否存在”（如 `override === undefined`），让无效值不再废掉 posture。

core/test/log.test.ts
```diff
@@ -0,0 +1,53 @@
...
+  it("at error level emits only errors", () => {
+    const { log, lines } = capture("error");
+    log.warn("w");
+    log.error("e");
+    expect(lines).toEqual(["ERROR e"]);
+  });
+});
+
+describe("log singleton + setLogLevel", () => {
+  // Assumes FASTAGENT_LOG_LEVEL is unset (an env override would win and skip setLogLevel by design).
```
kid7st, line 35: 问题: PR 描述把 “the singleton + setLogLevel (env-override-wins by design)” 列为已测，但本用例显式注释 “Assumes FASTAGENT_LOG_LEVEL is unset”，只验证 env 未设置时 setLogLevel 移动阈值。真正新增的承重语义——env 已设置时 override 胜出 / setLogLevel no-op，以及无效 env 的 fallback——一条都没覆盖。把 setLogLevel 的 env 守卫删成无条件赋值，现有测试仍全绿，回归不可见。

建议方向: 用 `vi.resetModules()` + `vi.stubEnv("FASTAGENT_LOG_LEVEL", …)` 重新 import log.ts，断言：(1) 有效 env override 下 setLogLevel("debug") 不改变级别；(2) 无效 env 下落回 posture（并发出 warn）。

core/src/log.ts
kid7st: Fixed in 3b1a5ac. FASTAGENT_LOG_LEVEL now parses to three states: a valid value locks the level (overrides posture); a present-but-invalid value (incl. empty) warns at load — `WARN [fastagent] unknown FASTAGENT_LOG_LEVEL "<x>"; using the posture default` (raw console.error reusing `format`, since the singleton is not built yet) — and is treated as absent; absent leaves the posture default. The setLogLevel guard now keys off `override === undefined`, so an invalid value no longer pins the level to info nor disables the posture default. Verified: `FASTAGENT_LOG_LEVEL=trace` warns at load and a subsequent setLogLevel("debug") still shows debug.

core/test/log.test.ts
kid7st: Fixed in 3b1a5ac. Two tests added (vi.resetModules + vi.stubEnv + dynamic re-import, so the load-time parse runs under the stubbed value): (1) a valid env override (`error`) makes setLogLevel("debug") a no-op — debug stays gated; (2) an invalid env (`trace`) warns at load (asserted on the captured stderr) and setLogLevel("debug") still emits debug. Deleting the guard / unconditional assignment now fails (1). The PR description was also corrected to stop claiming the env path was covered.

kid7st (commented):
No findings. 上一轮两条均已修复且正确：(1) parseEnvOverride() 把 FASTAGENT_LOG_LEVEL 三态化——valid 锁定 override / present-but-invalid 经 console.error 警告（不经级别网关、直达 stderr）并落回 posture / absent 返回 undefined，setLogLevel 守卫改判 override===undefined，写错的 env 不再静默钉死 info 也不再废掉 dev 的 debug trace；(2) 新增两条测试用 resetModules+stubEnv+fresh import 精准锁住 valid→setLogLevel no-op 与 invalid→警告+posture 仍生效，且在 import 前 spy console.error 捕获 load-time 警告。typecheck 通过、log.test 6 测试全绿。

PR #92, opened 2026-06-30
refactor(scaffold): templates as real files, not escaped TS strings

kid7st:
The files `init`/`add` scaffold lived as TS string literals in scaffold-templates.ts, so any
nested code had to escape its backticks/backslashes — `\\s+` in the word-count tool, every
`\`…\`` in the channel templates, and the whole telegram-send tool written with string
concatenation purely to dodge nested backticks. Template code was not real code: no
highlighting, no editor lint, escaping noise.

Move the content to real files under templates/ (AGENTS.md, house-style.md, tool.ts,
fastagent.config.mjs, gitignore, env.example, channels/{github,telegram}.ts,
tools/telegram-send.ts), read at scaffold time via `new URL("./templates/…", import.meta.url)`.
The exports keep their names (SKILL_MD, TOOL_TS, …) so init.ts/add-channel.ts are untouched;
only the parametric pieces (the AGENTS.md tool line, package.json, toPackageName) stay as code.

- templates/ is excluded from this package's tsc + biome (it imports the published
  @kid7st/fastagent, not this source) and copied into dist/ by the postbuild step; the
  package already ships dist, so npm pack includes it.
- Resolution works in both trees: src/ (vitest) and dist/ (the built CLI), since the loader
  resolves relative to its own module URL and the build colocates templates.
- telegram-send now reads as clean template literals instead of "a" + b concatenation — the
  generated tool is functionally identical, just legible. The de-escaped word-count regex is
  `/\s+/`, not `/\\s+/`.

Verified: init.test green; the built CLI scaffolds init + `add telegram` from dist with the
de-escaped content; npm pack lists all nine template files.

### DIFF
--- a/core/biome.json
+++ b/core/biome.json
@@ -1,7 +1,7 @@
 {
   "$schema": "https://biomejs.dev/schemas/2.5.0/schema.json",
   "files": {
-    "includes": ["src/**/*.ts", "test/**/*.ts"]
+    "includes": ["src/**/*.ts", "test/**/*.ts", "!src/engines/pi/scaffold/templates/**"]
   },
   "formatter": {
     "enabled": true,
--- a/core/package.json
+++ b/core/package.json
@@ -62,7 +62,7 @@
     "typecheck": "tsc --noEmit",
     "lint": "biome check src test",
     "format": "biome check --write src test",
-    "postbuild": "node -e \"require('node:fs').chmodSync('dist/cli.js', 0o755)\""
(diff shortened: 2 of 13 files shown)


kid7st:
Superseded: reorganizing channels into per-channel folders (src/channels/<x>/{runtime + scaffold/}) so a channel is the cohesion + future-package boundary, then reworking the scaffold on top. Reusing this branch's de-escaped template files in the follow-up.

PR #94, opened 2026-06-30
refactor(scaffold): relocate out of engines/pi to a neutral home

kid7st:
The scaffold writes SPEC-level workspace shape (AGENTS.md, skills/, tools/, channels/), not
pi internals — it does not belong under engines/pi. Move src/engines/pi/scaffold/ →
src/scaffold/ and untangle the coupling that kept it there:

- assertInsideWorkspace + loadRootIgnore were engine-neutral (pure fs/path + the `ignore`
  matcher) but lived in engines/pi/definition.ts and were shared by the engine's channel
  discovery AND the scaffold. Extract them to a neutral src/workspace.ts.
- fastagentVersion (the package's own version, read for the scaffolded dependency range) was
  neutral but under engines/pi; move it to src/version.ts.
- scaffold now depends on the engine only through vendor-skill's loadAgentDefinition — the one
  legitimate edge, since vendoring a skill validates it loads under the engine that will run it.

Pure relocation: imports repointed (channel.ts, cli.ts, index.ts, test), definition.ts loses
the two functions and their now-unused imports. typecheck/lint/248 tests green; the built CLI
init + --version work from dist.


kid7st (commented):
No findings. 纯重定位，逐项核验零行为变化：(1) assertInsideWorkspace/loadRootIgnore 逐字移到 src/workspace.ts，函数体完全一致；(2) version.ts 承重行 new URL(...package.json) 深度随移动正确调整——原 src/engines/pi 用 ../../../、新 src/ 用 ../，均解析到 core/package.json（已实测），这是 relocation 唯一易错点；(3) definition.ts 删函数后把随之不再用到的 import 裁到只剩 writeFile+join，无残留未用 import；(4) 全部 import 重指向正确、src/test 无对旧路径的残留引用，vendor-skill 保留唯一合法 engine 边 loadAgentDefinition；(5) index.ts 公共面仅改 scaffoldWorkspace import 路径，workspace.ts 为内部模块、无 surface 变化。typecheck 通过、init+channel 38 测试全绿。

PR #96, opened 2026-06-30
fix(telegram): live preview via sendMessage + editMessageText, not sendMessageDraft

kid7st:
sendMessageDraft (Bot API 9.3) streams a partial message TO A USER and is tied to forum-topic
mode — it is not a general group mechanism, so a normal group fails with TEXTDRAFT_PEER_INVALID.
The live preview was built entirely on it, so groups (the whole point of recent work) got NO
live preview, only the final reply. Switch to the standard, everywhere-working pattern.

- One preview message: sendMessage("💭 Thinking…") once (capturing its message_id, threaded under
  the asker), then editMessageText in place with the latest tool calls + partial text. The preview
  is PLAIN — a partial answer often carries unbalanced HTML that Telegram would reject.
- The final answer edits that message as HTML (plain fallback inside editMessageText); if it
  overflows one message the first chunk edits the preview and the rest are sent after it; if the
  edit fails (message gone) it falls back to a fresh send, so the answer always lands.
- The single-writer pump carries over (one write in flight keeps frames monotonic). No keepalive:
  a real message does not expire (unlike the 30s draft). The throttle is now interruptible —
  finish() cuts it so the final write is not delayed up to EDIT_THROTTLE_MS after completion.
- callBotApi now returns the result payload so sendMessage can hand back the message_id.

Drops sendMessageDraft + DRAFT_KEEPALIVE_MS; EDIT_THROTTLE_MS = 1500 (edits to one message are
rate-limited tighter than sends). Tests rewritten to the edit mechanism (placeholder → edits →
clean HTML final; not-modified skip; HTML→plain fallback; never-two-writes-in-flight; edit-fail
falls back to a send). 248 tests green.

This is the streaming half of the Telegram review (#1/#2/#5). Route fixes (#3 drop the group
slash-command summon, #4 entity-based @mention + strip) follow in a separate PR.

### DIFF
--- a/core/src/channels/telegram/telegram-api.ts
+++ b/core/src/channels/telegram/telegram-api.ts
@@ -44,14 +44,17 @@ async function callBotApi(
   botToken: string,
   method: string,
   params: Record<string, unknown>,
-): Promise<{ ok: true } | { ok: false; status: number; description: string }> {
+): Promise<{ ok: true; result?: { message_id?: number } } | { ok: false; status: number; description: string }> {
   for (let attempt = 0; ; attempt++) {
     const res = await fetch(`${api}/bot${botToken}/${method}`, {
       method: "POST",
       headers: { "content-type": "application/json" },
       body: JSON.stringify(params),
     });
-    if (res.ok) return { ok: true };
+    if (res.ok) {
+      const data = (await res.json().catch(() => ({}))) as { result?: { message_id?: number } };
+      return { ok: true, result: data.result };
+    }
     const data = (await res.json().catch(() => ({}))) as {
       description?: string;
       parameters?: { retry_after?: number };
@@ -66,7 +69,7 @@ async function callBotApi(
 }
 
 /** Split text into ≤4096-char chunks (Telegram's limit), preferring newline boundaries. */
-function chunkText(text: string): string[] {
+export function chunkText(text: string): string[] {
   if (text.length <= TELEGRAM_MAX_TEXT) return [text];
   const chunks: string[] = [];
   let rest = text;
@@ -81,12 +84,20 @@ function chunkText(text: string): string[] {
 }
 
 /**
- * Persist a final reply: split to Telegram's 4096-char limit, and send each chunk as HTML, falling back
- * to plain text only if Telegram rejects the markup — so a model formatting slip degrades instead of
- * losing the message. Format is the channel's job (Telegram's own parser); the developer never picks it.
- * `message_thread_id` is dropped from the JSON when undefined (linear chat).
+ * Send a message: split to Telegram's 4096-char limit, each chunk as HTML by default (`html:false` for
+ * a plain live-preview), falling back to plain only if Telegram rejects the markup — so a model
+ * formatting slip degrades instead of losing the message. Returns the FIRST chunk's message_id (so the
+ * caller can edit it as a live preview). `message_thread_id` is dropped from the JSON when undefined.
  */
-export async function sendMessage(api: string, botToken: string, t: Target, body: string): Promise<void> {
+export async function sendMessage(
+  api: string,
+  botToken: string,
+  t: Target,
+  body: string,
+  opts: { html?: boolean } = {},
+): Promise<number | undefined> {
+  const html = opts.html ?? true;
+  let firstId: number | undefined;
   let first = true;
   for (const chunk of chunkText(body)) {
     const base: Record<string, unknown> = { chat_id: t.chatId, message_thread_id: t.threadId, text: chunk };
@@ -96,31 +107,53 @@ export async function sendMessage(api: string, botToken: string, t: Target, body
     if (first && t.replyTo !== undefined) {
       base.reply_parameters = { message_id: t.replyTo, allow_sending_without_reply: true };
     }
-    first = false;
-    let result = await callBotApi(api, botToken, "sendMessage", { ...base, parse_mode: "HTML" });
-    if (!result.ok && PARSE_ERROR.test(result.description))
+    let result = html
+      ? await callBotApi(api, botToken, "sendMessage", { ...base, parse_mode: "HTML" })
+      : await callBotApi(api, botToken, "sendMessage", base);
+    if (html && !result.ok && PARSE_ERROR.test(result.description))
       result = await callBotApi(api, botToken, "sendMessage", base);
     if (!result.ok) throw new Error(`telegram sendMessage failed: ${result.status} ${result.description}`.trim());
+    if (first) firstId = result.result?.message_id;
+    first = false;
   }
+  return firstId;
 }
 
-/** Push an ephemeral draft (animated preview, capped at the text limit). Same draft_id animates updates. */
-export async function sendMessageDraft(
+/**
+ * Edit a message in place — the live-preview mechanism (one message, repeatedly updated). Plain by
+ * default (a partial preview may contain unbalanced HTML); `html:true` for the final answer, falling
+ * back to plain if Telegram rejects the markup. "message is not modified" is NOT an error: the pump may
+ * re-render an unchanged view. The message is identified by (chat_id, message_id) — no thread needed.
+ */
+export async function editMessageText(
   api: string,
   botToken: string,
   t: Target,
-  draftId: number,
+  messageId: number,
   body: string,
+  opts: { html?: boolean } = {},
 ): Promise<void> {
-  const result = await callBotApi(api, botToken, "sendMessageDraft", {
+  const base: Record<string, unknown> = {
     chat_id: t.chatId,
-    message_thread_id: t.threadId,
-    draft_id: draftId,
+    message_id: messageId,
     text: body.slice(0, TELEGRAM_MAX_TEXT),
-  });
-  if (!result.ok) throw new Error(`telegram sendMessageDraft failed: ${result.status} ${result.description}`.trim());
+  };
+  let result = opts.html
+    ? await callBotApi(api, botToken, "editMessageText", { ...base, parse_mode: "HTML" })
+    : await callBotApi(api, botToken, "editMessageText", base);
+  if (opts.html && !result.ok && PARSE_ERROR.test(result.description))
+    result = await callBotApi(api, botToken, "editMessageText", base);
+  if (!result.ok && /message is not modified/i.test(result.description)) return;
+  if (!result.ok) throw new Error(`telegram editMessageText failed: ${result.status} ${result.description}`.trim());
 }
 
+/** Delete a message — the wire primitive for removing a preview placeholder (the finalize POLICY lives
+ *  in telegram.ts). Throws on failure, like the other primitives; a caller that wants best-effort cleanup
+ *  catches explicitly rather than the primitive swallowing it. */
+export async function deleteMessage(api: string, botToken: string, t: Target, messageId: number): Promise<void> {
+  const result = await callBotApi(api, botToken, "deleteMessage", { chat_id: t.chatId, message_id: messageId });
+  if (!result.ok) throw new Error(`telegram deleteMessage failed: ${result.status} ${result.description}`.trim());
+}
 /** getMe: the bot's own @username (so default routing recognises group @mentions) and whether group
  *  privacy mode is OFF (`can_read_all_group_messages`) — needed to receive the un-summoned group
  *  messages that feed the context buffer. */
--- a/core/src/channels/telegram/telegram.ts
+++ b/core/src/channels/telegram/telegram.ts
@@ -5,8 +5,10 @@
  * is a JSON POST, outbound is a `fetch` to the Bot API. The developer writes only `route` (policy); the
  * channel owns transport + format + attachments.
  *
- * Live streaming: tool calls + partial text stream into an ephemeral draft (sendMessageDraft, a 30s
- * animated preview); the final text is persisted with sendMessage. Threaded Mode (topics in private
+ * Live streaming: a single "💭 Thinking…" message is sent once, then editMessageText'd in place with the
+ * latest tool calls + partial text (PLAIN — a partial answer may carry unbalanced HTML); on completion
+ * the same message is edited into the final answer as HTML. One message, works in groups and private
+ * (unlike sendMessageDraft, which is private/forum-topic only). Threaded Mode (topics in private
  * chats, a @BotFather toggle) is auto-adapted: an update carrying message_thread_id replies into that
  * thread; without one the chat is linear. Same code, both modes.
  *
@@ -14,7 +16,6 @@
  * kit: readBodyCapped / text), so it is exactly what a third-party `fastagent-channel-*` package would write.
  */
 import { timingSafeEqual } from "node:crypto";
-import { setTimeout as sleep } from "node:timers/promises";
 import type { Agent, AgentEvent, ImageRef, Json } from "../../agent.ts";
 import { log } from "../../log.ts";
 import { readBodyCapped } from "../body.ts";
@@ -24,8 +25,10 @@ import {
   resolveBotInfo,
   resolveFiles,
   resolveImages,
+  chunkText,
+  deleteMessage,
+  editMessageText,
   sendMessage,
-  sendMessageDraft,
 } from "./telegram-api.ts";
 import { text } from "../respond.ts";
 
@@ -43,11 +46,9 @@ interface BufferEntry {
   body: string;
 }
 
-/** How often (ms) to push a streamed draft update; tool events flush immediately for snappy feedback. */
-const DRAFT_THROTTLE_MS = 800;
-
-/** Re-push the draft at least this often, under Telegram's ~30s draft expiry, so a long step keeps the preview alive. */
-const DRAFT_KEEPALIVE_MS = 20_000;
+/** How often (ms) to edit the live-preview message; tool events still flush on the next loop. Edits to
+ *  one message are rate-limited tighter than sends, so pace them ~1.5s (vs every token). */
+const EDIT_THROTTLE_MS = 1500;
 
 /** Max length of a tool's arg preview in the live view. */
 const TOOL_ARG_MAX = 48;
@@ -200,18 +201,55 @@ function defaultErrorMessage(failed: TelegramFailure): string {
 }
 
 /**
- * Consume one turn's event stream into a Telegram chat, live. Tool calls + partial text stream into an
- * ephemeral draft; the final text is persisted with sendMessage (the draft is a ~30s preview, so the
- * real message MUST be sent to keep it, and a keepalive re-pushes it so a long step does not blank it).
- * Draft updates are best-effort: they are only a preview, and the authoritative sendMessage surfaces
- * any real failure (bad token, etc.) — a client that does not support drafts degrades to "final only".
+ * The terminal-write POLICY: resolve the single preview message into `text`. streamReply owns the
+ * preview lifecycle, so this composition of transport primitives lives here, not in telegram-api. One
+ * message → edit the preview in place; if the edit fails (preview gone, or a persistent 429/5xx) fall
+ * back to deleting the placeholder and sending fresh, so no "Thinking…" is left pinned above the answer.
+ * Many messages → delete the preview and send the whole answer as consecutive fresh messages (editing
+ * would pin the first chunk where an active group has scrolled past). No preview → fresh send. EMPTY text
+ * = "say nothing" → just delete the preview.
+ */
+async function finalize(
+  api: string,
+  botToken: string,
+  target: Target,
+  messageId: number | undefined,
+  text: string,
+  opts: { html?: boolean } = {},
+): Promise<void> {
+  const html = opts.html ?? true;
+  if (text.trim() === "") {
+    if (messageId !== undefined) await deleteMessage(api, botToken, target, messageId);
+    return;
+  }
+  if (messageId !== undefined) {
+    if (chunkText(text).length === 1) {
+      try {
+        await editMessageText(api, botToken, target, messageId, text, { html });
+        return;
+      } catch {
+        // Edit failed — the preview may be gone, or still there (429 retries exhausted / 5xx). Fall through
+        // to delete + fresh send below so a still-present "Thinking…" is not left pinned above the answer.
+      }
+    }
+    // Many chunks, OR a failed single-chunk edit: remove the placeholder best-effort (a lingering one above
+    // the answer is worse than the extra call), then send the whole reply as fresh, consecutive messages.
+    await deleteMessage(api, botToken, target, messageId).catch(() => {});
+  }
+  await sendMessage(api, botToken, target, text, { html });
+}
+
+/**
+ * Consume one turn's event stream into a Telegram chat, live. A single preview message is sent once and
+ * editMessageText'd in place with the latest tool calls + partial text (PLAIN); on completion it is
+ * edited into the final answer (HTML). Preview edits are best-effort (logged once if they fail); the
+ * final write is authoritative and surfaces a real failure (bad token, etc.).
  */
 async function streamReply(
   events: AsyncIterable<AgentEvent>,
   api: string,
   botToken: string,
   target: Target,
-  draftId: number,
   formatError: (failed: TelegramFailure) => string | undefined,
 ): Promise<void> {
   const tools: { label: string; status: "running" | "ok" | "error" }[] = [];
@@ -233,22 +271,46 @@ async function streamReply(
       .filter((s) => s.trim() !== "")
       .join("\n\n")
       .trim();
-    // Before the first reasoning/tool/text arrives the draft is empty — Telegram renders an empty draft
-    // as a bare "…", so show an explicit placeholder instead (this is also what the keepalive re-pushes).
+    // Before any reasoning/tool/text arrives, show an explicit placeholder rather than an empty edit.
     return v === "" ? "💭 Thinking…" : v;
   };
+
+  // The live preview is ONE real message: sent once (capturing its id + threading under the asker),
+  // then edited in place. messageId/lastSent are shared with the final write on completion.
+  let messageId: number | undefined;
+  let previewSent = false; // a placeholder send was attempted — guards against re-sending when no id came back
+  let finalized = false; // a terminal write (completed/failed) ran — the finally skips its orphan cleanup
+  let lastSent = "";
+  const flushPreview = async (): Promise<void> => {
+    const text = view();
+    if (text === lastSent) return; // skip an unchanged edit (Telegram rejects "message is not modified")
+    lastSent = text;
+    if (messageId !== undefined) {
+      await editMessageText(api, botToken, target, messageId, text); // plain — a partial answer may carry unbalanced HTML
+      return;
+    }
+    // No preview message yet. Send the placeholder ONCE; never re-send (that would spam a new message per
+    // frame). If Telegram returns ok WITHOUT a message_id (proxy / odd API base / unparseable body) we
+    // cannot edit — fail visibly and stop previewing (the final write still lands via finalize).
+    if (previewSent) return;
+    previewSent = true;
+    messageId = await sendMessage(api, botToken, target, text, { html: false });
+    if (messageId === undefined)
+      throw new Error("telegram sendMessage returned ok without a message_id — live preview disabled for this turn");
+  };
+
   // ── Live-preview pump: a SINGLE serialized writer. ──────────────────────────────────────────
-  // Events mutate state (thinking / tools / answer) and mark the preview dirty; the pump sends the
-  // LATEST view() with at most ONE sendMessageDraft in flight, paced by a throttle between sends, plus a
-  // trailing send after the final change. One-in-flight is the whole point: concurrent draft sends (an
-  // event update racing the keepalive, or a slow send overtaken by the next) can reach Telegram out of
-  // order — an older frame landing over a newer one is the "shows 3-4 steps, blanks, re-fills" flicker.
-  // Serializing through one writer keeps frames strictly monotonic; the keepalive marks dirty through
-  // the same writer, so it cannot race.
+  // Events mutate state (thinking / tools / answer) and mark the preview dirty; the pump edits the
+  // message to the LATEST view() with at most ONE edit in flight, paced by a throttle. One-in-flight is
+  // the whole point: concurrent edits can reach Telegram out of order — an older frame landing over a
+  // newer one is the "shows 3-4 steps, blanks, re-fills" flicker. Serializing keeps frames monotonic.
+  // (No keepalive: a real message does not expire, unlike the old 30s draft.)
   let dirty = false;
   let pumping = false;
   let stopped = false;
-  let draftErrLogged = false;
+  let previewErrLogged = false;
+  let pumpDone: Promise<void> | undefined;
+  let wakeThrottle: (() => void) | undefined; // set while the pump is mid-throttle; finish() cuts it short
 
   const runPump = async (): Promise<void> => {
     pumping = true;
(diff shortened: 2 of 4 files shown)


core/src/channels/telegram/telegram.ts
```diff
@@ -233,61 +231,82 @@ async function streamReply(
...
+
+  // The live preview is ONE real message: sent once (capturing its id + threading under the asker),
+  // then edited in place. messageId/lastSent are shared with the final write on completion.
+  let messageId: number | undefined;
+  let lastSent = "";
+  const flushPreview = async (): Promise<void> => {
+    const text = view();
+    if (text === lastSent) return; // skip an unchanged edit (Telegram rejects "message is not modified")
+    lastSent = text;
+    if (messageId === undefined) messageId = await sendMessage(api, botToken, target, text, { html: false });
```
kid7st: 「一条预览消息」不变量无守卫, 违反时是静默刷屏. flushPreview 靠 messageId === undefined 判断是否已发出预览; messageId 源自 callBotApi 成功分支的 await res.json().catch(() => ({})) (telegram-api.ts:53). 一旦 200 响应体解析失败 (代理 / 非标准 API base / 空 body), result 为 undefined、message_id 为 undefined, 却仍返回 ok:true. 于是 messageId 永远 undefined, 之后每个变化的 view() 都再发一条新消息, 把「原地编辑」退化成逐帧刷屏, 且无错误、不可诊断 (违反 fail-visibly). 整个设计的核心不变量寄生在 Telegram 必返 message_id 上却无守卫, 失败模式既坏又静默.
方向: send 路径上把「ok 但无 message_id」当作失败 (抛错, 让 previewErrLogged 落日志), 或在 flushPreview 对「已成功发过但拿不到 id」加守卫, 避免重复 send.

core/src/channels/telegram/telegram.ts
```diff
@@ -309,21 +328,29 @@ async function streamReply(
         if (t) t.status = e.isError ? "error" : "ok";
         touch();
       } else if (e.type === "completed") {
-        finish();
-        if (answer.trim() !== "") await sendMessage(api, botToken, target, answer);
+        await finish();
+        // Edit the preview into the final answer (HTML, plain fallback); the persisted message is the
+        // answer alone — the process (thinking/tools) was preview-only.
+        await sendFinalReply(api, botToken, target, messageId, answer.trim() !== "" ? answer : "(no reply)");
```
kid7st: 空答案的用户可见行为被静默改变. 旧实现 if (answer.trim() !== "") await sendMessage(...) 下, 空答案 turn 不留任何持久消息 (draft 自行过期). 新实现把占位消息编辑成 "(no reply)", 空答案 turn 现在会留下一条 (no reply) 消息. 这是合理选择 (持久消息无法像 draft 一样消失), 但 commit message 只讲机制、没提这个行为变化.
方向: 确认这是有意决策, 并在 commit/PR 描述里点明.

kid7st (commented):
Thesis: 用「一条真实消息」(sendMessage 发一次 → editMessageText 原地编辑, 完成时编辑成最终答案) 替换临时 sendMessageDraft (30s, 仅私聊/topic 可用), 使群里也能流式预览. 系统不变量 = 「一条预览消息, 反复编辑」.

---

**docs/telegram.md:10,149-156 — 文档与新机制脱节 (漏改的 companion text, 不在本 PR diff 内故放这里)**

PR 整体替换了流式机制, 但 docs/telegram.md 仍写着 "ephemeral Telegram draft"、"periodic keepalive while long steps run"、"final answer is sent with sendMessage". 这些现在全错: 没有 draft, 没有 keepalive, 最终答案是 editMessageText 编辑预览消息而非 sendMessage. 改了机制必须同步改描述它的文档, 否则是对下一个维护者撒谎.
方向: 把 "Streaming behavior" 一节改写为「一条预览消息 + 原地编辑 + 完成时编辑成 HTML 答案」, 删掉 draft/keepalive 措辞.

core/src/channels/telegram/telegram.ts
```diff
@@ -309,21 +339,29 @@ async function streamReply(
...
+        await finish();
         // Two audiences: the chat (customer-facing — formatError, neutral by default) and the operator
         // log (dev-facing — the full details, via the throw below + the handler's catch).
         const msg = formatError({ details: e.details, retryable: e.retryable });
-        if (msg && msg.trim() !== "") await sendMessage(api, botToken, target, msg).catch(() => {});
+        if (msg && msg.trim() !== "") {
+          const write =
+            messageId === undefined
+              ? sendMessage(api, botToken, target, msg, { html: false })
+              : editMessageText(api, botToken, target, messageId, msg);
```
kid7st: failed 的 user-facing 写入没有 completed 路径的「edit 失败→fresh send」兜底。completed 用 sendFinalReply：preview 不可编辑时 fallback 到一次全新 send，保证终态消息一定落地（这正是本 PR 引入 sendFinalReply 的目的，有测试覆盖）。这里却用裸 editMessageText(...).catch(() => {})：当 messageId 存在但消息已被删/不可编辑时，editMessageText 抛错被 .catch 吞掉，用户既看不到答案也看不到错误提示，只剩一条停在 "💭 Thinking…"/中间态的 preview。

两条都是面向用户的终态消息，却被区别对待：答案 authoritative + 兜底，错误提示 best-effort + 删了就丢。与 fail-visibly 冲突——责任链终点（用户）在 system error 时反而拿不到任何信号。此分支也无对应测试（"a failed event tells the user" 用 edit 必成功的 okFetch，未覆盖 edit 失败）。

建议：failed 的 user-facing 写入复用 sendFinalReply(api, botToken, target, messageId, msg)，与 completed 共享同一套兜底（formatError 默认中性文案 HTML-safe，HTML+plain fallback 无副作用）。operator 侧的 throw 不变。

kid7st (commented):
本次改动把 live preview 从 sendMessageDraft（私聊/forum-topic 限定，群组失败）换成「单条真实消息 sendMessage 一次 → editMessageText 原地更新 → 完成时编辑成最终答案」，并新增 sendFinalReply 处理 overflow 与「消息不可编辑则 fresh send」兜底。

核心概念与命名自洽，单写者 pump 的 in-flight 序列化 + finish 的 await 保证最终写是最后一帧，no-message_id 的防刷屏逻辑也正确。一个责任分配上的不一致见 inline。

kid7st (commented):
Thesis: replace the draft-based live preview with a single sendMessage("💭 Thinking…") → editMessageText-in-place mechanism, so groups (not just private/forum-topic) get a live preview, with the final answer editing that same message. Core mechanism is clear. Two findings below.

core/src/channels/telegram/telegram-api.ts
```diff
@@ -94,31 +105,72 @@ export async function sendMessage(api: string, botToken: string, t: Target, body
...
+  messageId: number | undefined,
+  body: string,
+  opts: { html?: boolean } = {},
+): Promise<void> {
+  const html = opts.html ?? true;
+  if (messageId !== undefined) {
+    const chunks = chunkText(body);
+    try {
+      await editMessageText(api, botToken, t, messageId, chunks[0] ?? body, { html });
+      for (const chunk of chunks.slice(1)) await sendMessage(api, botToken, { ...t, replyTo: undefined }, chunk, { html });
```
kid7st: Over-broad `try`: it wraps both `editMessageText(chunks[0])` AND this continuation loop. When a reply >4096 chars edits the first chunk fine but a continuation `sendMessage` throws, the `catch` below falls through to `sendMessage(t, body)` and re-sends the ENTIRE body as fresh messages — the user sees the first chunk twice (edited preview + fresh) plus a duplicated full answer. The catch comment says its intent is only "the preview message is gone / un-editable", i.e. it should guard `editMessageText` alone. A continuation-send failure is an ordinary send failure (sendMessage already retries 429) and should propagate, not trigger a full re-send.

Direction: narrow the `try` to wrap only `editMessageText(chunks[0])`; move the continuation loop out of the try so its failures propagate.

core/src/channels/telegram/telegram.ts
```diff
@@ -309,21 +339,27 @@ async function streamReply(
...
+        await sendFinalReply(api, botToken, target, messageId, answer.trim() !== "" ? answer : "(no reply)");
         return;
       } else if (e.type === "failed") {
-        finish();
+        await finish();
         // Two audiences: the chat (customer-facing — formatError, neutral by default) and the operator
         // log (dev-facing — the full details, via the throw below + the handler's catch).
         const msg = formatError({ details: e.details, retryable: e.retryable });
-        if (msg && msg.trim() !== "") await sendMessage(api, botToken, target, msg).catch(() => {});
+        if (msg && msg.trim() !== "") {
```
kid7st: Asymmetric terminal handling leaves an orphaned placeholder. The `completed` path floors an empty answer to `(no reply)`, so the "💭 Thinking…" placeholder is always edited away. But here, when a custom `formatError` returns undefined/"" (the developer's "say nothing to the user" intent), the final write is skipped and the placeholder real message stays stuck showing "💭 Thinking…" forever. The old draft mechanism made "say nothing" = draft expires = clean; a persisted message does not expire, so the suppress intent now produces a dead "thinking" message.

Direction: when `failed` and `msg` is empty, edit the placeholder into a neutral fallback, or deleteMessage it, so "suppress" again equals "no residue".

core/src/channels/telegram/telegram-api.ts
```diff
@@ -94,31 +105,87 @@ export async function sendMessage(api: string, botToken: string, t: Target, body
...
+      await editMessageText(api, botToken, t, messageId, chunks[0] ?? body, { html });
+      edited = true;
+    } catch {
+      // The preview message is gone / un-editable — fresh-send the whole body so the reply still lands.
+      await sendMessage(api, botToken, t, body, { html });
+    }
+    // Only after a SUCCESSFUL edit do the continuation chunks follow. A continuation-send failure is an
+    // ordinary send failure (sendMessage already retries 429) and propagates — it must NOT re-send the
+    // whole body (the user would see the first chunk twice).
+    if (edited) for (const chunk of chunks.slice(1)) await sendMessage(api, botToken, { ...t, replyTo: undefined }, chunk, { html });
```
kid7st: 长回复在活跃群里被割裂，且 tradeoff 未在 PR body 说明。chunk[0] 被编辑进回合开始时发的 preview 占位消息（钉在它的原始位置），continuation chunks 在回合结束时追加到底部。本 PR 核心动机正是“群里之前没有 live preview”——但群是消息流动的：preview 占位发出后、回合结束前，别人的消息会把它顶上去，于是多 chunk 回答的第 1 段停在被顶走的旧位置、第 2 段及之后落在底部，读者在底部看到没有上文的后半截。旧实现在结束时用一次 sendMessage 连续发全部 chunk，不存在割裂。这是面向群场景的回归，PR body 只描述了机制（first chunk edits preview, rest sent after），没点出 chunk[0] 被钉在可能已被顶走的位置这个后果。

方向：多 chunk 时删掉 preview 占位、结束时连续 fresh-send 全部 chunk（保证连续，仅长回答放弃就地编辑的连续性），单 chunk 仍走就地编辑；或保留现状但在 PR body / docs 显式声明该 tradeoff。

core/test/telegram.test.ts
```diff
@@ -18,8 +18,11 @@ function replyingAgent(reply = "") {
   return { agent, calls };
 }
 
-/** Let fire-and-forget turns (started after the 200) run before asserting. */
-const flush = () => new Promise((r) => setImmediate(r));
+/** Let fire-and-forget turns (started after the 200) run before asserting. The streaming flow has
+ *  several hops (placeholder send → finish awaits the pump → final edit), so settle multiple ticks. */
+const flush = async () => {
+  for (let i = 0; i < 8; i++) await new Promise((r) => setImmediate(r));
```
kid7st: flush() 用固定 8 个 tick 兜底 fire-and-forget，把测试 settle 耦合到实现的 await 跳数。注释明说是因为 streaming flow 有好几跳。这把测试通过条件钉死在当前实现的 await 数量上：未来给 streamReply / finalizeReply 多加一跳，非 fake-timer 的测试就会在状态未结算完时断言 → false green 或 flaky，失败原因与被测语义无关。

方向：fire-and-forget 的测试等一个确定信号（如 finalizeReply 完成时 resolve 的 barrier / 轮询最终 write 出现），而非固定 tick 数；fake-timer 那批已用 advanceTimersByTimeAsync，是更稳的模板。

kid7st (commented):
Thesis: 用 sendMessage 占位 + editMessageText 就地编辑（终态走 finalizeReply: edit→fresh-send fallback, 空 body=delete）替换 sendMessageDraft，让群里也有 live preview。机制设计自洽，throttle 中断、single-writer 单调性、no-message_id 兜底、HTML→plain fallback 都正确。两条 finding 见 inline。

core/src/channels/telegram/telegram.ts
```diff
@@ -233,61 +231,93 @@ async function streamReply(
...
+    const text = view();
+    if (text === lastSent) return; // skip an unchanged edit (Telegram rejects "message is not modified")
+    lastSent = text;
+    if (messageId !== undefined) {
+      await editMessageText(api, botToken, target, messageId, text); // plain — a partial answer may carry unbalanced HTML
+      return;
+    }
+    // No preview message yet. Send the placeholder ONCE; never re-send (that would spam a new message per
+    // frame). If Telegram returns ok WITHOUT a message_id (proxy / odd API base / unparseable body) we
+    // cannot edit — fail visibly and stop previewing (the final write still lands via sendFinalReply).
```
kid7st: Comment names `sendFinalReply`, which doesn't exist — the terminal-write function is `finalizeReply`. A reader chasing this name finds nothing (companion text contradicts the code). Fix the name to `finalizeReply`.

core/src/channels/telegram/telegram.ts
```diff
@@ -309,21 +339,27 @@ async function streamReply(
...
+          const msg = formatError({ details: e.details, retryable: e.retryable }) ?? "";
+          await finalizeReply(api, botToken, target, messageId, msg, { html: false }).catch(() => {});
+        }
         throw new Error(`agent failed: ${e.details} (retryable=${e.retryable})`);
       }
     }
     throw new Error("stream ended without a terminal event"); // violates SPEC MUST 1
   } finally {
-    finish();
+    await finish();
```
kid7st, line 409: Abnormal-termination paths leave a dead "💭 Thinking…" placeholder. `completed`/`failed` each call `finalizeReply` to finish; but the `stream ended without a terminal event` throw (and any mid-loop exception) reaches only `finally { await finish() }`, which stops the pump without finalizing/cleaning the preview. The old draft auto-expired in ~30s (self-healing); a persisted message lingers forever. This is a behavior change vs the old mechanism, not mentioned in the PR body — on a SPEC MUST 1 violation the user sees a stuck "Thinking…". Direction: in `finally`, if not already finalized (add a `finalized` flag) and `messageId !== undefined`, delete the placeholder so the exception path leaves no residue, matching the old draft's self-cleanup.

kid7st (commented):
Thesis: replace sendMessageDraft (forum-topic-only, fails in groups) with one preview message edited in place (sendMessage once → editMessageText), with finalizeReply as the single terminal-write path (edit→fresh-send fallback; empty body=delete). The mechanism switch and the single-writer/interruptible-throttle invariant hold up. Two findings, both inline.

core/src/channels/telegram/telegram-api.ts
```diff
@@ -94,31 +105,91 @@ export async function sendMessage(api: string, botToken: string, t: Target, body
...
+  if (opts.html && !result.ok && PARSE_ERROR.test(result.description))
+    result = await callBotApi(api, botToken, "editMessageText", base);
+  if (!result.ok && /message is not modified/i.test(result.description)) return;
+  if (!result.ok) throw new Error(`telegram editMessageText failed: ${result.status} ${result.description}`.trim());
 }
 
+/**
+ * Deliver the terminal message (the completed answer or a failure notice) into the chat — the single
+ * final-write path for both. Edits the live-preview message into `body` (HTML default, plain fallback
+ * inside editMessageText); a long body keeps its first chunk in the edit and sends the rest after it
```
kid7st: Docstring contradicts the code. It says "a long body keeps its first chunk in the edit and sends the rest after it (without re-quoting the asker)" — but the multi-chunk branch calls `removePreview()` (deleteMessage) then `sendMessage` for the WHOLE body; it does not keep the first chunk in the edit. The else-branch inline comment and docs/telegram.md both describe the real behavior (delete preview, re-send the whole answer consecutively); only this contract docstring describes the rejected approach. Rewrite: long body → delete the preview and send the entire answer as fresh consecutive messages (first chunk quotes the asker, the rest don't).

core/src/channels/telegram/telegram.ts
```diff
@@ -309,21 +340,32 @@ async function streamReply(
...
         // Two audiences: the chat (customer-facing — formatError, neutral by default) and the operator
-        // log (dev-facing — the full details, via the throw below + the handler's catch).
-        const msg = formatError({ details: e.details, retryable: e.retryable });
-        if (msg && msg.trim() !== "") await sendMessage(api, botToken, target, msg).catch(() => {});
+        // log (dev-facing — the full details, via the throw below + the handler's catch). Same terminal
+        // write as completed: edit → fresh-send if the preview is gone; an empty notice deletes the
+        // placeholder (suppress = no residue). Best-effort — we throw below regardless.
+        {
+          const msg = formatError({ details: e.details, retryable: e.retryable }) ?? "";
+          await finalizeReply(api, botToken, target, messageId, msg, { html: false }).catch(() => {});
```
kid7st: Silent behavior change for the failure notice. Pre-PR it went through `sendMessage` (HTML with plain fallback); now `finalizeReply(..., { html: false })` sends plain only, while the completed answer still uses HTML. For the default message and for surfacing `failed.details` (angle brackets) the result is unchanged or better, so it's defensible — but a custom `onError` returning valid HTML markup now renders literal tags, and answer/error formatting are now asymmetric. Not mentioned in the PR description or docs/telegram.md. Confirm intent and document that `onError` output renders as plain text (or pass `html: true` to keep prior behavior).

kid7st (commented):
Thesis: replace the `sendMessageDraft` (private/forum-only, 30s ephemeral) live preview with a standard `sendMessage`-once → `editMessageText`-in-place mechanism that also works in groups, plus a single `finalizeReply` terminal-write path. Two findings, both inline.

core/src/channels/telegram/telegram.ts
```diff
@@ -309,21 +340,34 @@ async function streamReply(
...
         throw new Error(`agent failed: ${e.details} (retryable=${e.retryable})`);
       }
     }
     throw new Error("stream ended without a terminal event"); // violates SPEC MUST 1
   } finally {
-    finish();
+    await finish();
+    // Abnormal exit (stream ended without a terminal, the generator threw, or the consumer abandoned): no
+    // terminal write ran, so remove the dangling "💭 Thinking…" placeholder rather than leave it stuck.
+    if (!finalized && messageId !== undefined) await finalizeReply(api, botToken, target, messageId, "").catch(() => {});
```
kid7st: Abnormal exit (no terminal event / generator threw / consumer abandoned) silently deletes the preview. Two problems: (a) the comment calls it "the dangling 💭 Thinking… placeholder", but by this point the message may show accumulated thinking + tool activity + partial answer — you're deleting visible work, not a placeholder (the thinking-only test even deletes "💭 …"). (b) User-facing inconsistency: a `failed` event shows a neutral "something went wrong" notice, but an abnormal end shows the user nothing — the message just vanishes. Both are "the turn died", only one fails visibly to the user; the operator log records it, the user gets silence. Direction: leave the last partial preview in place, or edit it to a neutral terminal notice (reuse the formatError path), instead of deleting. If deletion is intended, fix the comment to say "delete whatever partial preview exists" and justify the no-notice disappearance.

core/src/channels/telegram/telegram-api.ts
```diff
@@ -94,31 +105,92 @@ export async function sendMessage(api: string, botToken: string, t: Target, body
...
 
+/**
+ * Deliver the terminal message (the completed answer or a failure notice) into the chat — the single
+ * final-write path for both. A single-message body edits the live-preview in place (HTML default, plain
+ * fallback inside editMessageText), falling back to a fresh send only if the edit fails (preview gone)
+ * or there was no preview. A body that spans MULTIPLE messages instead deletes the preview and sends the
+ * whole answer as consecutive fresh messages — editing would pin the first chunk where an active group
+ * has scrolled past. An EMPTY body means "say nothing" — delete the preview, leaving no residue.
+ */
+export async function finalizeReply(
```
kid7st: finalizeReply puts channel UX policy in the wire-transport module. The file header says this module is "the wire calls the channel orchestrates (telegram.ts). No agent, no rendering — just fetch … protocol's hard edges". finalizeReply isn't a wire primitive — it encodes product UX decisions: "edit-in-place vs delete-and-resend by chunk count because an active group scrolled past the preview" and "empty body = delete the placeholder". sendMessage's chunking/HTML-fallback are genuine protocol edges (4096 limit, parse_mode); the edit-vs-resend scroll decision is not. This orchestration belongs in telegram.ts alongside streamReply, which owns the preview lifecycle. Direction: move finalizeReply into telegram.ts (it only composes the editMessageText/sendMessage/deleteMessage primitives this module already exports), or relabel the transport module's contract if policy is intentionally allowed here.

kid7st (commented):
Thesis: replaces the forum-topic-only `sendMessageDraft` ephemeral draft with the standard one-message `sendMessage`+`editMessageText` live-preview, plus a single `finalizeReply` terminal-write path. Core mechanism is sound; findings below.

**telegram.ts:618 — stale comment references the removed draft id**
`// Run at DEQUEUE time (serialized), so the draft id, lifecycle log, and engine turn all reflect …` — this PR removed `draftSeq`/the draft id, so the comment now lies about what dequeue-time ordering protects. (Not in this diff's hunks, so flagged here rather than inline.) Direction: drop "the draft id,".

core/src/channels/telegram/telegram.ts
```diff
@@ -419,15 +516,36 @@ export function telegramEnvelope(m: TelegramMessage): string {
...
+function messageMentionsBot(m: TelegramMessage, botUsername: string | undefined): boolean {
+  const name = normalizedBotUsername(botUsername);
+  if (!name) return false;
+  const target = `@${name}`;
+  const text = m.text ?? m.caption ?? "";
+  const lower = text.toLowerCase();
+  if (lower.includes(target)) return true;
+
+  // Telegram offsets are UTF-16 code units, which matches JS string slice indexing.
+  for (const e of [...(m.entities ?? []), ...(m.caption_entities ?? [])]) {
```
kid7st: The `mention`-entity loop is inert. `lower.includes(target)` at line 530 runs first, and a `mention` entity by definition means the text literally contains `@username` — so whenever this loop *could* match, the substring check has already returned true. The loop adds no recall (captions + case are handled by the substring branch, which reads `m.text ?? m.caption` and lowercases) and no precision (`bob@mybot.com` still false-positives, since the substring check fires first). Dead code dressed as entity-based routing.

Direction: pick the intent. For precision (only real mentions) drop the substring check and gate on `mention` entities whose slice equals `target`, keeping plain `includes` only as a fallback when entities are absent. For loose matching, delete the entity loop and keep the case-insensitive substring. Substring-then-entity delivers neither.

core/test/telegram.test.ts
```diff
@@ -59,6 +82,30 @@ describe("defaultTelegramRoute + telegramEnvelope", () => {
     expect(defaultTelegramRoute(mention, { botUsername: "mybot" })).toEqual({});
   });
 
+  it("recognizes @mentions anywhere, case-insensitively, including caption entities", () => {
```
kid7st: This test can't exercise what its name claims. Both assertions pass through the substring branch (`"file for @mybot".includes("@mybot")` is already true before the entity loop is reached), so deleting the entity loop leaves the test green — it protects nothing about entity handling. It's the proof that the entity loop is inert: no scenario with a `mention` entity can be true while the substring is false.

Direction: fold into the `messageMentionsBot` fix — once substring/entity precedence is corrected, add a case the substring branch alone gets wrong (`bob@mybot.com` must NOT summon; a real `mention` entity must).

kid7st (commented):
Thesis: replace the private-only `sendMessageDraft` preview with "one real message, edited in place" (`sendMessage` once → `editMessageText`), with `finalize` as the terminal-write policy. That core is coherent. One cross-cutting concern that anchors to no single diff line:

**Scope contradicts the description.** The body says #4 (entity-based @mention + strip) "follow[s] in a separate PR," but #4's @mention routing is in *this* diff: `messageMentionsBot`, the new `entities`/`caption_entities` fields, the pre-ACK `await botInfoReady` gate, and two tests. Only "+ strip" is absent. So a partial — and (see the inline on `messageMentionsBot`) inert — slice of the deferred work landed here. Either the body is stale, or this belongs in the follow-up PR where it can be built to add real precision instead of dead code.

core/src/channels/telegram/telegram.ts
```diff
@@ -309,21 +388,39 @@ async function streamReply(
...
     }
     throw new Error("stream ended without a terminal event"); // violates SPEC MUST 1
   } finally {
-    finish();
+    await finish();
+    // Abnormal exit (stream ended without a terminal, the generator threw, or the consumer abandoned): no
+    // terminal write ran. Show the SAME neutral notice a `failed` event would — the preview may show real
+    // partial work, so don't delete it silently, and don't leave the user in silence. A suppressing
+    // onError still collapses to a delete (finalize on empty text).
+    if (!finalized && messageId !== undefined) {
```
kid7st: This notice is gated on `messageId !== undefined`, so an abnormal stream end (no terminal event / generator threw) leaves the user silent when the preview never got an id — exactly the placeholder-send-threw / ok-without-`message_id` case the pump treats as "preview disabled." That contradicts the adjacent comment "don't leave the user in silence." The `completed`/`failed` paths call `finalize` with a possibly-undefined `messageId` and fresh-send in that case; only this path refuses to.

Direction: call `finalize` unconditionally here (it already fresh-sends when `messageId` is undefined, and its `.catch` swallows a hard failure), or narrow the comment to match the guard.

core/src/channels/telegram/telegram-api.ts
```diff
@@ -94,31 +105,51 @@ export async function sendMessage(api: string, botToken: string, t: Target, body
...
+  if (opts.html && !result.ok && PARSE_ERROR.test(result.description))
+    result = await callBotApi(api, botToken, "editMessageText", base);
+  if (!result.ok && /message is not modified/i.test(result.description)) return;
+  if (!result.ok) throw new Error(`telegram editMessageText failed: ${result.status} ${result.description}`.trim());
 }
 
+/** Delete a message — the wire primitive for removing a preview placeholder (the finalize POLICY lives
+ *  in telegram.ts). */
+export async function deleteMessage(api: string, botToken: string, t: Target, messageId: number): Promise<void> {
+  await callBotApi(api, botToken, "deleteMessage", { chat_id: t.chatId, message_id: messageId });
```
kid7st: `deleteMessage` discards `callBotApi`'s result — no `ok` check, no throw, no log — while its siblings `sendMessage`/`editMessageText` throw on failure. In `finalize`'s empty/suppressed-notice path (`text.trim() === ""` → `deleteMessage`), a failed delete strands the placeholder in the chat *forever* as `💭 Thinking…`: no answer, no operator signal, bot looks permanently stuck with nothing diagnosable. (Swallowing is right in the multi-chunk path — a failed cleanup mustn't abort the answer — so don't throw unconditionally.) Make the failure visible to the policy layer: return `ok` (or throw) and have `finalize` log-once on a failed cleanup, mirroring the pump's `previewErrLogged` pattern. Fail-visibly gap.

core/src/channels/telegram/telegram.ts
```diff
@@ -309,21 +376,41 @@ async function streamReply(
...
     throw new Error("stream ended without a terminal event"); // violates SPEC MUST 1
   } finally {
-    finish();
+    await finish();
+    // Abnormal exit (stream ended without a terminal, the generator threw, or the consumer abandoned): no
+    // terminal write ran. Show the SAME neutral notice a `failed` event would — the preview may show real
+    // partial work, so don't delete it silently, and don't leave the user in silence. A suppressing
+    // onError still collapses to a delete (finalize on empty text).
+    if (!finalized) {
+      const notice = formatError({ details: "the turn ended without completing", retryable: true }) ?? "";
```
kid7st: The new `finally` orphan-cleanup hardcodes `retryable: true`, so *every* abnormal exit — a generator that threw a fatal (bad config, permanent), a SPEC-violating stream, an abandoned consumer — tells the user the default `⚠️ Temporary problem — please try again.` Previously an abnormal exit produced no user notice, so this is a new user-facing decision not stated in the PR description, and `try again` on a permanent failure invites a failing retry loop. Confirm intent; if retryability is genuinely unknown here, prefer the neutral non-retry wording, or document why retry is the right default for unknown crashes.

kid7st (commented):
Thesis: replace the ephemeral `sendMessageDraft` (30s, private/forum-topic only) with one real message — `sendMessage` a `💭 Thinking…` placeholder once, `editMessageText` it in place while streaming, then `finalize` it into the answer (edit / delete+fresh-send for long answers / delete on empty). Clean layering (finalize = policy in telegram.ts, primitives in telegram-api.ts) and the single-writer pump is preserved. Two findings, both inline.

kid7st (commented):
Thesis: reemplaza el preview basado en `sendMessageDraft` (private/forum-only) por "un mensaje real, editado en sitio" — `sendMessage` una vez → `editMessageText` en bucle → `finalize` (edit / delete+send / fresh-send). Nuevos conceptos: primitivas `editMessageText`/`deleteMessage`, la POLICY `finalize`, el pump single-writer con throttle interrumpible, y el flag `finalized`. El mecanismo de ordering (`await finish()` espera el envío en vuelo del placeholder antes de `finalize`) es sólido. Findings:

**Cambio de comportamiento `(no reply)` sin test que lo proteja** (`core/test/telegram.test.ts`). El commit lo marca deliberado: un turno con answer vacío deja `"(no reply)"`. Ningún test afirma que un `completed` sin texto edite el preview a `"(no reply)"` (sin `deleteMessage`). `replyingAgent()` con reply vacío se usa en varios tests pero ninguno asserta ese texto — una regresión (volver a borrar/no enviar) pasaría verde.

**Throttle interrumpible afirmado pero sin test** (`telegram.ts:328-337`, `355-358`). El commit promete que `finish()` corta el throttle para que la escritura final no se retrase hasta `EDIT_THROTTLE_MS`. El test fake-timer de streaming avanza 4000ms (> throttle), así que no distinguiría un throttle no-interrumpible; quitar `wakeThrottle` retrasaría la final hasta 1.5s sin fallar ningún test. Test posible: entrar en throttle, luego `completed`, y afirmar que el edit final ocurre sin avanzar otro `EDIT_THROTTLE_MS`.

core/src/channels/telegram/telegram.ts
```diff
@@ -309,21 +377,43 @@ async function streamReply(
         if (t) t.status = e.isError ? "error" : "ok";
         touch();
       } else if (e.type === "completed") {
-        finish();
-        if (answer.trim() !== "") await sendMessage(api, botToken, target, answer);
+        await finish();
+        // Edit the preview into the final answer (HTML, plain fallback); the persisted message is the
+        // answer alone — the process (thinking/tools) was preview-only.
+        await finalize(api, botToken, target, messageId, answer.trim() !== "" ? answer : "(no reply)");
+        finalized = true;
```
kid7st, line 388: `finalized = true` se pone *después* de `finalize`, que aquí NO va envuelto en `.catch`. Si `finalize` lanza (fallo persistente de entrega: `sendMessage` agota reintentos / 403 bad-token, tras fallar también el `editMessageText` interno), `finalized` queda `false`, el throw sube al `finally`, y `!finalized` dispara el aviso orphan `"the turn ended without completing"` → `"something went wrong"`. Pero el turno SÍ completó; solo falló la entrega. Resultado: (a) se etiqueta mal al usuario como turno incompleto, (b) se intenta una segunda entrega redundante. El camino `failed` (L396-398) no sufre esto porque envuelve `finalize` en `.catch` y setea `finalized` incondicionalmente.

Contradice el comentario del propio flag (L279, "a terminal write (completed/failed) ran"): pretende significar "se manejó un evento terminal", pero rastrea "finalize retornó sin lanzar". El orphan-cleanup del `finally` es para "no llegó ningún terminal" — un fallo de entrega de un turno completado no es orphan; ya se hace visible por el throw propagado + el log del caller.

Dirección: setear `finalized = true` al ENTRAR a la rama `completed` (antes de `finalize`), acorde al intent documentado.

core/src/channels/telegram/telegram.ts
```diff
@@ -200,18 +201,52 @@ function defaultErrorMessage(failed: TelegramFailure): string {
...
+    if (messageId !== undefined) await deleteMessage(api, botToken, target, messageId);
+    return;
+  }
+  if (messageId !== undefined) {
+    if (chunkText(text).length === 1) {
+      try {
+        await editMessageText(api, botToken, target, messageId, text, { html });
+        return;
+      } catch {
+        // preview gone — fall through to a fresh send
```
kid7st: This catch treats EVERY editMessageText failure as "preview gone", but only "message to edit not found" actually means the placeholder vanished. 429 rate-limit exhaustion (callBotApi gives up after 3 retries; a streaming turn issues many edits, so the final one hitting the limit is realistic) and transient 5xx/network reach the same catch while the placeholder STILL exists — the fallthrough fresh-sends the answer without deleting it, leaving a dead "💭 Thinking…"/partial preview pinned above the real answer. That is exactly the residue the multi-chunk branch (:233) and the suppressing-error test work to eliminate.

Direction: `await deleteMessage(...).catch(() => {})` before the fresh send, mirroring the multi-chunk branch — when truly gone it no-ops harmlessly; when it survived, the orphan is cleared.

Same function, cleanup discipline: deleteMessage is invoked three inconsistent ways — empty-text (:220, throws), multi-chunk (:233, swallowed), single-chunk-fail (absent). Placeholder removal is one responsibility; make it uniformly best-effort. The :220 throw is masked only because `completed` never passes empty text and the failed/abnormal callers wrap finalize in `.catch` — the function leans on caller discipline instead of owning its own best-effort contract.

kid7st (commented):
Thesis: replace the ephemeral `sendMessageDraft` preview with ONE persisted message (`sendMessage` once → `editMessageText` in place → `finalize`), so live preview works in groups. Concept is sound; the weak spot is `finalize`'s error-handling responsibility.

Test gap (not anchorable to a diff line): the orphaned-placeholder scenario — a single-chunk final edit that FAILS while the message still exists (429 exhaustion / transient 5xx) — is uncovered. The existing fresh-send-fallback test uses "message to edit not found" (genuinely gone), so it cannot catch the missing delete described in the inline comment.

kid7st (commented):
No findings.

Thesis: live preview migrated from sendMessageDraft (private/forum-topic only) to one real message (sendMessage once -> editMessageText in place -> finalize: edit / delete+fresh-send). Reviewed the finalize policy branches, the single-writer pump + finish() ordering, error-responsibility split (completed surfaces via throw; failed swallows because it re-throws), the no-message_id guard, and companion docstrings/docs — all consistent and covered by failure-path tests. The empty-completed -> (no reply) behavior change is explicitly documented in the commit message.

PR #97, opened 2026-07-01
feat(auth): default the credentials file to project-level

kid7st:
## What

Auth resolves to **`<dir>/.fastagent/auth.json`** by default (was the global `~/.fastagent/auth.json`), as `--auth-path` > `FASTAGENT_AUTH_PATH` > project default, with **no implicit project↔global fallback**.

- `login`/`dev`/`start`/`invoke`/`info` are project-level; **`chat` is exempt** (uses pi's own `~/.pi` TUI auth).
- **Migration:** existing global logins aren't used automatically — `dev`/`start`/`invoke` print a hint pointing at the global credential (`set FASTAGENT_AUTH_PATH=…`). Sharing one file across projects stays safe (cross-process refresh lock).

## Why

Project-level default gives **isolation** (different agents → different accounts) and **fail-visibly** (a missing credential surfaces instead of being masked by a machine-global one absent on a fresh box). The fallback is refused specifically to avoid copy-on-rotate divergence of a read-global/write-project shape.

## API (breaking, pre-1.0)

- Exported `FASTAGENT_AUTH_PATH` constant → **`GLOBAL_AUTH_PATH`** (now the global *location*, not the default; no alias — the name would mislead against the same-named env var).
- `createPiAgent`, `createPiAgentFromDefinition`, `createPiModels` gain an `authPath` option; the dir-aware `createPiAgentFromDefinition` defaults it project-level.

## Notable internals

- **Leak guard:** `ensureInTreeStateSelfIgnored` is the single owner (home exclusion via canonical-path compare + path-location containment via `isUnderStateDir`); `ensureStateDirSelfIgnored` is now module-private. `.fastagent` root + auth/sessions defaults derive from one `projectStateDir()`.
- Path overrides now expand a leading `~` (was a silent `<cwd>/~` footgun).
- Fix: `login` loads `.env` from cwd (was `./<provider>`).

## Review tier

Touches the public API surface (`GLOBAL_AUTH_PATH` rename, `authPath` option) → **waits for review** per CONTRIBUTING.

## Test

`cd core && npm run typecheck && npm test` — 260 pass, typecheck clean. New coverage: `authPath` credential seam, `~` expansion, leak-guard predicate + `$HOME` exclusion (symlink-safe), migration hint, `login` `.env`-from-cwd.


kid7st (commented):
Thesis: auth gains an `authPath` credential seam defaulting project-level (`<dir>/.fastagent/auth.json`), with a renamed `GLOBAL_AUTH_PATH` constant and a re-architected leak guard (`ensureInTreeStateSelfIgnored` as single owner). One finding, cross-file so it can't anchor to the diff:

**Leak-guard "single owner" omits telegram-files** — `definition.ts` frames `ensureInTreeStateSelfIgnored` as "The single owner of the leak-guard invariant" and `config.ts` `projectStateDir` as "THE single definition of that path segment," the stated point being that "a new command can't half-implement it and leak a secret." But `core/src/channels/telegram/telegram-api.ts:22,180` independently hardcodes `join(".fastagent", "telegram-files")` and writes user media to `<cwd>/.fastagent/telegram-files/...`, with a comment asserting it is "git-ignored." That path is never registered with the single owner — its git-ignored-ness is only *incidentally* true because auth now defaults in-tree and trips the self-ignore. Run a telegram bot with both `--sessions-dir /vol` and `--auth-path <global>`: `ensureInTreeStateSelfIgnored` sees no in-tree path, `.fastagent/.gitignore` is never written, and downloaded media lands committable — contradicting both the single-owner/single-definition claim and telegram's own "git-ignored" comment. Pre-existing gap (only sessions triggered self-ignore before), but the new abstraction claims a completeness it lacks.

Direction: either thread the telegram base path through `ensureInTreeStateSelfIgnored` (derive it from `projectStateDir`), or scope both docstrings to auth+sessions and note telegram-files relies on the incidental self-ignore.

Everything else checks out (no leftover `FASTAGENT_AUTH_PATH` constant refs, typecheck clean, tests are integration-level covering the failure paths, `~`-expansion + login `.env`-from-cwd both tested).

kid7st:
Addressed the review finding (telegram-files vs the leak guard) in `01aeab9`.

The clean fix — registering `<cwd>/.fastagent/telegram-files` with `ensureInTreeStateSelfIgnored` — is blocked by layering: channels must not import engine internals (`engines/pi/definition.ts`), per the engine-neutrality rule. And an "always self-ignore" alternative would cascade into reworking the `.env`-read test's observable.

So I made the abstraction's claim honest instead of leaving it overclaiming:

- **`ensureInTreeStateSelfIgnored` docstring** — narrowed from "a new command can't half-implement it and leak a secret" to: it owns the self-ignore *mechanism* (home exclusion + the single `.gitignore` write, with `ensureStateDirSelfIgnored` private so no bypass), but it is **not auto-discovery** — it fires only for the paths it's given. Explicitly documents that `telegram-files` is covered only incidentally (via the opener self-ignoring for sessions/auth), so an all-external `--sessions-dir` + `--auth-path` config leaves it committable, and a channel wanting a guarantee must register its dir.
- **`telegram-api.ts` `FILES_SUBDIR` comment** — stopped it claiming a flat "git-ignored"; it now states the coverage depends on dev/start self-ignoring the state dir (the default) and is not registered with the guard.

Net: the residual gap (both sessions and auth pointed outside the tree, with a telegram channel writing media, then the repo committed) is narrow, pre-existing, and user-media rather than secrets — now a documented limitation instead of a hidden one. Closing it properly wants a neutral channel-side state-dir seam, which is a separate change.

core/src/channels/telegram/telegram-api.ts
```diff
@@ -17,8 +17,10 @@ const PARSE_ERROR = /can't parse|entit|unsupported|unclosed|tag/i;
  *  engine resizes images to the model's needs, so this is a transport guard, not the model size limit. */
 const MAX_DOWNLOAD_BYTES = 20 * 1024 * 1024;
 
-/** Where inbound files land (the agent reads them by path). Machine state under .fastagent: persists,
- *  git-ignored, not auto-cleaned (like sessions) — a long-running bot's operator manages the dir. */
+/** Where inbound files land (the agent reads them by path). Machine state under `.fastagent`: persists,
+ *  not auto-cleaned (like sessions) — a long-running bot's operator manages the dir. Git-ignored via
+ *  `<dir>/.fastagent/.gitignore="*"`, which dev/start write when self-ignoring the state dir (the
+ *  default; NOT registered with that guard, so an all-external sessions+auth config leaves it tracked). */
```
kid7st, line 23: 此 PR 把 self-ignore 从「sessions 默认在树内才触发」改成「sessions **或** auth 落在树内就触发」，于是 `telegram-files` 成了唯一**未向守卫注册**的在树状态路径，而它装的是用户上传的下载文件（比 sessions/auth 更敏感的第三方内容）。当 `start --sessions-dir /vol --auth-path /vol/auth.json` 两者都外部时，`.fastagent/.gitignore` 永不写入，adapted repo（无根 `.gitignore`）里这些用户文件变成 untracked-but-committable。注释诚实记录了缺口，但此 PR 恰好重做了该机制并编辑了这条注释——现在是「有意保留」还是「顺手漏掉」值得确认。

方向：若接受此权衡，当前注释已足够；若要闭合，让 telegram channel 启动时把 `FILES_SUBDIR` 注册进 `ensureInTreeStateSelfIgnored`（或存在 telegram channel 时守卫无条件覆盖整个 `.fastagent`），而不是依赖 sessions/auth 恰好落在树内。

kid7st (commented):
Thesis: 凭证文件默认从全局改为项目级 `<dir>/.fastagent/auth.json`（`--auth-path` > `FASTAGENT_AUTH_PATH` > 项目默认，无隐式回退），`authPath` 贯穿 create/models/workspace，常量 `FASTAGENT_AUTH_PATH`→`GLOBAL_AUTH_PATH`，抽出泄漏守卫 + `~` 展开 + `login` `.env` 修复。概念清晰、命名到位、错误处理是 surface-hint-then-fail-visibly、测试都带失败路径且无 tautological/过度 mock。

一条需确认意图的项见 inline（telegram-files 泄漏缺口）。其余改动语义自洽，无发现。

core/src/channels/telegram/telegram-api.ts
kid7st: Intentional trade-off, not an accidental miss. Registering telegram-files with the guard is blocked by the engine-neutrality rule (a channel must not import `engines/pi/definition.ts`), and always-self-ignoring would rework the .env-read test's observable — both out of scope for this auth PR. So I narrowed the guard's docstring to stop claiming completeness and made this comment state the dependency, turning a hidden gap into a documented one. The residual leak is narrow (both --sessions-dir and --auth-path external + a telegram channel + committing an adapted repo) and pre-existing in spirit (before this PR, an external --sessions-dir alone skipped the self-ignore too). Closing it properly wants a neutral channel-side state-dir seam — a separate change. Resolving as accepted.

PR #98, opened 2026-07-01
fix(telegram): precise group summoning + HTML-aware message split (#3/#4/#6)

kid7st:
Implements the **route/summon** and **split** items from the Telegram review (#3, #4, #6). The streaming half (#1/#2/#5) is PR #96, separately owned — this PR does not touch it.

## #3/#4 — precise group summoning

`defaultTelegramRoute` (default group policy) changes:

- **Bare `/cmd` no longer summons in a group** — ambiguous, often meant for another bot. Private chats keep slash commands (they always answer).
- **@mention is entity-based, not a substring** — a Telegram `mention` entity that IS `@botname`, or a `bot_command` targeting it (`/cmd@botname`), in **text or a media caption**. `@botname_typo` / look-alikes can't false-match.
- **Reply arm matches THIS bot** — the bot's own id (from `getMe`) instead of "any bot in the group". Falls back to "any bot" only before `getMe` resolves / when the id is unknown.
- **The bot's own `@handle` is stripped from the prompt** (text/caption) before the model sees it — it only routed the message.

## #6 — HTML-aware split

`chunkText` no longer cuts through an HTML tag: the split point backs up before an unclosed `<`, so a chunk stops degrading to plain text just because a tag straddled the 4096 boundary. Tags that *span* a boundary still rely on the existing per-chunk plain fallback (documented).

## API

- `MessageEntity` type exported from `@kid7st/fastagent/telegram` (entities a custom route can inspect).
- `resolveBotInfo` now returns the bot's `id`.
- Public surface change → **review-tier** (waits for review per CONTRIBUTING).

## Coordination with PR #96

`chunkText` (in `telegram-api.ts`) is the one function both PRs touch — #96 only adds `export` to it, this PR makes its body tag-aware. A trivial rebase merge on that one function when the second of the two lands. Everything else is in functions #96 leaves untouched (`defaultTelegramRoute`, `telegramEnvelope`, `resolveBotInfo`).

## Test

`cd core && npm run lint && npm run typecheck && npm test` — 265 pass, lint + typecheck clean. New coverage: entity mention (incl. caption), `/cmd@bot` vs `/cmd@othbot`, look-alike username, bare `/cmd` no-summon, reply-to-this-bot vs another (unit + channel `getMe`→id→route chain), handle strip (mid/end/double-mention/caption), chunkText tag-safety incl. the closed-tag "don't rewind" case.

### DIFF
--- a/core/src/channels/telegram/telegram.ts
+++ b/core/src/channels/telegram/telegram.ts
@@ -75,12 +75,25 @@ function summarizeArgs(args: Json): string {
   return values.length > 0 ? clip(JSON.stringify(args)) : "";
 }
 
+/** A Telegram message entity (subset): `mention` (`@username`), `bot_command` (`/cmd` or `/cmd@bot`),
+ *  a link, etc. `offset`/`length` are UTF-16 code units into the text — matching JS string slicing. */
+export interface MessageEntity {
+  type: string;
+  offset: number;
+  length: number;
+  [k: string]: unknown;
+}
+
 /** A Telegram message (the common subset; `[k]` keeps the rest reachable without a types dependency). */
 export interface TelegramMessage {
   message_id: number;
   text?: string;
+  /** Entities over `text` (mentions, commands, links…) — used to detect a bot @mention precisely. */
+  entities?: MessageEntity[];
   /** Caption on a media message (photo/document/…) — often the user's instruction for the attachment. */
   caption?: string;
+  /** Entities over `caption` (same shape as {@link entities}). */
+  caption_entities?: MessageEntity[];
   /** Photo sizes, smallest → largest. The channel sends the largest to the model as a vision image. */
   photo?: { file_id: string; file_unique_id: string; width: number; height: number; file_size?: number }[];
   /** Structured payloads worth rendering into the prompt as text (no new modality needed). */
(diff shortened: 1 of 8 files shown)


core/src/channels/telegram/telegram.ts
```diff
@@ -501,7 +573,9 @@ export function telegramEnvelope(m: TelegramMessage): string {
   const replyTo = r
     ? `\n[in reply to ${fromLabel(r.from) ?? `msg ${r.message_id}`} (msg ${r.message_id}): ${(r.text ?? r.caption ?? attachmentSummary(r) ?? "(empty)").slice(0, 280)}]`
     : "";
-  const parts = [m.text ?? m.caption ?? attachmentSummary(m) ?? ""];
+  const { text: raw, entities } = textWithEntities(m);
+  const body = raw !== undefined && options?.botUsername ? stripBotHandle(raw, entities, options.botUsername) : raw;
```
kid7st, line 579: Silent decision + a getMe-timing race. `stripBotHandle` ends with `return out.trim()`, so the envelope body is trimmed ONLY when `botUsername` is known, left raw otherwise. Since `mentionName` resolves asynchronously via getMe, two identical group messages — one before getMe lands, one after — produce different prompts (untrimmed vs trimmed). This trim is a side effect of the strip beyond what the function name promises (it should remove the bot's handle, not trim the whole body), and the behavior change isn't in the PR narrative/CHANGELOG.

Direction: pick one — either trim the body unconditionally in `telegramEnvelope` (decoupled from `botUsername`, kills the race), or make `stripBotHandle` touch only the handle + its adjacent space and leave leading/trailing whitespace to the envelope. Don't let the trim depend on whether getMe has resolved.

kid7st (commented):
Thesis: precise entity-based group summoning (exact @handle / /cmd@bot, reply-to-THIS-bot by id, own-handle stripping) + HTML-tag-aware chunk split. Clean and well-documented — breaking change has CHANGELOG+Migration+docs, public surface flagged review-tier, tests cover the fail paths (look-alike username, /cmd@othbot, reply-to-another-bot, closed-tag-no-rewind). One silent-decision item below; nothing else to flag.

core/src/channels/telegram/telegram.ts
kid7st: Fixed in the latest push (option b). Dropped the whole-body `.trim()` from `stripBotHandle` — the per-cut adjacent-space consumption already prevents strip-induced gaps, so the trim only ever affected pre-existing leading/trailing whitespace, and only on the strip path (when botUsername/getMe had resolved). Now the strip touches only the handle + one adjacent space; pre-existing whitespace is left alone, consistent with the non-strip (`raw`) path, so the getMe-timing race is gone. Docstring updated to state the deliberate no-trim.

kid7st:
Superseded: #3/#4 (group summoning + boundary @mention) landed in #99 (the decided version — boundary regex, mention-strip explored and dropped as semantically unresolvable). Only #6 (HTML-aware chunk split) is additive; extracting it into a focused PR. Closing this to avoid the CONFLICTING #3/#4 duplication.

PR #100, opened 2026-07-02
feat(telegram): HTML-aware message split — balance tags across chunks

kid7st:
A long reply (>4096 chars) was split at a newline or the hard limit, which could split an HTML
tag or, more commonly, cut a `<pre>` code block so the first chunk's tag was left open —
Telegram then rejects the chunk's markup and the whole chunk degrades to a plain-text fallback,
losing formatting for content that was otherwise fine.

chunkText is now a tag-balancer for HTML: a tag that SPANS a boundary is closed at the chunk's
end and reopened at the next chunk's start, with its attributes preserved (so a spanning
`<a href=…>` / `<code class=…>` keeps them), so every chunk is self-contained valid HTML. It also
never cuts THROUGH a tag token (backs the cut up before a `<` it would land inside). Plain text
(html:false) is unaffected — a `<` there is literal content — and the html flag now gates the
tag logic just as it gates parse_mode in sendMessage.

The stack (`unclosedTags`) is a flat name-matched stack, which fits Telegram's shallow HTML
subset; a boundary reserves TAG_CLOSE_BUDGET for the closers, and the loop always advances ≥1
char (no empty chunk / infinite loop even on a pathological head-`<`).

(Extracted from the closed #98, then reworked into a real balancer after research showed the
reliable approaches are entity-based splitting — which would mean changing the send model — or
exactly this close/reopen-across-chunks pattern. chunkText was already exported by #96.)
Tests: spanning-`<pre>` close/reopen with size caps, attribute preservation, no mid-token cut +
head-`<` progress, and the plain path. CHANGELOG + docs/telegram.md updated. 271 tests green.

### DIFF
--- a/core/src/channels/telegram/telegram-api.ts
+++ b/core/src/channels/telegram/telegram-api.ts
@@ -10,7 +10,7 @@ import { setTimeout as sleep } from "node:timers/promises";
 import type { ImageRef } from "../../agent.ts";
 
 /** Telegram's hard text limit per message. */
-const TELEGRAM_MAX_TEXT = 4096;
+export const TELEGRAM_MAX_TEXT = 4096;
 const PARSE_ERROR = /can't parse|entit|unsupported|unclosed|tag/i;
 
 /** Download sanity cap (Telegram's own getFile limit); a larger file/image is rejected visibly. The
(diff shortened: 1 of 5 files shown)


kid7st (commented):
Thesis: chunkText becomes an HTML tag-balancer — a tag spanning a chunk boundary is closed at the chunk end and reopened (attributes preserved) at the next chunk head, gated by the same `html` flag that gates parse_mode. Findings: 2 inline + 1 below.

**`core/src/channels/telegram/telegram.ts:226` (outside this diff) — the html send path answers "fits in one message?" with plain-mode chunkText.** The PR threads the `html` flag through every chunkText semantic branch, but this call site was not updated: finalize uses `chunkText(text)` (plain) to decide whether `editMessageText(…, { html })` can be used. Today the result is equivalent — `length === 1` equals `text.length <= 4096` in both modes via the early return — but that equivalence is accidental, resting on the undeclared invariant that html mode never produces a single chunk for over-limit text. The real semantics here is a length check. Direction: compare `text.length` against the 4096 constant directly (export it), or at minimum pass `{ html }` so the decision and the send use the same mode.

core/src/channels/telegram/telegram-api.ts
```diff
@@ -68,25 +68,93 @@ async function callBotApi(
...
+    if (cut <= 0) cut = room; // no newline in range → hard cut at the limit
+    if (opts.html) {
+      // Don't cut through a tag token: if the last `<` before the cut has no `>` after it, back up to it
+      // (lt > 0 keeps at least one content char, so the loop still progresses).
+      const lt = rest.lastIndexOf("<", cut - 1);
+      if (lt > 0 && lt > rest.lastIndexOf(">", cut - 1)) cut = lt;
+      // Likewise don't split an entity token (`&amp;`): entities are short, so a `&` within ~12 chars of
+      // the cut with no `;` yet means the cut is inside one — back up to that `&`.
+      const amp = rest.lastIndexOf("&", cut - 1);
+      if (amp > 0 && cut - amp < 12 && amp > rest.lastIndexOf(";", cut - 1)) cut = amp;
```
kid7st: The entity back-up rule can move the cut INTO a tag token, breaking this module's own "never cuts THROUGH a tag token" invariant. A raw `&` inside an attribute value is legal and common (`<a href="https://x.com/?a&b">` — any link with query params). If the cut lands within 12 chars after that `&` (e.g. just past the opening tag's `>`), `amp > lastIndexOf(";")` holds and `cut = amp`, so the chunk ends with `<a href="https://x.com/?a`. `unclosedTags` requires a closing `>` and never sees the partial tag: no closers are appended, the next chunk starts with `&b">…`, attribute debris leaks as literal text, and both chunks are invalid HTML. The tag-token check (the `lt` rule) runs BEFORE this back-up with the old cut, so it cannot catch the new one.

Direction: re-run the tag-token check after the entity back-up (or iterate both rules to a fixpoint); alternatively guard the amp rule with "this `&` is not inside a tag token" (compare against the last `<`/`>` positions before it).

core/src/channels/telegram/telegram-api.ts
```diff
@@ -96,22 +164,34 @@ export async function sendMessage(
...
+    if (mode && !result.ok && PARSE_ERROR.test(result.description)) {
+      if (first) {
+        // Nothing sent yet — re-chunk the whole body as plain (no injected boundary tags leak) and restart.
+        mode = false;
+        chunks = chunkText(body, { html: false });
+        i = -1;
+        continue;
+      }
+      // A later chunk failed after the first parsed cleanly (rare) — best-effort plain resend of this chunk.
       result = await callBotApi(api, botToken, "sendMessage", base);
```
kid7st, line 202: This later-chunk fallback branch is untested. Its behaviour is the PR's named trade-off — a non-first chunk that fails to parse is resent as plain with the injected boundary tags (`<pre>`/`</pre>`) leaking as literal text — but no test pins it: the existing fallback test's mock rejects ALL HTML sends, so it only exercises the first-chunk restart path. The newest, most subtle branch on the failure path has no protection; a future edit (e.g. someone "fixing" it into a whole-body resend) passes the suite silently.

Direction: add a case where the first HTML chunk succeeds and the second is rejected, asserting the second chunk is resent as plain byte-for-byte (boundary tags included) — freeze the trade-off as spec.

core/src/channels/telegram/telegram-api.ts
kid7st: Fixed in 790d722: the amp rule is now guarded with "this `&` is CONTENT, not inside a tag token" (`lastIndexOf("<", amp) <= lastIndexOf(">", amp)`). Added a test with an `href` carrying query params (`?a&b`) positioned so the unguarded rule would cut into the tag — verified it fails without the guard (chunk0 ended `<a href="https://x.com/?a`) and passes with it.

core/src/channels/telegram/telegram-api.ts
kid7st: Fixed in 790d722: added a test where only the SECOND HTML chunk is rejected — asserts exactly one plain resend, byte-for-byte equal to the failed HTML chunk (injected boundary `<pre>` included), freezing the documented trade-off.

kid7st:
Review-body finding (finalize using plain-mode `chunkText` to decide single-message) fixed in 790d722: the decision is now an explicit `text.length <= TELEGRAM_MAX_TEXT` (the constant is exported) — the real semantics, no reliance on the accidental mode-equivalence.

PR #101, opened 2026-07-02
fix(telegram): transport timeouts + honest flood-wait handling

kid7st:
The Bot API transport had long-run reliability gaps (found comparing against what a bot
framework hardens for): no fetch timeout anywhere — a wedged connection hangs the turn, the
per-session serial queue behind it, or `--tunnel` webhook registration, forever — an unbounded
429 sleep (Telegram's retry_after reaches minutes–hours on a flood ban, silently parking a turn
and its queue), no retry at all on a 429 *without* retry_after, and two fake-success holes: a
timeout mid-body-read was swallowed by `.json().catch(() => ({}))` into `ok:true`, and a proxy's
HTTP 200 + error page read as a sent message.

Every fetch now carries AbortSignal.timeout — API_TIMEOUT_MS=30s for JSON calls (callBotApi,
getMe, getFile, setWebhook) and DOWNLOAD_TIMEOUT_MS=120s for file bytes and the scaffold
send-tool's uploads. Failures are named and self-describing: fetch/body-read errors rethrow as
`telegram <method>: …` (not a bare TimeoutError); the body is read via text() inside the guarded
region and only the JSON *parse* is forgiven; success requires the body's own `ok:true`, not
just HTTP 200; a flood ban beyond FLOOD_WAIT_MAX_S=30 refuses to wait and says so; an exhausted
429 says "gave up after N retries" — each transport decision is distinguishable in the log.

429 handling: retry_after is honoured up to the cap; a missing retry_after gets a short linear
backoff (bounded by the same attempt < 3). Deliberately NO new retry classes for 5xx/network/
timeout errors: the request may have been processed, and a retried sendMessage would
double-deliver — those keep failing visibly for the caller to surface.

The retry sleep moved off node:timers/promises onto the global timer so fake timers can drive
it. Tests pin the mechanisms, not just shapes: timeout signal identity (30s send / 120s
download), both 429 retries actually WAIT (two-phase clock advance), retry exhaustion is bounded
(4 calls) and labeled, the flood-cap fast-fail, named throws (fetch reject + mid-body reject),
and the 200-with-garbage named failure. 287 tests green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ While the project is pre-1.0, minor versions may include breaking changes.
   (`github`, `telegram`, `channel-development`), and maintainer notes under `docs/design/`.
 
 ### Changed
+- Telegram transport hardening: every Bot API call, file download, and the `--tunnel` webhook registration now carry a timeout (30s API / 120s download) so a wedged connection cannot hang a turn, its session queue, or dev startup; a 429 without `retry_after` gets a short backoff (the scaffold send-tool is the one exception — it stays fail-fast so the agent sees the error and decides); a flood ban demanding a wait beyond 30s fails visibly instead of silently parking the turn. Failures self-describe (`telegram <method>: …`, "gave up after N retries", the flood-wait cap), a success requires the body's own `ok:true` — a proxy's 200 + error page can no longer read as a sent message.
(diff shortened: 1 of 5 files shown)


kid7st:
Superseded by a first-principles redesign: the transport invariants here (timeouts, bounded flood-wait, ok:true validation, self-describing failures) get carried structurally by a single-pipeline rewrite of telegram-api.ts (gramIO's architecture lesson: one generic call function, zero per-method code, errors as typed objects). Landing them per-call-site first would mean merging code the rewrite deletes.

PR #103, opened 2026-07-02
fix(telegram): detect @mention summon from Telegram's mention entities

kid7st:
mentionsBot scanned the raw text with a boundary-anchored regex — re-deriving structure the platform
already hands us: every message carries `entities` (`caption_entities` for captions) that Telegram's
server parsed, with exact UTF-16 offsets (= native JS string indexing). The regex's fix history was
the tell: each iteration (includes → boundary anchors → command lookbehind) patched a false-match
class that the entity TYPE excludes by construction — and two classes were still live: `@botname`
pasted inside a code block or a URL is not a `mention` entity but matched the regex (a false summon
waiting to happen).

mentionsBot now takes the message and matches `mention` entities against the bot name by exact
offset/length slice, case-insensitive. `/cmd@bot` stays excluded (its entity is `bot_command`),
`@fast` vs `@fastagent` confusion is impossible (exact range), and code/URL false positives are gone.
Deliberately NO text fallback: mention entities are produced server-side, so their absence means
there is no mention — a regex fallback would re-admit the false-positive classes this removes.

TelegramMessage gains `entities`/`caption_entities`. Route tests now build the payload shape Telegram
actually sends (mention/bot_command entities) and pin the two new exclusion classes; the regex-specific
boundary cases are superseded by entity-range cases. docs/telegram.md + CHANGELOG updated.


kid7st (commented):
No findings.

PR #105, opened 2026-07-02
fix/telegram summon precision

kid7st:
- **chore(telegram): drop callback_query from TelegramUpdate — a promise the channel can't keep**
- **fix(telegram): group summon precision — reply-to-THIS-bot by token id, and edits never summon**


kid7st (commented):
No findings.

kid7st (commented):
No findings.

PR #111, opened 2026-07-03
chore: surface hygiene — unexport TelegramApiError, annotate the scaffold rollback

kid7st:
Two local findings from the structural scan:

- TelegramApiError was exported with zero importers: the public facade (src/telegram.ts) does not
  re-export it and no caller matches on the type — a half-public surface. It is module-private
  now, with a note to export it THROUGH the facade if a custom channel ever needs typed matching
  (not to leave it dangling in between).
- scaffold/init.ts's rollback swallowed rm() failures without a word; the swallow is correct (a
  cleanup failure must not mask the original scaffold error) but now says so.

### DIFF
--- a/core/src/channels/telegram/telegram-api.ts
+++ b/core/src/channels/telegram/telegram-api.ts
@@ -70,8 +70,11 @@ interface Api {
 
 /** A named Bot API failure. `status` 0 = the transport itself failed (network error / timeout) before
  *  any HTTP status existed; otherwise the HTTP status with Telegram's own description (or a note that
- *  the body carried no usable answer). */
-export class TelegramApiError extends Error {
+ *  the body carried no usable answer). Module-private: no EXTERNAL caller matches on the type; it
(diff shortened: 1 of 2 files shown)


kid7st (commented):
No findings.

surface hygiene 两处均核对无误:
- `TelegramApiError` 去 export 安全 —— 全部引用(3 处 `instanceof` + `.description`、5 处 `new`)都在 `telegram-api.ts` 内部,test 目录 0 引用,facade(`src/telegram.ts`)/`index.ts` 不 re-export,docs/README/AGENTS 无任何提及,去 export 不与文档矛盾。docstring 关于「保留 typed class 因内部 retry(sendMessage parse-mode fallback、editMessageText not-modified swallow)用 instanceof」的理由与 line 272/309/314 相符。
- `init.ts` rollback 注释准确:`rm(..., {force:true}).catch(()=>{})` 确实会留下删不掉的文件,`throw error` re-throw 原始错误——注释点明「cleanup 失败不得掩盖原始 scaffold 错误」正是正确的责任分配,吞错有据。

PR #114, opened 2026-07-03
fix(tunnel): never parse cloudflared's API endpoint as the assigned tunnel URL

kid7st:
Found live during the 0.8.0 release verification: under a flaky proxy, cloudflared printed an
ERROR line containing its request endpoint ("failed to request quick Tunnel: Post
https://api.trycloudflare.com/tunnel: ..."), and TUNNEL_URL_RE — any host on trycloudflare.com —
matched it. The "tunnel URL" became Cloudflare's API host, the health poll happened to pass, and
setWebhook registered the bot against api.trycloudflare.com/telegram: the bot silently received
nothing while every log line said success.

The assigned quick-tunnel URL is always a generated multi-word host; the API endpoint is always
`api.trycloudflare.com`. The regex now excludes `api.` — an error line parses as nothing (the
retry loop keeps waiting for the real assignment), and an error line followed by the real box
line still resolves to the real URL. Both pinned by tests.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,13 @@ While the project is pre-1.0, minor versions may include breaking changes.
 
 ## [Unreleased]
 
+### Fixed
+
+- **`--tunnel` no longer mistakes cloudflared's API endpoint for the assigned tunnel URL.** Under a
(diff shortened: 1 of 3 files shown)


kid7st (commented):
No findings.

PR #115, opened 2026-07-03
fix(engine): wire provider request retries — retry-capable adapters no longer fail the turn on a transient

kid7st:
The OpenAI-family / Anthropic / Azure / Codex pi-ai adapters implement client-side retries
(429/5xx/request-phase network errors with backoff, honoring Retry-After; a Codex websocket that
fails before the stream starts falls back to SSE) — but all default maxRetries to 0 (even
SDK-backed ones override the SDK default), and the bare AgentHarness sets nothing. So a single
transient `fetch failed` killed the whole turn (seen live in the 0.8.0 release verification: two
consecutive provider blips each surfaced as a failed turn that the pi CLI would have
auto-retried).

Harnesses are now built with `streamOptions: { maxRetries: 2 }` (the OpenAI/Anthropic SDK
default). Scope limits, both deliberate here:

- The google / vertex / bedrock / mistral adapters ignore `maxRetries` (a pi-ai upstream gap) —
  transients there still fail the turn. Setting the option is harmless where ignored.
- Mid-stream drops are NOT retried: partial output was already streamed and the SPEC event
  stream cannot retract it — a mid-stream failure surfaces as a `failed` event.

## Test
- `test/harness.test.ts`: the factory-built harness carries `maxRetries: 2` (`getStreamOptions()`) — the wiring contract; whether an adapter honors it is pi-ai behavior
- full suite: 327 passed

### DIFF
--- a/core/src/engines/pi/harness.ts
+++ b/core/src/engines/pi/harness.ts
@@ -35,6 +35,19 @@ export interface PiHarnessFactoryOptions {
   skills?: Skill[];
 }
 
+/**
+ * Provider request retries. The OpenAI-family / Anthropic / Azure / Codex pi-ai adapters
+ * implement client-side retries (429/5xx/request-phase network failures with backoff, honoring
+ * Retry-After; a Codex websocket that fails before the stream starts falls back to SSE) but all
+ * default maxRetries to 0 (even SDK-backed ones override the SDK default), so a single transient
+ * `fetch failed` would otherwise kill the whole turn; 2 matches the OpenAI/Anthropic SDK default.
+ * The google / vertex / bedrock / mistral adapters ignore this option (a pi-ai upstream gap) —
+ * transients there still fail the turn. Mid-stream drops are deliberately NOT retried anywhere on
+ * this path — partial output was already streamed and the SPEC event stream cannot retract it,
+ * so a mid-stream failure surfaces as a `failed` event.
+ */
+const PROVIDER_MAX_RETRIES = 2;
+
 /** Open-or-create the session per invoke: existing → open (history via buildContext); missing → create. */
 export function piHarnessFactory(options: PiHarnessFactoryOptions): PiHarnessFactory {
   return async (sessionId) => {
@@ -48,6 +61,7 @@ export function piHarnessFactory(options: PiHarnessFactoryOptions): PiHarnessFac
       tools: options.tools,
       systemPrompt: typeof systemPrompt === "function" ? systemPrompt() : systemPrompt,
       resources: options.skills ? { skills: options.skills } : undefined,
+      streamOptions: { maxRetries: PROVIDER_MAX_RETRIES },
     });
   };
 }
--- a/core/test/harness.test.ts
+++ b/core/test/harness.test.ts
@@ -1,0 +1,24 @@
+import { describe, expect, it } from "vitest";
+import { NodeExecutionEnv } from "@earendil-works/pi-agent-core/node";
+import { inMemorySessionStore } from "../src/index.ts";
+import { piHarnessFactory } from "../src/engines/pi/harness.ts";
+import { makeFaux } from "./faux.ts";
+
+describe("piHarnessFactory: provider retry wiring", () => {
+  it("built harnesses carry maxRetries — the wiring retry-capable pi-ai adapters honor", async () => {
+    // Retry-capable adapters (OpenAI-family / Anthropic / Azure / Codex) default to 0 retries —
+    // this wiring enables their client-side retries (a transient `fetch failed` killed a whole
+    // turn live in the 0.8.0 release verification). google / vertex / bedrock / mistral ignore
+    // the option; transients there still fail the turn.
+    const { faux, models } = makeFaux();
+    const factory = piHarnessFactory({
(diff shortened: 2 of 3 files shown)


core/src/engines/pi/harness.ts
```diff
@@ -35,6 +35,17 @@ export interface PiHarnessFactoryOptions {
   skills?: Skill[];
 }
 
+/**
+ * Provider request retries (pi-ai adapters retry 429/5xx/request-phase network failures with
+ * backoff, honoring Retry-After; a websocket that fails before the stream starts falls back to
+ * SSE). pi-ai adapters all default maxRetries to 0 (even SDK-backed ones override the SDK
```
kid7st: This comment (and the CHANGELOG entry, `CHANGELOG.md:13-16`) over-claims provider coverage. Only the openai-completions / openai-responses / azure / anthropic-messages / openai-codex-responses / openrouter-images adapters consume `maxRetries`; `google-generative-ai.js`, `google-vertex.js`, `bedrock-converse-stream.js`, and `mistral-conversations.js` never reference the option (mistral explicitly pins `retries: { strategy: "none" }`). Since `createPiModels()` registers ALL builtin providers, for Gemini/Vertex/Bedrock/Mistral this fix is a no-op — a transient `fetch failed` still kills the turn — while the CHANGELOG tells users "Transient provider request failures no longer kill the turn" unconditionally. "even SDK-backed ones override the SDK default" also implies universal support, which is not the case.

Suggested direction: scope both texts to the actual coverage, e.g. "providers whose adapters implement client-side retries (OpenAI-family / Anthropic / Azure / Codex); google / vertex / bedrock / mistral adapters ignore `maxRetries`, transients there still fail the turn". The coverage gap itself is a pi-ai upstream matter — not for this PR — but the claim must match the behavior.

kid7st (commented):
Thesis: inject `streamOptions.maxRetries: 2` into every factory-built `AgentHarness` so request-phase transient provider failures are retried at the pi-ai adapter layer. Wiring verified end-to-end (`createStreamFn` forwards `maxRetries` to `models.streamSimple`; harness does not restore stale streamOptions from the session). One finding, inline on harness.ts (also applies to CHANGELOG.md:13-16).

core/src/engines/pi/harness.ts
kid7st: Fixed in 2437c11 — both texts now name the exact coverage: OpenAI-family / Anthropic / Azure / Codex adapters implement client-side retries (all defaulting to 0); google / vertex / bedrock / mistral ignore `maxRetries` (pi-ai upstream gap), transients there still fail the turn. No code change: setting `maxRetries: 2` is harmless where ignored.

core/test/harness.test.ts
```diff
@@ -0,0 +1,22 @@
+import { describe, expect, it } from "vitest";
+import { NodeExecutionEnv } from "@earendil-works/pi-agent-core/node";
+import { inMemorySessionStore } from "../src/index.ts";
+import { piHarnessFactory } from "../src/engines/pi/harness.ts";
+import { makeFaux } from "./faux.ts";
+
+describe("piHarnessFactory: provider retry wiring", () => {
```
kid7st, line 7: The test name and following comment still express the old global-provider retry semantics: "request-phase transients are retried, not turn-fatal" / "pi's raw-fetch adapters default to 0 retries". The implementation only means "factory-built harnesses carry `maxRetries: 2`, which retry-capable pi-ai adapters honor"; google / vertex / bedrock / mistral ignore the option and transients there still fail the turn.

Suggested direction: rename this test/comment to the wiring contract it actually verifies, e.g. "built harnesses carry maxRetries for pi-ai adapters that honor provider retries" and "retry-capable adapters default to 0; this wiring enables their client-side retries".

kid7st (commented):
Thesis: this PR wires `streamOptions.maxRetries: 2` into factory-built pi harnesses; after the last update, code comments and CHANGELOG correctly scope the behavior to pi-ai adapters that honor `maxRetries`.

Unanchored finding: the PR description still uses the old global wording: "pi-ai's adapters have a built-in retry loop" / "all pi-ai adapters previously defaulted to maxRetries 0". That conflicts with the implementation's scoped semantics, because google / vertex / bedrock / mistral ignore `maxRetries`. Update the PR description to the same scoped language as the code and CHANGELOG.

core/test/harness.test.ts
kid7st: Fixed in f167655 — test name now states the wiring contract ("built harnesses carry maxRetries — the wiring retry-capable pi-ai adapters honor"), and the comment names the retry-capable set vs the adapters that ignore the option. PR description and commit message updated to the same scoped wording.

kid7st (commented):
Finding: the PR title still preserves the old unscoped semantics. The body, CHANGELOG, code comment, and test now correctly say `maxRetries` only helps retry-capable pi-ai adapters, while google / vertex / bedrock / mistral still fail transients. But the title says `a transient fetch failure no longer kills the turn`, which reads as global provider behavior. Rename it to the scoped concept, e.g. `fix(engine): wire retries for retry-capable pi providers` or `fix(engine): enable provider retries where pi-ai adapters honor maxRetries`.

PR #118, opened 2026-07-04
refactor: flatten core/ to the repository root — single package, standard OSS shape

kid7st:
The core/ nesting was a pre-paid boundary for sibling packages that, on re-evaluation, are not
coming: channels are zero-dependency subpath exports (the Hono model), so the module boundary is
the exports map, not a workspace. Meanwhile the nesting cost was permanent — `cd core` in every
command, a second README/LICENSE to keep in sync, working-directory special cases in both
workflows, and "why core/?" as every newcomer's first question.

Flattened while the repo is still private (zero external clones affected; the npm tarball is
byte-identical in file count — the `files` whitelist is unchanged). The root README becomes the
npm README, absorbing core/README's one unique section (the public-surface stability table); the
duplicate LICENSE is gone; CI/publish lose their working-directory overrides; AGENTS.md /
CONTRIBUTING / docs paths updated. start.md also moves to docs/ai-start.md — at the root it read
like "how to start the service"; it is an AI-guided onboarding prompt.

A packages/ workspace split stays deliberately deferred until a second published artifact with
independent dependencies/versioning actually exists (documented in README "Repository layout").


package-lock.json
```diff
@@ -1256,7 +1256,7 @@
         "typebox": "1.1.38"
       },
       "bin": {
-        "pi-ai": "./dist/cli.js"
+        "pi-ai": "dist/cli.js"
```
kid7st: Drive-by unrelated to the flatten: `"./dist/cli.js"` → `"dist/cli.js"` is a transitive dep (pi-ai) bin-path normalization from a different npm version regenerating the lockfile, not a core/ rename. Note line 750 has the same entry unchanged — partial re-gen noise. Revert this line (`git checkout main -- package-lock.json`, keep only the rename) so the PR carries only the flatten delta.

kid7st (commented):
Thesis: mechanical flatten of `core/*` to the repo root — single-package standard OSS shape; near-all files are 0-change renames, substance is path updates in config/CI/docs. Typecheck passes, no content loss (root README already carries Install/CLI/Library API; it absorbs core/README's public-surface table). Two findings below.

**.github/workflows/ci.yml:45-46 — the `Core checks` name is a residue of the concept this PR removes.** The aggregate gate keeps job id `core-checks` / name `Core checks`. With `core/` gone, a newcomer reads it as pointing at the deleted directory — the exact "why core/?" question this PR set out to kill. It's deliberately the branch-protection required check, so renaming means also updating the branch-protection config (an admin action outside this diff). Decide here: either rename + update branch protection, or state explicitly in the PR that the name is intentionally kept.

package-lock.json
kid7st: Fixed in 480021c — lockfile bytes restored from main's core/package-lock.json, so it is now a pure rename with zero content delta (verified: npm ci + full suite green with the original bin-path form).

kid7st:
On the `Core checks` name (review body finding): **kept, deliberately.** Renaming the branch-protection required check costs a two-step protection migration (the renamed context can't satisfy the old required name, deadlocking this PR's own merge) for marginal clarity — and the misreading dies with one comment line instead: 480021c adds "(\"Core\" = the essential aggregate gate, not the removed core/ directory.)" to ci.yml.

kid7st (commented):
No findings. Re-review of the mechanical flatten (core/* → root, single package) after both prior findings were addressed: the package-lock pi-ai bin drive-by is gone, and the `Core checks` name now carries an inline comment clarifying it is the essential aggregate gate, not the removed core/ directory (name kept stable so branch protection needn't change). lint + typecheck clean, no stray core/ path refs, no content loss.

kid7st (commented):
No findings. Review of the new `feat(dev): live definition + narrow watch` commit on top of the (already-clean) flatten. Thesis: split dev/serving change classes by whether they need a new process — definition (AGENTS.md + skills/) becomes per-invoke live re-read via a new `live` source on the harness factory (one read yields prompt AND skills), and the watcher narrows to code inputs (tools/, channels/, config, package.json, .env) so the agent's own workspace writes never SIGTERM its in-flight turn.

Verified sound: (1) error responsibility — a `live()` throw rejects the factory and is caught in invoke.ts's factory try/catch into a diagnostic `failed` event (SPEC MUST 2), never a process crash; non-fatal loader findings warn once via `findingsSignature` dedup. (2) Production `start` also going live is not a silent decision — CHANGELOG names the scope (`createPiAgentFromDefinition, dev, start`) and the `createPiAgent`/`chat` exceptions. (3) The watch allowlist matches what is actually loaded — `.env` only, and `process.loadEnvFile` reads only `.env` (no cascade), no missed input. Tests cover happy path, bad-skill dedup, throw-class recovery, and the watch allowlist boundaries; 24 new tests pass locally.

PR #122, opened 2026-07-04
fix(auth): drop the pre-0.x global-credential migration hint

kid7st:
## Summary

The startup auth report probed `GLOBAL_AUTH_PATH` whenever the project-level auth file was empty and pointed a pre-0.x upgrader at it (`set FASTAGENT_AUTH_PATH=…`). That was a backward-compat aid with no place in a pre-release — removed.

Deleted along with it: the now-dead `GLOBAL_AUTH_PATH` import in `cli.ts`, the one test that covered the hint, and the stale companion comments (the `invoke` call-site note + the migration block's own). The report falls back to the generic `fastagent login` hint, consistent with the fail-visibly, no-implicit project/global fallback design (`auth.ts` / `docs/design/core.md` §10.4).

Net: source -16, tests -21.

## Validation

- [x] `npm run typecheck`
- [x] `npm test`
- [x] Added or updated the smallest relevant tests (removed the test for the deleted behavior)

## Checklist

- [x] Docs were updated when behavior or public APIs changed (none needed — core.md §10.4 already describes the no-fallback design this aligns to)

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -30,7 +30,6 @@ import {
   resolveSessionsDirOverride,
 } from "./engines/pi/config.ts";
 import { formatModelsCommand } from "./cli-models.ts";
-import { GLOBAL_AUTH_PATH } from "./engines/pi/auth.ts";
 import { type LoginIO, loginFlow } from "./engines/pi/login.ts";
 import { createPiModels, probeAuthSource } from "./engines/pi/models.ts";
 import { ensureInTreeStateSelfIgnored, loadAgentDefinition } from "./engines/pi/definition.ts";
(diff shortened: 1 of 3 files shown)


kid7st (commented):
No findings. The PR removes the pre-0.x global-credential migration hint from the startup auth report, falling back to the generic `fastagent login` hint — consistent with the no-implicit-fallback / fail-visibly design in core.md §10.4. Verified: the `GLOBAL_AUTH_PATH` removal in cli.ts is complete (all import + if-block usages gone, `provider` still used, no unused import), and the symbol is not orphaned — login.ts and the index.ts public export still use it legitimately, so this correctly deletes only the cli.ts call site. Companion text is cleaned in the same pass (both stale migration comments removed; config.test.ts comment retied to the startup report with its assertion untouched). No docs reference the removed hint (grep empty), matching the PR's claim. The deleted test covered exactly the removed branch, so dropping it loses no protection for surviving behavior.

PR #123, opened 2026-07-04
refactor(telegram): drop the turn WAL — in-memory per-session queue

kid7st:
## Summary

Replaces the telegram turn store's file-backed WAL with an in-memory per-session queue. The load-bearing feature — one turn at a time per session (FIFO), different sessions concurrent — is unchanged; only the restart-durability layer is removed.

**Why.** The WAL (recovery of `queued` turns, `started` loud-drop, redelivery tombstones, shape validation) was a partial durability layer bolted onto a feature that shipped in-memory first and never depended on it. Telegram already redelivers *never-ACKed* updates on its own; the only window a WAL added was an *ACKed-but-unfinished* turn lost to a crash — a window a single-process bot accepts (the asker re-asks), and whose proper fix is durable execution at a K-axis backend (an external queue with distributed-locking recovery, SPEC §11), not a hand-rolled per-process log in the channel. This also makes turn durability consistent with `start`'s existing no-graceful-drain on SIGTERM.

**Cascade.** No recovery ⇒ no replay ⇒ the tombstone/`suppressed` dedup (whose only job was to stop a WAL-replay from double-running a Telegram redelivery) is unnecessary too. `turn-store.ts` (186 lines) becomes `turn-queue.ts` (~40) — the in-memory FIFO chains it always had. `telegram.ts` sheds `isPendingTurn`, the `suppressed` gate, and the persisted-record fields; the ⏳ queued-notice id is now a direct mutation on the shared record (gated by the same dequeue await), not a store-owned patch.

**Kept.** The un-summoned group context buffer stays durable — its case is stronger (those messages are never redelivered as a summon).

Net **-317 lines**.

## Behavior change (documented)

Turn durability is now deliberately partial: a crash between the webhook 200 and turn completion loses that turn (re-ask). Recorded in `CHANGELOG.md` (with an upgrade note: a prior `queue.json` is no longer read — a turn queued-but-not-started at the upgrade moment is dropped rather than replayed; the now-inert file under the gitignored `.fastagent/channels/telegram/` can be deleted), `docs/design/core.md` §9.2, `docs/telegram.md`, and the module/`AGENTS.md` maps.

## Validation

- [x] `npm run typecheck`
- [x] `npm test` (326 passing; WAL recovery matrix removed, FIFO/concurrency/queue-notice kept, `turn-queue.test.ts` covers the in-memory queue incl. the runner-rejection backstop)

## Checklist


kid7st (commented):
No findings. Replaces the telegram file-backed WAL turn-store (186 lines) with an in-memory `turn-queue.ts` (~40), keeping the load-bearing per-session FIFO + cross-session concurrency and dropping only restart-durability (plus its cascade: `suppressed` dedup, `isPendingTurn` validation, `state`, persisted-record fields).

The one real risk — dropping `suppressed` causing a duplicate reply — was checked against the old turn-store from main and does NOT regress: the old `suppressed(id)` only queried `tombstones`, which were populated ONLY at construction-time recovery from a prior process's WAL (empty in same-process steady state), so it guarded solely against post-restart self-replay colliding with a redelivery. With no WAL there is no self-replay, and because `accept` schedules synchronously then returns 200 while the turn runs async at dequeue, output always follows the 200 — Telegram never redelivers a turn that produced output, so no double-answer is possible. The crash cases are handled (ACKed-then-lost = loss/re-ask, documented; never-ACKed = redelivered once). The pre-existing 200-lost-in-transit-without-crash double-run edge is unchanged (the old suppressed never covered it either).

Error handling is sound (runner owns its surface; an escaped rejection is caught and logged loudly with session context, never silent). `turn-queue.test.ts` is non-tautological and covers failure paths — notably the runner-rejection case pins the "never unhandled rejection" half with a real `process.on('unhandledRejection')` listener and separately asserts the catch is loud, explicitly noting the chain-continues assertions can't stand in for it. Deleted tests correspond exactly to removed behavior; docs/CHANGELOG/AGENTS/state.ts comments all updated, with the behavior change and the inert `queue.json` upgrade note documented. CHANGELOG's older entry describing the WAL is correctly left as history.

PR #124, opened 2026-07-04
refactor(telegram): tidy after the decompose — bodyOf, split invokeWithAttachments, co-locate transport tests

kid7st:
Post-refactor structural scan of `src/channels/telegram/` (the churn hotspot; 3 recent structural refactors — single transport pipeline, subsystem decompose, WAL removal). Structure judged **sound** — the big corrections already landed. This PR is the second-pass local cleanup only; no redesign.

## Changes

- **`bodyOf()`** — the `text ?? caption ?? attachmentSummary` fallback was written 3× (messageText, envelope reply-quote, envelope body), each with its own tail default. One helper, fallbacks can't drift.
- **split `invokeWithAttachments`** — it fused attachment resolution + manifest building + prompt assembly + the invoke stream behind a **12-positional-parameter** signature. Extract `resolveTurnAttachments -> {images, promptSuffix}` (pure, directly testable); `invokeTurn` is now a thin stream wrapper. Params grouped into `TurnTransport` / `TurnAttachments`. Behavior unchanged; both new symbols are module-private (public surface untouched).
- **co-locate transport tests** — 3 of 4 extracted subsystems had co-located test files; telegram-api's tests stayed in the 1799-line `telegram.test.ts`. Moved chunkText / sendMessage-fallback / callApi-pipeline / file-download into `test/telegram-api.test.ts`. Pure move.

## Considered and dropped

Trimming the "duplicate" buffer tests between `telegram.test.ts` and `context-buffer.test.ts`: on inspection they are **integration** (channel orchestration, driven through the webhook) vs **unit** (`createContextBuffer` internals) — orthogonal layers, not duplication. Dropping the integration tests would lose "does the channel drive the buffer correctly" coverage.

## Verification

`npm run typecheck` clean; `npm test` 325 passing (test files 31 → 32 from the split; test count unchanged).


kid7st (commented):
No findings. Reviewed the post-decompose cleanup delta (bodyOf, invokeWithAttachments split, transport-test co-location); the stacked WAL/auth/dev commits carry their own PR reviews (#123/#122).

Verified behavior-preserving: (1) `bodyOf()` collapses the shared `text ?? caption ?? attachmentSummary` chain into one helper returning undefined, while each caller keeps its own caller-specific tail default (`"(empty)"` for the reply-quote vs `""` elsewhere) — no over-consolidation, no drift. (2) The `invokeWithAttachments` split preserves the two-tier failure policy exactly: primary `resolveImages/resolveFiles` throw out of `resolveTurnAttachments` and are caught by `invokeTurn`'s try into a `failed` event (never a silent drop), while buffered attachments still degrade per-attachment via allSettled; the prompt assembly order (`text + bufferedNote + manifest + HTML_INSTRUCTION`) and the empty-images→undefined mapping are unchanged; the 12-positional signature becomes `TurnTransport`/`TurnAttachments`, and the old docstring's onCompleted commit-point note is correctly redistributed to `invokeTurn`. (3) The test move is pure — removed and added describe/it blocks match name-for-name and order (chunkText/sendMessage-fallback/callApi-pipeline/file-download), nothing lost, 325 passing. `resolveTurnAttachments` is called "pure" in the PR prose (it does IO) but the code docstring makes no such claim; the extraction stays covered by the existing integration tests through `invokeTurn`, so no protection is lost by not adding a direct unit test.

PR #126, opened 2026-07-05
feat(telegram): durable turn intent (L1 crash/deploy recovery)

kid7st:
## What

Adds **L1 durable turn intent** to the Telegram channel: an accepted turn's intent is persisted before the webhook ACK (`turns.json`) and removed when the turn ends. A turn interrupted mid-run — a crash **or** a SIGTERM deploy (`start` has no graceful drain) — is replayed on the next start, recovering the ACKed-but-unfinished window Telegram won't redeliver.

This is a deliberate reversal of 0.8.3's "accept the loss" stance. It's a per-process *intent* log made cheap by the channel's existing atomic-rename state primitive (`state.ts`) — distinct from the queue-level WAL 0.8.3 removed.

## Semantics (at-least-once, not exactly-once)

- Replay re-runs the **whole** turn → side-effecting tools re-run. Safe for a Q&A bot; the bar for adding side-effecting tools. The trigger is not rare: it fires on every deploy that interrupts an in-flight turn, so the idempotency bar is judged against deploy frequency.
- The pre-ACK window can overlap Telegram redelivery (same `update_id`, no dedup) → a turn can run twice.
- Exactly-once delivery (a persisted delivery key) and deterministic step-replay remain **L2** (a K-axis backend + engine resume hook).

## Key invariants (each with a test)

| Invariant | Mechanism |
|---|---|
| Pre-ACK durability | `add()` persists before the 200; a write failure throws → 500 → Telegram redelivers, and rolls memory back (no phantom entry) |
| Redelivery doesn't reset the ceiling | `add()` is idempotent on an existing id |
| Poison turn can't loop forever | per-turn execution ceiling, bumped at dequeue |
| A poison head doesn't penalize never-run siblings | ceiling counts per turn, not per restart cycle |
| Disk failure can't cause an untracked loop | the bump fails **closed** — defers to the next start rather than run untracked (no user notice — a transient degrade) |
| Replay is the WHOLE turn | the completed hook removes intent **before** committing the context buffer, so a crash between the two can't replay a context-stripped turn |
| Per-session FIFO on recovery | replay in `update_id` (arrival) order |
| Dropped turn is visible to the asker | notifies (taking over the ⏳ notice in place if present — no orphan/double-post) |

## Companion docs

`docs/design/core.md`, `docs/telegram.md`, `AGENTS.md` repo map, `cli.ts` comment, `CHANGELOG.md`.

## Review

Iterated to clean under isolated local review (`rv.sh local`). `npm run lint && typecheck && test` green (346 tests).

## Tier

Touches an architecture decision (reverses a documented stance) but not the SPEC / `Agent` contract / public API surface. New `turn-store.ts` is Telegram-internal.


src/channels/telegram/telegram.ts
```diff
@@ -503,9 +525,23 @@ export function telegramChannel(
...
       notices.delete(rec.id);
+      // Count this execution against the durable record (poison-turn ceiling) before running it again.
+      const decision = store.startAttempt(rec.id, MAX_TURN_ATTEMPTS);
+      if (decision === "exceeded") {
+        // Started MAX_TURN_ATTEMPTS times without finishing — tell the asker (reusing its ⏳ notice if
+        // any), drop it.
+        notifyDropped(rec);
+        return;
+      }
+      if (decision === "defer") {
```
kid7st, line 540: `defer` silently breaks per-session FIFO — the one path that inverts it, named nowhere. Defer returns and the session chain proceeds to the next queued turn: if the disk failure is transient (A's bump fails, disk recovers before B dequeues), sibling B runs and answers before deferred A, and A replays only after a restart — possibly days later, after its successor. Per-session serial FIFO is the turn-queue's entire reason to exist and the PR's invariant table claims "Per-session FIFO on recovery"; neither the defer comment nor docs/telegram.md names this inversion window, in a module that otherwise enumerates every window.

Direction: the disk-failure corner doesn't warrant chain-stalling machinery — name the reorder window explicitly in this comment and in docs/telegram.md's L1 bullet. If ordering must hold, defer would have to skip the whole session chain, not just the head.

src/channels/telegram/telegram.ts
```diff
@@ -503,9 +525,23 @@ export function telegramChannel(
...
+        // Couldn't record the attempt (disk failure): skip this cycle — a restart replays it, self-
+        // healing. No notify: telling the asker to re-ask would double-answer once the deferred turn runs.
```
kid7st: `defer` orphans the ⏳ notice — the exact orphan + double-post this design refuses elsewhere. The drop path carefully takes the notice over (the PR's own "no orphan/double-post" invariant), but defer leaves a present notice pinned at "Queued" forever; `previewId` is not persisted, so the eventual replay sends a fresh preview — old "Queued" message orphaned, two messages for one turn.

Direction: on defer with `rec.previewId` set, best-effort delete the notice (deleteMessage; the replay sends a fresh preview anyway) — or at minimum document the accepted orphan here.

src/channels/telegram/turn-store.ts
```diff
@@ -0,0 +1,186 @@
+/**
+ * Durable turn intent: the L1 half of durable execution (the L2 exactly-once / deterministic
+ * step-replay layer is the K-axis backend — docs/design/core.md §11). Persists an accepted turn BEFORE
+ * the webhook 200 (pre-ACK, like context-buffer) and removes it when the turn ENDS — the runner's
```
kid7st, line 4: "removed when the turn ENDS" overstates L1's coverage: removal fires at the `completed` event (invokeTurn's onCompleted), which is BEFORE streamReply's terminal finalize delivers the answer to Telegram. A crash in that window: the answer is in the session, the user never sees it, and it is not replayed (correctly — the session is committed, replay would double-append the turn). So the window L1 recovers is ACKed-but-un-COMPLETED, not ACKed-but-un-DELIVERED as "removed when the turn ends" (here, docs/telegram.md, CHANGELOG) reads. This header enumerates every other window (pre-ACK redelivery overlap, the orphan preview) but not this one.

Direction: one clause naming the residual undelivered window and why removing at `completed` rather than after delivery is the right trade (replay of a session-committed turn would double-append).

kid7st (commented):
Thesis: adds `turn-store.ts` — L1 durable turn intent (persist pre-ACK, remove at turn end, replay on next start, per-turn execution ceiling) layered on the in-memory turn queue. The concept is clean and its at-least-once trade-offs are documented exhaustively. 3 findings, all inline: the `defer` path's two unnamed semantic gaps (FIFO inversion, orphaned ⏳ notice) and one doc-precision gap (the completed-but-undelivered residual window).

kid7st:
Addressed the 3 inline review comments (threads resolved):

- **defer breaks per-session FIFO** — named the reorder window explicitly in the defer branch comment and `docs/telegram.md` L1 bullet (no chain-stalling machinery — a disk-degraded corner).
- **defer orphans the ⏳ notice** — on defer with a `previewId`, best-effort `deleteMessage` the stale "Queued" notice (consistent with the drop path's no-orphan invariant); the replay sends a fresh preview.
- **"removed when the turn ENDS" overstates coverage** — clarified the header: removal fires at `completed` (session-committed), so L1 recovers ACKed-but-un-COMPLETED, not un-DELIVERED; a crash in the undelivered tail is deliberately not replayed (would double-append). Same nuance in `docs/telegram.md`.

Amended into the feature commit. lint + typecheck + test green (346).

kid7st (commented):
No findings. Adds L1 durable turn intent (`turns.json`): persist an accepted turn pre-ACK, remove it when the turn ends, replay an interrupted turn (crash OR SIGTERM deploy) on the next start. A deliberate reversal of #123 (0.8.3), but NOT the removed queue-WAL — a separate intent log with at-least-once semantics, an execution ceiling, and a fail-closed bump; L2 exactly-once stays at the K-axis backend.

Verified across the whole flow. Error responsibility is correctly allocated at every write: `add` throws pre-ACK → rolls memory back → the handler rejects → nodeListener maps it to 500 (confirmed in http.ts) → Telegram redelivers; `remove` is post-ACK best-effort (never aborts delivery); `startAttempt`'s bump is the one fail-closed post-ACK write (defer, not untracked-run, so a poison turn's ceiling always advances on disk). The ceiling is per-turn, bumped at dequeue, untouched by `recover()`, so a poison head is dropped on its own N+1th run without penalizing never-run siblings. The completed hook clears in the right order (remove intent, THEN commit the buffer) so a crash between them only re-folds the same context, never replays a context-stripped turn; replay re-peeks the buffer fresh, composing correctly with the buffer's own pre-ACK durability. `PendingTurn extends Omit<StoredTurn,'attempts'>` makes execution fields durable by construction. The pre-ACK-window double-run (recover + redelivery) is documented as the at-least-once cost, not a hidden bug.

Tests are strong: 14 unit invariants (rollback, defer, drop, 1→2→3→drop convergence, the `>` vs `>=` boundary, wrong-shape degrade) plus integration that drives the real failure path — a directory at `turns.json.tmp` forces a genuine EISDIR in `saveStateFile` to prove the pre-ACK reject and no phantom on disk — alongside recover-replay, ceiling drop+notify, and defer-without-notify. Non-tautological, failure-paths covered, mocking restrained.

The reversal of #123 is explicitly owned (PR body, module header, CHANGELOG, core.md §11) with the trade named honestly (a visible/self-correcting loss traded for invisible duplicate side effects, triggered by every deploy that interrupts an in-flight turn; at-least-once with no opt-out, the idempotency burden on the agent author). That is the correct first-principles response to the earlier stance, consciously documented — surfaced here only for the record, not as an open question.

PR #129, opened 2026-07-05
feat!: channel mount context + one state root (FASTAGENT_STATE_DIR)

kid7st:
## What & why

Two coupled changes, one commit.

**1. Channel mount context.** The channel contract becomes `ChannelModule = (ctx: { agent, stateRoot }) => Routes`. Bundled adapters are now policy-only factories: `telegramChannel(opts)` / `githubChannel(opts)` take options (no `agent` arg) and RETURN a `ChannelModule` mounting their route (`POST /telegram`, `POST /webhook`). A scaffolded channel file is one policy expression — `export default telegramChannel({…})` — and `agent`/`stateRoot` flow framework → adapter without transiting user glue.

This fixes a real anchor bug: telegram state was rooted at `process.cwd()/.fastagent` while engine state used `<dir>/.fastagent`, so `fastagent start <dir>` from elsewhere split the state home (L1 turn replay depended on files in the wrong place).

**2. One state root.** `.fastagent/` is now formally the single-lifecycle machine-state home, resolved once (`FASTAGENT_STATE_DIR` > `<dir>/.fastagent`, absolute) with sessions/auth/channel state all derived from it. A container mounts ONE volume; the finer `--sessions-dir`/`--auth-path` knobs still override on top. The self-ignore leak guard is now root-based, so a custom in-tree root (a `FASTAGENT_STATE_DIR` inside the agent dir) is git-ignored too — closing a gap where credentials under a non-`.fastagent` in-tree root could show as committable.

## Why the interface change

Channels need the framework's *resolved* state root (the channel can't compute it — it doesn't know `dir` or the env precedence). Handing it via env would conflate the operator input plane with the in-process handoff (process-global, breaks embedding, kills multi-workspace); threading it through user glue taxes every author with plumbing. A single ctx object is the extensible seam — future fields (logger, shutdown) add without churning the signature.

## BREAKING

- `ChannelModule` signature: `(agent) => Routes` → `(ctx: { agent, stateRoot }) => Routes`.
- `telegramChannel` / `githubChannel`: drop the `agent` argument; return a `ChannelModule`.
- `TelegramChannelOptions.stateDir` removed (the home is derived from the state root; glue can't bypass the operator's one knob).
- Migration notes in CHANGELOG, including the telegram state-relocation warning for `start <dir>` with cwd ≠ dir.

## Surface touched (why this waits for review)

Public API: `ChannelModule`, `ChannelContext`, adapter signatures. Per CONTRIBUTING this is the review tier, not self-merge.

## Verification

`npm run typecheck && npm test` → 353 passed. Local isolated review loop iterated to `No findings`.


kid7st (commented):
No findings.

ChannelContext = { agent, stateRoot } 挂载握手 + resolveStateRoot 单状态根: 概念自洽, 命名修正到位 (isUnderStateDir→isUnderDir, ensureInTreeStateSelfIgnored→ensureStateRootSelfIgnored, defaultProject*→default*(stateRoot)), 绝对路径双重断言 fail-visibly, 无残留旧符号引用, register-webhook /telegram 路径与新 route key 匹配, 测试含失败/边界/派生 scenario. stateDir 移除收窄了单 workspace 多 telegram 能力但属已声明的有意约束, 不计为 finding.

PR #153, opened 2026-07-06
feat(deploy): railway target — Phase 1 (generate + runbook)

kid7st:
## Phase 1 of the Railway deploy target — generate + runbook

Adds `fastagent deploy railway` (second host). Confirmed trade-offs up front: **not a copy of Fly**, and where Railway is weaker we name it rather than fake parity.

### The neutral-core extraction the second host forces

- `src/deploy/container.ts` — Dockerfile + .dockerignore (the portable container, host-neutral; both targets ship it unchanged).
- `src/deploy/secrets.ts` — `isEnvKey` + `requiredSecrets` (the required-secret NAME list; identical on every host, only the *set* command differs).
- `fly.ts`/`fly-run.ts` now import from these — **no behavior change** (all Fly tests green).
- `runDeploy` pre-flight (config/model/channels/container facts + warnings) is host-neutral; the host branch adds its config file + runbook.

### `src/deploy/railway.ts` — three named asymmetries vs Fly

1. **Thin config file.** `railway.json` holds only build/deploy (`builder=DOCKERFILE`, `healthcheckPath=/health` — which also fixes Fly's "routed before the server is listening" boot race). Volume, variables (state root + secrets), and App Sleeping are CLI/dashboard service settings — Railway's source of truth is the linked project, **not a committed file** (Fly's fly.toml-single-source does not carry).
2. **Scale-to-zero is not scriptable.** App Sleeping is dashboard-only (no CLI/API), so it's a stated manual step; the github-no-replay floor becomes "do NOT enable App Sleeping."
3. **Minted public URL.** `railway domain` mints a `*.up.railway.app` domain, read back for the webhook (vs Fly's deterministic `.fly.dev`) — minted once even when both channels are present.

`--stop`/`--no-scale-to-zero`/`--run` are Fly-only (warn/degrade on Railway; `railway --run` prints the manual runbook then exits non-zero). `--run` for Railway is Phase 2.

### Tests / verification

`test/deploy-railway.test.ts` (7 cases: thin json, shared container, state-root+secrets, minted-domain webhook, App Sleeping manual + github-forbid, non-env auth → guidance, domain-minted-once). Smoke-tested the CLI end-to-end (channel discovery → webhook + secrets in the runbook). `npm run typecheck && npm test` → 401 passed. Local isolated review to convergence.

### Review tier

New public-ish surface (`deploy/container.ts`, `deploy/secrets.ts`, `deploy/railway.ts`) + CLI change — flagging for review, not self-merge.


src/deploy/railway.ts
```diff
@@ -0,0 +1,148 @@
...
+    ``,
+    `# One-time: create + link a project (writes .railway link state in this dir; the project — not a`,
+    `# committed file — is Railway's source of truth for identity, variables, and the volume).`,
+    `railway init            # or \`railway link\` to attach an existing project`,
+    ``,
+    `# Persistent volume at ${MOUNT} — sessions, auth, channel state. FASTAGENT_STATE_DIR is set to match.`,
+    `railway volume add --mount-path ${MOUNT}`,
+    ``,
+    `# Variables: the state root (must equal the mount path) + secrets. Railway injects PORT itself.`,
+    `railway variables --set FASTAGENT_STATE_DIR=${MOUNT}`,
```
kid7st: `railway variables --set KEY=VALUE` is the DEPRECATED legacy-flag form per docs.railway.com/cli/variable. Current syntax is the `set` subcommand: `railway variable set KEY=VALUE`, multiple as `railway variable set A=1 B=2`. This hits lines 83, 89, 100. Worse, line 89 chains multiple `--set` flags (`--set A=<value> --set B=<value>`) — not a documented form even for the legacy flag. Fix: `railway variables set FASTAGENT_STATE_DIR=/data`; secrets as one `railway variables set A=<value> B=<value>`.

src/deploy/railway.ts
```diff
@@ -0,0 +1,148 @@
...
+  const runbook: string[] = [
+    `# Deploy to Railway. railway.json / Dockerfile / .dockerignore are generated above.`,
+    `# Prereqs: the Railway CLI (https://docs.railway.com/guides/cli) and \`railway login\`.`,
+    ``,
+    `# One-time: create + link a project (writes .railway link state in this dir; the project — not a`,
+    `# committed file — is Railway's source of truth for identity, variables, and the volume).`,
+    `railway init            # or \`railway link\` to attach an existing project`,
+    ``,
+    `# Persistent volume at ${MOUNT} — sessions, auth, channel state. FASTAGENT_STATE_DIR is set to match.`,
+    `railway volume add --mount-path ${MOUNT}`,
```
kid7st, line 96: Ordering violates Railway's service model. `railway init` (line 77) creates only a PROJECT, no service. `railway volume add` attaches to a service (prompts to select one) and `railway variable set` (lines 83/89) is service-scoped — but the service isn't created until `railway up` (line 107). On a fresh project this step fails: nothing to attach the volume to / set variables on. And reversing it (up first, no vars) crash-loops the first boot (missing model key / FASTAGENT_STATE_DIR) against `restartPolicyType: ON_FAILURE` + healthcheck /health, so the order isn't cosmetic. Fix: create a service first (`railway add --service <name>`) after init, then volume/variables/up; or state the real flow (up → crash-loop → set vars+volume → redeploy).

kid7st (commented):
Thesis: adds `deploy railway` (second host) + extracts the host-neutral container/secret modules; Railway is deliberately not a Fly copy (thin config / non-scriptable scale-to-zero / minted URL). The module extraction and asymmetry framing are clean. Two runbook-correctness findings below, both in `railway.ts` (inline).

Both escaped detection because the only verification is substring assertions on the plan strings (test/deploy-railway.test.ts) + a channel-discovery smoke test — the runbook was never run against a real Railway CLI, so command-level syntax/ordering errors pass through. Phase 1's whole deliverable is an executable runbook; worth validating the exact command forms against the current CLI (docs.railway.com/cli, last updated 2026-07-01) before landing.

src/deploy/railway.ts
kid7st: Fixed. Switched to the `set` subcommand: `railway variables set FASTAGENT_STATE_DIR=/data` and secrets as one space-separated `railway variables set A=<value> B=<value>` (dropped the deprecated `--set` legacy flag and the invalid chained form). Verified against the real Railway CLI 5.15.0 (`railway variable --help` lists `set` as the subcommand; `--set` is under deprecated legacy flags) and docs.railway.com/cli/variable. Test now asserts `variables set` and `.not.toContain("--set")`.

src/deploy/railway.ts
kid7st: Fixed with option (a). Inserted `railway add --service <name>` right after `railway init` (service name = slugged dir basename; Railway service names are project-scoped, not globally unique). `add` auto-links the service, so the service-scoped volume/variables/up resolve it; and variables are set BEFORE the first `railway up` so the box boots with the model key + FASTAGENT_STATE_DIR (no crash-loop against ON_FAILURE + healthcheck). Verified `railway add --service [NAME]` against the real CLI 5.15.0. Left a breadcrumb that Phase 2 `--run` must pass `--service` explicitly to stay non-interactive. New test asserts the service is created and ordered before volume/variables/up.

kid7st (commented):
Re-review after the revision. Both prior findings fixed cleanly: `railway variables set` (not the deprecated `--set` flag; secrets space-joined in one command) + a `railway add --service` step ordered before the service-scoped volume/variables/up, with new tests asserting the order and a `not.toContain('--set')` regression guard. One remaining parity gap, inline in railway.ts.

src/deploy/railway.ts
```diff
@@ -0,0 +1,160 @@
...
+  // variables are service-scoped and `railway up` deploys THE service — so the service must exist first
+  // (`railway add --service`), and variables must be set BEFORE the first `up` or the box boots without a
+  // model key / FASTAGENT_STATE_DIR and crash-loops against restartPolicy ON_FAILURE + the healthcheck.
+  const runbook: string[] = [
+    `# Deploy to Railway. railway.json / Dockerfile / .dockerignore are generated above.`,
+    `# Prereqs: the Railway CLI (https://docs.railway.com/guides/cli) and \`railway login\`.`,
+    ``,
+    `# One-time: create + link a project (writes .railway link state in this dir; the project — not a`,
+    `# committed file — is Railway's source of truth for identity, variables, and the volume).`,
+    `railway init            # or \`railway link\` to attach an existing project`,
```
kid7st, line 88: Redeploy footgun the Fly runbook guards but this one doesn't. Fly wraps all setup under one banner: `# One-time setup (skip on a redeploy — a second run makes a SECOND app/volume, splitting state)`. Here only `railway init` is labeled one-time; `railway add --service` (89), `railway volume add` (92), and `railway variables set` (95) carry no redeploy note. A user re-running the whole printed runbook for a second deploy gets a new project, a DUPLICATE service (Railway names are project-scoped, not unique — no error), and another volume — splitting state. The real redeploy path is just `railway up`. Fix: mark the init→add→volume→variables block as one-time ("skip on redeploy — a redeploy is just `railway up`; re-running splits state"), mirroring the Fly banner.

kid7st: Fixed, mirroring the Fly banner. The whole init→service→volume→variables block now sits under `# One-time setup … SKIP all of it on a redeploy — a redeploy is just `railway up`. Re-running these makes a second project, a DUPLICATE service (Railway names are project-scoped, NOT unique — no error), and another volume, splitting state.`, and the `railway up` step is labeled as the entire redeploy path. Railway needs this guard MORE than Fly (Fly app names are globally unique so `apps create` errors on a re-run; Railway silently duplicates). New test asserts the runbook is marked one-time and names `railway up` as the redeploy.

kid7st (commented):
No findings. Re-review after the third revision — all prior findings closed with tests: (1) `railway variables set` not the deprecated `--set` flag; (2) `railway add --service` ordered before the service-scoped volume/variables/up; (3) the one-time setup now carries a skip-on-redeploy banner naming the duplicate-service/volume state-split, plus a note that `railway up` alone is the whole redeploy. Also added a CLI-level test for `deploy railway --run` fail-visible degrade (exit 1 + actionable stderr + runbook on stdout). Railway target Phase 1 is clean.

src/deploy/railway-run.ts
```diff
@@ -0,0 +1,211 @@
...
+  // 2. Gate missing required secret VALUES before any side effect (no half-created infra).
+  if (plan.missingSecrets.length > 0) {
+    return gate(
+      `no local value for: ${plan.missingSecrets.join(", ")} — set them in .env (or the environment) and re-run`,
+    );
+  }
+
+  // 3. Linked? (stdout-based, see parseLinked). Unlinked → create the project + service; linked → this is
+  //    a redeploy, skip creation (re-running init/add would duplicate the project + service, splitting state).
+  const status = await railway(["status", "--json"], { capture: true });
```
kid7st, line 142: `railway status --json`'s exit code is never checked — only `parseLinked(status.stdout)`. The stdout-presence signal correctly handles the documented "unlinked also exits 0" quirk, but it can't distinguish a genuine FAILURE: whoami passed, then status errors (network / CLI bug / rate limit) with non-zero exit and empty stdout → parseLinked=false → misread as unlinked → `railway init` runs. If the dir was actually linked, init DUPLICATES the project + service — the exact split-state outcome the linked-check exists to prevent. The Fly runner gates this class explicitly (fly-run.ts:101, `if (appsList.code !== 0) return gate(...)` — an errored list ≠ absent); the Railway runner is missing the equivalent. Fix: before `parseLinked`, `if (status.code !== 0) return gate("\`railway status\` failed — …")` — the documented unlinked case is exit 0 + empty stdout, so a non-zero exit is a real error and gating it doesn't false-trip the normal unlinked path. Add a status-failure test alongside it.

kid7st (commented):
Reviewing the new Phase 2 commit (b7ecf7d, `railway --run`). Thesis: `deployRailwayRun` behind a `RailwayRunner` seam mirrors the Fly runner but is reshaped by Railway's model (stdout-based linked-check, per-command `--service`, `variable set --stdin`, minted domain), with the host-neutral secret assembly extracted to `assembleSecrets`. Structure and test coverage (command-sequence benchmark + gates + stdin-not-argv + redeploy-skip + half-provisioned volume heal + parser edges) are solid. One robustness gap, inline in railway-run.ts.

PR #171, opened 2026-07-06
feat(deploy): declare agent env + system tools in config (G4 + G6)

kid7st:
### The gaps (real-world validation, B mode)

Turning a repo into an agent, the deploy path couldn't express two things it needed, so I hand-set a Railway variable and hand-wrote a Dockerfile:
- the agent's tools needed `GH_TOKEN` — deploy only knew the model + channel secrets → manual `railway variables set`.
- the agent shelled out to `git` — `node:22-slim` has none → the generated Dockerfile was useless, hand-wrote one.

### Fix — a `deploy` section in `fastagent.config`

| Field | Gap | Behavior |
|---|---|---|
| `deploy.secrets: string[]` | G4 | Extra secret env-var names the agent needs. `requiredSecrets` lists them in the runbook; `assembleSecrets` carries each value from the local env under `--run`, gating a missing one like a channel secret (deduped against the model/channel names). |
| `deploy.apt: string[]` | G6 | Extra apt packages baked into the generated Dockerfile (Debian default repos). A package needing a custom repo (`gh`) or a different base still uses a hand-written Dockerfile — deploy keeps an existing one (and **warns** if `deploy.apt` is declared but a kept Dockerfile drops it). |

Both are **shape-validated at the config boundary** (UPPER_SNAKE env names, Debian package names) — a typo or an injection-shaped value in an otherwise-trusted config fails visibly (a `git; rm -rf /` apt entry is refused). Threaded through fly + railway (runbook + `--run`). `init` scaffolds the commented `deploy` section; configuration.md + core.md document it.

### Tests

config validation (valid + bad-shape + unknown-key), `assembleSecrets`/`requiredSecrets` carry + dedup, the apt Dockerfile layer (present when declared, absent otherwise). `npm run typecheck && npm test` → 430 passed. Local review to convergence (caught + fixed: dedup, and the kept-Dockerfile-drops-apt warn).

Config schema + both hosts — review tier. Closes the P2 'real repo needs env + tools' gaps.

PR #173, opened 2026-07-06
refactor(deploy): runDeployFly/runDeployRailway take an opts object

kid7st:
## runDeployFly/runDeployRailway → opts object

A local-review follow-up flagged after #171: `runDeployFly` had 8 positional params, `runDeployRailway` 7, with adjacent same-type args (a `string` `flyTomlPath` next to a `string[]` `extraSecrets`) — mis-order-prone, and the type system couldn't catch a swap. Every new deploy declaration (the `extraSecrets` #171 added is what tipped it) appended another positional slot.

Now each takes a single opts object; the call sites pass the same values by name. Pure mechanical refactor — no behavior change, field names map 1:1 (`serviceName → name`, `appName` kept), both call sites updated. 431 tests unchanged (these are L0 CLI glue; the tested logic lives in `deployFlyRun`/`deployRailwayRun`/`assembleSecrets`).

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -551,7 +551,7 @@ async function runDeploy(): Promise<void> {
         .replace(/^-+|-+$/g, "") || "agent";
     const plan = planRailwayDeploy({ serviceName, modelAuth, channels, extraSecrets, ...container });
     await writeArtifacts(target, plan.artifacts);
-    if (values.run) return runDeployRailway(target, serviceName, modelAuth, authPath, channels, extraSecrets);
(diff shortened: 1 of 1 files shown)


kid7st (commented):
No findings. 纯机械重构(位置参数→opts 对象),字段映射逐一核对为 1:1:runDeployFly 7 字段全对应,runDeployRailway 的 `name: serviceName` 正确保留了旧第二位置参数的映射。两个函数各只有一处调用点(cli.ts:554/605)且都已更新,无遗漏调用方;typecheck 通过;函数体仅新增 destructure 行、内部变量名不变 → 无行为变更。(PR body 称 runDeployFly '8 positional params',实际 7 —— 纯描述性误差,不影响代码。)

PR #175, opened 2026-07-06
fix(runtime): run under Bun, not only Node

kid7st:
## Run under Bun, not only Node

Deploying a Bun agent (an Astro site whose `AGENTS.md` drives content ops), `bunx fastagent start` crashed before it could serve:

```
SyntaxError: Export named 'install' not found in module 'undici'
```
and, once past that:
```
TypeError: process.loadEnvFile is not a function
```

Two Node-only couplings, both on the entry path:

| File | Node-only | Fix |
|---|---|---|
| `proxy.ts` | `import { install } from "undici"` — Bun's undici is a native shim without `install`, and a **static named** import of a missing export fails at **link time**, crashing before any code runs | Namespace import (binds whatever exists) + a Bun guard. Bun's native fetch already honors proxy env vars and decompresses gzip, so the undici wiring is a Node-only no-op there. |
| `cli.ts`/`tunnel.ts` | `process.loadEnvFile` (Node 20.12+, absent on Bun) | Extract a portable `loadEnvFile` (`env.ts`): parse `KEY=VALUE` ourselves, matching Node's precedence — a real env var wins over the file (verified against Node). |

**Audited all of `src`** for runtime-sensitive APIs: only these two. The `node:` builtins (path/fs/child_process/http/crypto/net/os…) and the bare deps (chokidar, octokit, clack, ignore) are Bun-compatible.

**Verified end-to-end under Bun 1.3**: `info`, `invoke` (a real OAuth model call), and `start` (`/health` 200, telegram channel loaded) all work. 4 new `env.test.ts` cases (parse, first-`=` split, no-override precedence, ENOENT propagation); 435 tests.

Unblocks a single-runtime Bun image (`FROM oven/bun` + `bunx fastagent start`) instead of a Node+Bun dual base.

### DIFF
--- a/src/env.ts
+++ b/src/env.ts
@@ -1,0 +1,33 @@
+import { readFileSync } from "node:fs";
+
+/**
+ * Load a `.env` file into `process.env`, matching Node's `--env-file` / `process.loadEnvFile` precedence
+ * on BOTH axes (verified against Node): a real env var wins over the file (an already-set key is kept),
+ * and within the file a repeated key takes the LAST occurrence. Portable across Node and Bun: Bun has no
+ * `process.loadEnvFile`, so we parse the file ourselves rather than depend on a Node-only entry point.
+ *
+ * Two phases keep the two rules distinct: parse into a map (later line overrides earlier = last-wins),
+ * then apply only the keys `process.env` doesn't already have (env-wins). Minimal parser — `KEY=VALUE`
+ * per line, `#` comments and blank lines skipped, surrounding matched single/double quotes stripped:
+ * enough for the flat secret files fastagent reads (tokens, keys), not a full dotenv dialect (no
+ * multiline, no `export`, no interpolation). A missing file throws ENOENT for the caller to treat as
+ * "no .env"; any other read error propagates.
+ */
+export function loadEnvFile(file: string): void {
+  const content = readFileSync(file, "utf8");
+  const parsed = new Map<string, string>();
+  for (const raw of content.split("\n")) {
+    const line = raw.trim();
+    if (!line || line.startsWith("#")) continue;
+    const eq = line.indexOf("=");
+    if (eq < 1) continue; // no `=`, or an empty key
+    const key = line.slice(0, eq).trim();
+    let value = line.slice(eq + 1).trim();
+    const quote = value[0];
+    if ((quote === '"' || quote === "'") && value.at(-1) === quote) value = value.slice(1, -1);
+    parsed.set(key, value); // in-file: last occurrence wins (Map overwrite)
+  }
+  for (const [key, value] of parsed) {
+    if (!(key in process.env)) process.env[key] = value; // env-vs-file: a real env var wins
+  }
+}
--- a/test/env.test.ts
+++ b/test/env.test.ts
@@ -1,0 +1,76 @@
+import { describe, expect, it } from "vitest";
+import { mkdtemp, writeFile } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { loadEnvFile } from "../src/env.ts";
+
+// A portable .env loader (Node has process.loadEnvFile; Bun does not — same parse must run on both).
+describe("loadEnvFile", () => {
+  const write = async (content: string): Promise<string> => {
+    const dir = await mkdtemp(join(tmpdir(), "fa-env-"));
+    const file = join(dir, ".env");
+    await writeFile(file, content);
+    return file;
+  };
+
+  it("loads KEY=VALUE, skips comments/blank lines, strips matched quotes", async () => {
+    const key = `A_${Date.now()}`;
+    const file = await write(`# a comment\n\n${key}_PLAIN=value\n${key}_DQ="quoted spaces"\n${key}_SQ='single'\n`);
+    loadEnvFile(file);
+    expect(process.env[`${key}_PLAIN`]).toBe("value");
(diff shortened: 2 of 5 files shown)


src/env.ts
```diff
@@ -0,0 +1,27 @@
...
+  for (const raw of content.split("\n")) {
+    const line = raw.trim();
+    if (!line || line.startsWith("#")) continue;
+    const eq = line.indexOf("=");
+    if (eq < 1) continue; // no `=`, or an empty key
+    const key = line.slice(0, eq).trim();
+    let value = line.slice(eq + 1).trim();
+    const quote = value[0];
+    if ((quote === '"' || quote === "'") && value.at(-1) === quote) value = value.slice(1, -1);
+    if (!(key in process.env)) process.env[key] = value;
```
kid7st: 文件内重复 key 的 precedence 与 Node 相反,与本函数 docstring 声称的 'matching Node's precedence' 不符。实测 Node `process.loadEnvFile`(v26):env-vs-file 是 env-wins(这行 `if (!(key in process.env))` 匹配了这层,正确),但**文件内**重复 key Node 是 last-wins(`DUP=first\nDUP=second` → `second`);而 no-override 写法第一次设值后即 skip → first-wins(`first`)。一个 `if (!(key in process.env))` 把两种不同规则压成了同一个 first-wins —— Node 的规则是 env-vs-file 取 env、file 内取末个。

扁平 secret 文件里重复 key 罕见,但 docstring 正面声称匹配 Node precedence,这是声称与实现不符。方向:先 parse 进临时对象(文件内后者覆盖前者),最后逐 key `if (!(key in process.env))` 落地到 `process.env`;或在 docstring 现有的偏差声明(no multiline/export/interpolation)里补一句『文件内重复 key 取首个,与 Node 末个相反』。

test/env.test.ts
```diff
@@ -0,0 +1,45 @@
...
+  });
+
+  it("keeps a value with '=' and ':' intact (only splits on the first =)", async () => {
+    const key = `B_${Date.now()}`;
+    const file = await write(`${key}=7676:AA=bb-cc_dd\n`);
+    loadEnvFile(file);
+    expect(process.env[key]).toBe("7676:AA=bb-cc_dd");
+  });
+
+  it("does NOT override an already-set var — a real env var wins (matches Node's process.loadEnvFile)", async () => {
```
kid7st: 断言 'matches Node's process.loadEnvFile' 却从不调用 `process.loadEnvFile` 对照 —— tautological。测试名把『匹配 Node』写进 spec,但测试体只调用自己的 `loadEnvFile`,没有一处跑 `process.loadEnvFile` 比对,锁定的是实现自身行为而非『与 Node 等价』。整个 Bun fix 的正确性都押在这个等价上,却没有测试真正保护它:我手动跑 Node 才确认 env-vs-file 这层 no-override 成立(实现碰巧对),而 file 内重复 key 这层实测与 Node 相反(见 env.ts:25),恰恰是本测试没覆盖的 scenario。

方向:在 Node 上把 `loadEnvFile` 与 `process.loadEnvFile` 喂同一 .env 断言 `process.env` 结果一致(Node-only,`process.versions.bun` 时 skip),让『匹配 Node』这个 spec 真的被验证;至少补一个重复-key case 固定实际行为,与 env.ts:25 的决定对齐。

kid7st (commented):
本次改动让 fastagent 在 Bun 下运行,核心是 `env.ts` 的 portable `loadEnvFile` 替代 Node-only 的 `process.loadEnvFile`,必须语义等价。实测 Node:env-vs-file 这层是 env-wins(实现对,fix 主体成立),proxy.ts 的 namespace import + Bun guard 正确、所有 `process.loadEnvFile` 调用点都已替换、无遗漏 Node-only API。两条 finding 都落在 precedence 的一个未验证角落,见 inline。

src/env.ts
kid7st: Fixed by making the two precedence rules distinct (two phases). Parse into a Map first — a repeated key overwrites, so in-file is now LAST-wins like Node (verified: DUP=first/DUP=second → second). Then apply only keys process.env lacks — env still wins over the file. The docstring now claims Node parity on BOTH axes and it's true; the deviation note is trimmed to the real ones (no multiline/export/interpolation).

test/env.test.ts
kid7st: Fixed — the 'matches Node' claim is now a real differential test: it feeds the SAME .env to loadEnvFile and to process.loadEnvFile.bind(process) and asserts process.env agrees, across a plain key, an in-file duplicate, a quoted value, and a preset (real-env-wins) key. Node-only via it.skipIf(process.versions.bun) since Bun — the port target — has no process.loadEnvFile. Also added a standalone in-file-last-wins case. No longer tautological; 437 tests.

kid7st (commented):
No findings. 上一轮两条 finding 均正确解决:(1) loadEnvFile 改两阶段——先 parse 进 Map(文件内 last-wins),再对 process.env 做 no-override 落地(env-wins),两种 precedence 规则分开表达且都匹配 Node,docstring 更新为 'BOTH axes';(2) 测试名去掉未验证的 'matches Node' 承诺,新增 in-file last-wins case,并加差分测试把同一文件喂给 `loadEnvFile` 与 `process.loadEnvFile.bind(process)` 断言 process.env 一致(it.skipIf(bun),CI 在 Node 上真跑)。差分测试隔离正确(每次 run 重置 keys + preset)、覆盖两 axes+quote+comment+重复 key,是真验证非 tautological。proxy/cli/tunnel 未变。

PR #177, opened 2026-07-07
feat(devx): B-mode on-ramp (init adopt) + bun-aware generated Dockerfile

kid7st:
## B-mode on-ramp (init adopt) + bun-aware generated Dockerfile

Turning a real **Bun** repo (an Astro site whose `AGENTS.md` drives content ops) into a deployed agent surfaced two dead-ends the whole B-mode pitch ("existing repo → agent, without rewriting it") stumbles on.

### 1. On-ramp — no smooth path for an existing repo

`init` refused an existing-`AGENTS.md` repo (the B-mode case), and `add` gated on a dep with **npm-only** guidance on a **bun** project. Now:

- **`fastagent init` on an `AGENTS.md` repo with no config ADOPTS it**: adds the one missing piece (`fastagent.config.mjs`) + a secret-safe `.gitignore` (`.env`/`.fastagent`), keeps `AGENTS.md`/`skills`/`tools` **untouched**, and guides with the RIGHT package manager (`bun add` vs `npm install`). Never hand-edits `package.json` — the pm owns the lockfile. `--minimal` is a no-op here (noted, not silently dropped).
- **`add`'s dep gate + next-steps are package-manager-aware.**
- Routing: AGENTS.md + no config → adopt; fresh → scaffold; has config → refuse (tested end-to-end).

### 2. Bun Dockerfile — the generated one was npm/node-only

A bun workspace (`packageManager: bun` / a bun lockfile) now gets `FROM oven/bun` + `bun install` + `bunx fastagent start` (works since 0.10.2's Bun runtime support), instead of forcing a hand-written Dockerfile. `bunVersion` strips corepack's `+sha256` suffix (an invalid Docker tag).

### Behavior change (all node deploys)

The generated Dockerfile **no longer passes `--omit=dev`** — it installs devDependencies too. A repo-as-agent runs its own build/check toolchain (which lives in devDependencies) to do its work, and we can't tell a repo-as-agent from a purpose-built workspace, so the safe default installs everything (correctness over image size). Applies to both runtimes.

### Structure

New neutral `src/runtime.ts` (`detectRuntime`) is the ONE source both the Dockerfile and the CLI hints read. `FlyPlanInput`/`RailwayPlanInput` now `extends ContainerInput` — the container facts were duplicated in 3 places, and that drift is what broke the build when a field was added.

444 tests (new: runtime detection incl. corepack, adopt + its secret-safe .gitignore branches, init routing, bun Dockerfile).

---

## Also: the 🟡 secondary frictions from the same validation

| # | Fix |
|---|---|
| #3 | `models <search>` ranks a **provider-name** match above an incidental model-id match — `models anthropic` leads with `anthropic/*`, not `amazon-bedrock/anthropic.*`. |
| #5 | An expired/revoked OAuth reported a contradictory `auth: (none found)` then `OAuth refresh failed`. `reportAuth` now reads the store **without a refresh** to tell "nothing stored" from "stored but unusable", reporting the latter as expired → `fastagent login`. |
| #6 | The generate-mode runbook's `Model auth: none configured locally` now hints that a **global** login isn't read there — pass `--auth-path`, or `--run` carries it. |
| #9 | `login` in a non-TTY (pipe/CI/agent shell) hung on a menu it couldn't answer; it now **fails fast** with the reason, after the secret-hygiene self-ignore. |

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -20,6 +20,7 @@ import { type Routes, parseRouteKey, router, serveNode } from "./host/node.ts";
 import { runDevSupervisor } from "./dev-supervisor.ts";
 import { announceWebhooks, startCloudflareTunnel } from "./tunnel.ts";
 import { discoverChannelFiles, loadChannels } from "./engines/pi/channel.ts";
+import { detectRuntime } from "./runtime.ts";
 import { fastagentVersion } from "./version.ts";
 import {
   defaultAuthPath,
@@ -51,7 +52,7 @@ import {
   channelSetup,
   scaffoldChannel,
 } from "./scaffold/add-channel.ts";
-import { exists, nextStepCd, scaffoldWorkspace } from "./scaffold/init.ts";
+import { adoptWorkspace, exists, nextStepCd, scaffoldWorkspace } from "./scaffold/init.ts";
 import { vendorSkill } from "./scaffold/vendor-skill.ts";
 import { isGeneratedDockerfile } from "./deploy/container.ts";
 import { modelTravelIssue, parseFlyAppName, parseFlyRegion, planFlyDeploy, toFlyAppName } from "./deploy/fly.ts";
@@ -332,6 +333,12 @@ async function runInfo(): Promise<void> {
 }
 
 async function runInit(): Promise<void> {
+  // An existing repo (AGENTS.md) with no fastagent.config → ADOPT it (the B-mode on-ramp), don't refuse:
+  // the repo IS the agent, so add the missing config + guide, keeping AGENTS.md/skills/tools untouched.
+  const configNames = ["fastagent.config.ts", "fastagent.config.js", "fastagent.config.mjs"];
+  const hasConfig = (await Promise.all(configNames.map((n) => exists(join(dir, n))))).some(Boolean);
+  if (!hasConfig && (await exists(join(dir, "AGENTS.md")))) return runAdopt();
+
   const minimal = values.minimal ?? false;
   const { complete, created, skipped, intoNonEmpty, warnings } = await scaffoldWorkspace(dir, { minimal }).catch(
     failStartup,
@@ -364,6 +371,28 @@ async function runInit(): Promise<void> {
   console.error(`    fastagent add skill <owner/repo/path>   # vendor more skills from GitHub`);
 }
 
+/** `fastagent init` on an existing AGENTS.md repo: adopt it (add config + guide), keeping its own files. */
+async function runAdopt(): Promise<void> {
+  const r = await adoptWorkspace(dir).catch(failStartup);
+  console.error(`[fastagent] adopted ${dir} — kept your AGENTS.md`);
+  // --minimal shapes a fresh scaffold; adopt writes only the missing config, so it has no effect here.
+  // Say so rather than silently ignore the flag (the user may think they got a minimal scaffold).
+  if (values.minimal) console.error(`  note: --minimal ignored — adopting an existing repo, not scaffolding`);
+  if (r.created.length > 0) console.error(`  created: ${r.created.join(", ")}`);
+  if (r.patched.length > 0) console.error(`  updated: ${r.patched.join(", ")}`);
+  const add = r.runtime === "bun" ? "bun add @kid7st/fastagent" : "npm install @kid7st/fastagent";
+  console.error(`  next steps:`);
+  const cdTarget = nextStepCd(process.cwd(), dir);
+  if (cdTarget) console.error(`    cd ${cdTarget}`);
+  console.error(`    set model in fastagent.config.mjs   # \`fastagent models\` lists the specs`);
+  if (!r.hasPackageJson) {
+    console.error(`    ${r.runtime === "bun" ? "bun init" : "npm init -y"}   # a channel needs a package.json`);
+  }
+  if (!r.hasFastagentDep) console.error(`    ${add}   # a channel file imports it`);
+  console.error(`    fastagent add telegram   # or github — a channel to serve on`);
+  console.error(`    fastagent dev            # serve locally and iterate`);
+}
+
 /** `fastagent add <channel> [dir]`: scaffold `channels/<kind>.ts` — the adapter import plus a starter `on()`. */
 async function runAdd(): Promise<void> {
   const kind = positionals[1];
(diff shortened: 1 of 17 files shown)


src/deploy/container.ts
```diff
@@ -45,24 +51,42 @@ function dockerfile(input: ContainerInput): string {
  && rm -rf /var/lib/apt/lists/*
 `
       : "";
+  const isBun = input.runtime === "bun";
+  const base = isBun ? `oven/bun:${input.bunVersion ?? "1"}` : "node:22-slim";
```
kid7st, line 70: markdown 分支复用了由 runtime 决定的 `base`,而其后注释的推理链不完整(latent,非活 bug)。这里 `const base = isBun ? "oven/bun:…" : "node:22-slim"` 取自 `input.runtime`,与 `hasPackageJson` 无关。下方 markdown 分支 `if (!input.hasPackageJson)` 复用 `head`(内嵌 `base`)却跑 `npm i -g` + `CMD ["fastagent"]` —— 这些依赖 npm/全局 fastagent,存在于 `node:22-slim`,但 **`oven/bun` 镜像没有 npm**。

注释 'a markdown agent has no packageManager, so this is always the Node/npm path' 只覆盖 packageManager 一路,但 `detectRuntime` 判 bun 的条件是 `pm.startsWith("bun@") || bunLock`。一个无 package.json 但有 stray `bun.lock`/`bun.lockb` 的工作区 → `runtime:"bun"` → `base:"oven/bun"` → 走 markdown 分支在 Bun 镜像上 `npm i -g` → build 崩。触发需『无 package.json 却有 bun 锁文件』的不一致状态(bun 锁是 `bun install` 产物),现实近乎不可达,所以是 latent 而非活 bug —— 但注释所依赖的『无 package.json ⟹ 非 bun』这个不变量并未被 `detectRuntime` 保证。方向:markdown 分支显式 pin `node:22-slim`(靠全局 `npm i -g` 的 markdown agent 必须 node base),不复用可能为 bun 的 `base`;或修正注释,言明它依赖这个未强制的不变量。

kid7st (commented):
本次改动引入 `detectRuntime`(node/bun 单一来源,驱动生成的 Dockerfile base/install/run + CLI 的 pm 提示),`init` 的 B-mode adopt,并把 container facts 收敛进 `ContainerInput`(Fly/Railway extends 它)。

经 git 对象逐一核验(本次 review 的工具 stdout 有污染,故只信任 `git show <sha>:file` 抽取 + 计数/退出码):typecheck 通过、`detectRuntime` 三 case 覆盖含 corepack `+hash` 剥离、`readPackageJson` 的 try/catch 与其 docstring 一致(absent/malformed 均 → {})、`--omit=dev` 移除是已声明的行为变更 —— 均干净。一条 latent finding 见 container.ts inline。

注:本地 HEAD(4557263)比本 PR head(2c2fd74)领先一个无关的 'UX polish' commit;本 review 针对 2c2fd74,行号锚定该 commit。

src/cli.ts
```diff
@@ -858,14 +903,28 @@ function parsePort(value: string | undefined, source: string): number | undefine
...
+    log.info(`[fastagent] auth:   ${source} (${provider}) — ${authPath}`);
+    return;
+  }
+  // A `source` of undefined has TWO causes probeAuthSource can't tell apart (it swallows the throw): no
+  // credential at all, OR one that IS stored but couldn't be made usable — an expired/revoked OAuth whose
+  // refresh failed. Read the store WITHOUT triggering a refresh to distinguish them, so neither the status
+  // line nor the hint says "(none found)" when a login is merely stale (the actual failure would then be a
+  // contradictory "OAuth refresh failed"). Lead with `fastagent login` — it covers every provider (incl.
+  // OAuth-only ones) and writes this same path, fixing it in place.
+  const stored = await fastagentCredentialStore(authPath)
```
kid7st: #5 —— 本 commit 的核心错误处理修复(expired-vs-none 消歧)ships 无测试。commit 的 headline bug(#5:过期/吊销的 OAuth 报出矛盾的 '(none found)')靠这里『不触发 refresh 直接读 store』来区分 stored-but-unusable 与 none。但 #3(ranking)、#9(non-TTY)各有测试,#5 —— 语义权重最高的一项、真正的 bug —— 没有测试。一旦有人把 `stored`/none 两分支写反,或 probeAuthSource/store 的契约改变,就会 silently 退回到给过期登录报 '(none found)' 的矛盾行为,而无测试拦截。

此外 `.read(provider).catch(() => undefined)` 把**任何**读取错误都塌缩成『未存储』,所以一个 malformed/不可读的 auth.json(它本身正是 stored-but-unusable)仍会报 '(none found)' —— 恰是 #5 想修掉的那类消息。方向:加一个带 stored-but-unusable 凭证(过期 OAuth fixture)的测试,断言 reportAuth 输出 'expired/unusable → login to refresh' 而非 '(none found)';顺带决定 malformed-store 该归入哪一支。

kid7st (commented):
Re-review of the updated PR (now includes the 'B-mode UX polish' commit). Verified via git objects + counts/exit codes (this session's Bash stdout has been unreliable, so I trust only `git show <sha>:file` extracts).

Clean: #3 models ranking (provider-name match leads, order preserved within groups) + a meaningful test; #9 login non-TTY guard fails fast AFTER the secret-hygiene self-ignore + a real-spawn test; #6 runbook auth hint (--auth-path); the self-ignore test's companion comment correctly updated for the new gate ordering (typecheck=0).

One new finding (cli.ts inline). Also: the container.ts markdown+bun base-coupling finding from my prior review (on 2c2fd74) is byte-unchanged at this HEAD — still unaddressed.

src/cli.ts
kid7st: Fixed. Extracted the three-branch decision into a testable presenter (src/cli-auth.ts, mirroring cli-models.ts — reportAuth lives in cli.ts which self-executes on import), and test/cli-auth.test.ts pins all three network-free: a usable source → source line; none-but-stored → 'expired/unusable' + re-login (asserts it is NOT '(none found)'); neither → '(none found)'. Flipping the branches now fails a test. On the malformed-store point: store.read already WARNS on a corrupt/unreadable file (returning undefined), so it lands in '(none found)' alongside that warn — the real signal is surfaced there; this line stays about credential PRESENCE, documented in formatAuthReport's docstring. Re-verified the real expired-anthropic path still reports 'expired/unusable'.

kid7st (commented):
No findings. Re-review of the 2 new commits (verified via git objects + typecheck/test exit codes, this session's Bash stdout being unreliable). Both prior findings resolved correctly, each with a test pinning the exact scenario:

- container.ts markdown+bun base coupling → the markdown branch now returns EARLY (before the isBun/base computation) hardcoding `FROM node:22-slim` + `npm i -g`, so a stray bun lockfile without package.json can't select a bun base; deploy-fly.test pins it (hasPackageJson:false + runtime:"bun" → asserts node:22-slim, not oven/bun).
- #5 reportAuth untested → extracted the pure `formatAuthReport` (cli-auth.ts) + cli-auth.test.ts pinning all three branches incl. the expired case asserting NOT '(none found)'. The malformed-store residual I raised is correctly handled: the store's own `read()` warns 'corrupt auth file …' (auth.ts) and this line stays about presence — a sound, documented responsibility split (verified against auth.ts read()).

Also clean: WORKSPACE_CONFIG_NAMES consolidated the config-filename list across loader / adopt-routing / scaffold-refuse (no inline copy remains), preventing routing/refusal divergence. typecheck=0, the new tests pass 17/17.

PR #179, opened 2026-07-07
refactor(deploy): host-neutral preflight + shared runner seam + per-host directories

kid7st:
## What

A structural refactor of the `deploy` subsystem, from a post-iteration scan. Behaviour-preserving except one intended fix in the env cleanup (below); typecheck + 459 tests green.

### Why
`deploy`'s orchestration policy lived inline in `cli.ts`'s `runDeploy` (~130 lines). It was the top churn source: every deploy behaviour change edited the CLI dispatcher, and the host-neutral pre-flight had no isolated test surface. This gives it a home.

## Commits

**1. `refactor(env)` — promote `loadDotEnv`, fix tunnel's best-effort `.env` read**
The `<dir>/.env, missing is normal` pattern lived twice. `tunnel.ts` used a `catch {}` that swallowed *every* error (a corrupt/unreadable `.env` failed silently). The fail-visible version (only ENOENT tolerated) is promoted into `env.ts` and shared.

*Behaviour note (a review catch — thanks):* `announceWebhooks` is best-effort and **void-called** (`start --tunnel`, `dev-supervisor`) with no `unhandledRejection` handler, so a bare `loadDotEnv` there would let a non-ENOENT read error (EACCES / `.env` is a directory) crash the long-running dev/start server. The shared `loadDotEnv` keeps throwing (correct for synchronous command callers), and `announceWebhooks` adds a best-effort boundary: **warn and continue** instead of silently swallowing (old) or crashing (the first draft of this PR). So tunnel `.env` handling changes `silent-swallow → warn-and-continue` — a small, intended step toward fail-visible, not a crash.

**2. `refactor(deploy)` — extract preflight + runner seam, per-host dirs**
- `deploy/preflight.ts` — host-neutral pre-flight (model-travel gate, channel discovery, auth probe, container facts + warnings) as a function returning `{ ok:false, gate } | { ok:true, ...facts }`. Now unit-tested directly (`test/deploy-preflight.test.ts`) instead of only via CLI integration. `modelTravelIssue` moves here (it's host-neutral; leaving it in `fly.ts` would make the neutral kernel depend on a host dir).
- `deploy/runner.ts` — `CliRunner` + `spawnRunner`, replacing two verbatim-duplicated spawn wrappers + identical `FlyRunner`/`RailwayRunner` types.
- `readPackageJson` → `runtime.ts` (next to its consumer `detectRuntime`).
- Per-host dirs `fly/{plan,run}.ts`, `railway/{plan,run}.ts` — a hybrid layout: neutral kernel at top (horizontal) + one dir per host (vertical), so `host` is the legible extension axis (new host = new dir, copy `fly/`). Repo map in `AGENTS.md` updated.

## Layout

```
deploy/
├── preflight.ts container.ts secrets.ts runner.ts   # host-neutral kernel
├── fly/     { plan.ts, run.ts }
└── railway/ { plan.ts, run.ts }
```

## Test
`npm run typecheck && npm test` — 459 passed. New: `test/deploy-preflight.test.ts` (gate vs warn, container facts, `--force` suppression, lockfile/dep warnings); `loadDotEnv` cases in `test/env.test.ts` (missing OK / non-ENOENT propagates); a `tunnel.test.ts` regression (an unreadable `.env` warns + continues, never crashes the server).

### DIFF
--- a/src/tunnel.ts
+++ b/src/tunnel.ts
@@ -10,7 +10,7 @@ import { readdirSync } from "node:fs";
 import { join } from "node:path";
 import { setTimeout as sleep } from "node:timers/promises";
 import { registerTelegramWebhook } from "./channels/telegram/register-webhook.ts";
-import { loadEnvFile } from "./env.ts";
+import { loadDotEnv } from "./env.ts";
 import { log } from "./log.ts";
 
 export interface Tunnel {
@@ -136,9 +136,14 @@ function channelBasenames(dir: string): string[] {
 export async function announceWebhooks(dir: string, baseUrl: string): Promise<void> {
   log.info(`[fastagent] public URL: ${baseUrl}`);
   try {
-    loadEnvFile(join(dir, ".env")); // so telegram registration sees the tokens
-  } catch {
-    /* no .env is fine */
+    loadDotEnv(dir); // telegram registration reads tokens from .env
+  } catch (error) {
+    // best-effort boundary: a MISSING .env is already tolerated by loadDotEnv; an unreadable one (EACCES,
+    // or .env is a directory) must NOT crash the long-running dev/start server — announceWebhooks is
+    // void-called with no unhandledRejection handler, so a throw here would terminate the process. Warn
+    // (surface it, rule 8) and continue best-effort; telegram registration then degrades to its manual
+    // instruction if the token is absent. loadDotEnv keeps throwing for the synchronous command callers.
+    log.warn(`[fastagent] could not read ${join(dir, ".env")}: ${(error as Error).message} — continuing without it`);
   }
   const channels = channelBasenames(dir);
   if (channels.length === 0) return;
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -25,13 +25,15 @@ src/
 ├── collect.ts               # buffered consumption helper
 ├── index.ts                 # the public API surface (see README "Public API surface & stability")
 ├── cli.ts                   # command entry points (process side effects live here)
-├── invoke-stream.ts, cli-models.ts # command rendering layers (`invoke` stream → exit code, `models` output)
+├── invoke-stream.ts, cli-models.ts, cli-auth.ts # command rendering layers (`invoke` stream → exit code, `models`/auth-report output)
 ├── telegram.ts, github.ts   # subpath-export shims (@kid7st/fastagent/telegram etc. — the supported surface)
 ├── log.ts                   # leveled logging singleton (dev=debug, start=info)
 ├── observe.ts               # turn-trace logging around an Agent
 ├── tunnel.ts                # `--tunnel`: cloudflared + per-channel webhook dispatch
 ├── dev-supervisor.ts        # `dev` supervisor: restart on code-input edits (definition is live-read per invoke)
 ├── proxy.ts                 # HTTPS_PROXY wiring
+├── env.ts                   # `.env` → process.env loading (missing file is normal; anything else surfaces)
+├── runtime.ts               # workspace runtime/package-manager detection (node vs bun) + readPackageJson
 ├── workspace.ts, version.ts # neutral helpers (in-workspace guard, ignore files, version)
 ├── host/node.ts             # Node HTTP host: Routes/ChannelHandler/serveNode/router (public surface)
 ├── scaffold/                # `init` / `add <channel>` / `add skill` + templates/ (real files)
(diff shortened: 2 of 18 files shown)


src/tunnel.ts
```diff
@@ -135,11 +135,7 @@ function channelBasenames(dir: string): string[] {
  */
 export async function announceWebhooks(dir: string, baseUrl: string): Promise<void> {
   log.info(`[fastagent] public URL: ${baseUrl}`);
-  try {
-    loadEnvFile(join(dir, ".env")); // so telegram registration sees the tokens
-  } catch {
-    /* no .env is fine */
-  }
+  loadDotEnv(dir); // so telegram registration sees the tokens (missing .env is fine; a corrupt one surfaces)
```
kid7st: 这处 .env 读取改动把一个 best-effort 步骤变成了会终止进程的 unhandled rejection —— 与本 PR 'No behaviour change' 声称相悖。原代码是 try { loadEnvFile(…) } catch {},现在是裸 loadDotEnv(dir)。loadDotEnv 在非 ENOENT 读取故障时会 throw(不可读的 .env:EACCES / .env 是目录等;malformed 内容由 parser 容忍)。但 announceWebhooks 在两个调用点都是 void announceWebhooks(...)(cli.ts:945 start --tunnel、dev-supervisor.ts:71 dev),且全仓库**没有** unhandledRejection 处理器(grep=0)—— 于是 Node >=15 默认下,这里的 throw 会 reject 那个 void promise 并**终止长驻的 dev/start 服务**。旧代码吞掉故障并降级(telegram 注册在无 token 下继续 -> 打印手动指令)。

这既违背 announceWebhooks 的设计契约(best-effort、void-called、never fail the tunnel —— 见 startCloudflareTunnel/registerTelegramWebhook),也与 'no behaviour change' 不符(优雅降级 -> 崩溃)。rule-8 的意图(别静默吞掉真实读取错误)是对的,但对一个 best-effort 后台步骤,unhandled rejection 是错误的 surface。方向:在 announceWebhooks 里 try/catch 掉 loadDotEnv,非 ENOENT 用 log.warn 报出(.env 不可读、telegram token 可能缺失)但继续 best-effort 流程 —— 既 surface 故障(rule 8)又守住 never-crash-the-tunnel 契约。loadDotEnv 本身保持 throw-on-non-ENOENT(对同步的命令调用方是对的,throw 会干净退出)。

kid7st (commented):
结构重构,声称 no behaviour change。经 git 对象逐一比对 merge-base(本会话 Bash stdout 有污染,故只信任 git show <sha>:file 抽取 + typecheck 退出码):preflight.ts 忠实保留了 runDeploy 的 host-neutral 逻辑与消息顺序(modelIssue -> custom-channel notes -> lockfile -> dep -> apt);CLI 侧 gate('deploy stopped:' + exit(1))、消息前缀重建('[fastagent] {level}: {text}')、preflightDeploy(...).catch(failStartup)、authPath 解析、host-specific warnings 均与原逐字等价;runner.ts 的 spawnRunner 与两处原 inline wrapper 字节等价(同 stdio / code??1 / 127-on-error / stdin/stdout);env.ts loadDotEnv 为 ENOENT-only,typecheck=0。一条行为回归 finding 见 tunnel.ts inline —— 与 'no behaviour change' 声称相悖。

src/tunnel.ts
kid7st: Fixed in the amended commit 1. You are right on every count — verified: both call sites are `void announceWebhooks(...)` (cli.ts:945, dev-supervisor.ts:71), no `unhandledRejection` handler anywhere (grep=0), so a non-ENOENT throw here would terminate the long-running dev/start server. Root cause was mine: the rule-8 fix over-corrected a best-effort background path into a crash-via-unhandled-rejection.

Applied your exact direction: `loadDotEnv` keeps throwing (correct for the synchronous command callers → clean exit), and `announceWebhooks` now wraps it in a best-effort boundary — `log.warn` on the error and continue, so telegram registration degrades to its manual instruction instead of crashing. Improves on both the old silent `catch {}` (rule 8) and the crash. Added a `tunnel.test.ts` regression (`.env` is a directory → EISDIR → warns + resolves, never throws), and corrected the PR 'no behaviour change' claim to note tunnel `.env`: silent-swallow → warn-and-continue.

kid7st (commented):
No findings. 上一轮的 tunnel 行为回归 finding 已按建议精确修复(仅 tunnel.ts + 新增 test/tunnel.test.ts 变动,重构其余部分不变、上轮已验证忠实):announceWebhooks 现用 try/catch 包住 loadDotEnv,非 ENOENT 读取故障经 log.warn 报出(rule 8)并继续 best-effort;loadDotEnv 本身保持 throw-on-non-ENOENT(命令调用方仍干净退出),companion 注释准确。回归测试到位且精确:把 .env 造成一个目录 → loadDotEnv 抛 EISDIR(真实非 ENOENT 触发,非 mock),断言 announceWebhooks resolves(不崩)且日志含 'could not read'(surface 而非静默)—— no-crash 契约与 rule-8 surfacing 两面都 pin 住。整份 tunnel.test.ts 通过(exit=0)。

PR #180, opened 2026-07-07
fix(host): make pump total — a mid-flight stream error no longer crashes the server

kid7st:
## What

Closes a latent crash on the core HTTP host: no request path can terminate the long-running `dev`/`start` server via an unhandled rejection. Found by a scan for the same class as the tunnel bug (#179).

## The bug

`pump` (the node:http adapter in `channels/http.ts`) is **void-called** by `nodeListener`, and the process installs **no `unhandledRejection` handler**. So any throw that escapes `pump` terminates the server.

`pump` previously guarded only **two** regions — a `handler()` throw (→ 500) and a mid-flight stream error — leaving the code **between** them unguarded:
- `response.headers.forEach(...)` — a non-Response return (a channel that forgot to `return` → `response.headers` undefined → TypeError). The common case, and exactly the "non-conformant channel" threat.
- `res.writeHead(response.status, ...)` — a header value Node rejects.
- `response.body.getReader()`.

A throw there escaped as an unhandled rejection **and**, being pre-headers, left the response hanging.

## The fix (the contract, not another patch)

Collapse to a **single `try/catch` totality boundary** wrapping the whole request→response→stream path. Every failure lands in one `catch` that branches on `res.headersSent`:
- **not sent** → a clean `500` + end (client can retry);
- **already streaming** → `destroy` the socket (truncated stream, not a hang), never leaking the internal message.

`pump`'s returned promise now **never rejects by construction** — so `void pump(...)` is safe *by contract*, stated in the docstring.

**Deliberately not** a global `unhandledRejection` net: robustness against a background throw is each fire-and-forget's own contract (fail into a terminal state), not a blanket swallow that would hide real bugs (rule 8). The scan audited every other `void`-async site — all already total/guarded (`turn-queue` task `try/catch`, `github`/`telegram` `.then(f,r)`, `channel.ts` async-factory guard, `dev-supervisor`'s `process.exit`).

## Tests
Two new regressions in `test/http.test.ts`:
- a **non-Response** handler return (the common "forgot to return" gap) → the client sees `500`, no crash;
- a body stream that **errors mid-flight** → response `destroy`ed + logged.

Neither escapes as an unhandled rejection (vitest would surface it). `npm run typecheck && npm test` — 461 passed.

## Review note
Thanks for catching that the first draft only closed half the totality (the stream loop, not the sibling gap between the handler catch and the stream try). This revision makes the single-boundary structure so "never rejects" is actually true.

### DIFF
--- a/src/channels/http.ts
+++ b/src/channels/http.ts
@@ -79,10 +79,21 @@ export function nodeListener(
   handler: (req: Request) => Promise<Response>,
 ): (req: IncomingMessage, res: ServerResponse) => void {
   return (req, res) => {
-    void pump(handler, req, res);
+    void pump(handler, req, res); // safe: pump is TOTAL (never rejects) — see its contract below
   };
 }
 
+/**
+ * Consume ONE request and drive its response to a terminal state. pump is TOTAL: a SINGLE try/catch wraps
+ * the whole request→response→stream path, so EVERY failure — a handler throw, a non-Response return
+ * (`response.headers` undefined), a header Node rejects, `getReader`, or a body stream that errors
+ * mid-flight — ends the response and the returned promise NEVER rejects, which is what lets the
+ * `void pump(...)` above be safe. Before any byte goes out (headers not sent) it is a clean 500; once the
+ * response is streaming, the only honest signal left is to destroy the socket (truncated stream, not a
+ * hang). The process installs no `unhandledRejection` handler by design: robustness against a background
+ * throw is each fire-and-forget's OWN contract (fail into a terminal HTTP response here), not a global net
+ * that would blanket-swallow.
+ */
 async function pump(
   handler: (req: Request) => Promise<Response>,
   req: IncomingMessage,
@@ -91,16 +102,14 @@ async function pump(
   const controller = new AbortController();
   res.on("close", () => controller.abort());
 
-  const method = req.method ?? "GET";
-  const hasBody = method !== "GET" && method !== "HEAD";
-  const headers = new Headers();
-  for (const [k, v] of Object.entries(req.headers)) {
-    if (Array.isArray(v)) for (const vv of v) headers.append(k, vv);
-    else if (v != null) headers.set(k, v);
-  }
-
-  let response: Response;
   try {
+    const method = req.method ?? "GET";
+    const hasBody = method !== "GET" && method !== "HEAD";
+    const headers = new Headers();
+    for (const [k, v] of Object.entries(req.headers)) {
+      if (Array.isArray(v)) for (const vv of v) headers.append(k, vv);
+      else if (v != null) headers.set(k, v);
+    }
     const request = new Request(`http://${req.headers.host ?? "localhost"}${req.url ?? "/"}`, {
       method,
       headers,
@@ -108,29 +117,20 @@ async function pump(
       duplex: "half",
       signal: controller.signal,
     } as RequestInit & { duplex: "half" });
-    response = await handler(request);
-  } catch (error) {
-    // Log server-side (some channels' only failure sink), but return a generic body — don't leak the
-    // internal message to the client.
-    log.error(`[host] request handler failed: ${String(error)}`);
-    if (!res.headersSent) res.writeHead(500, textHeaders);
-    res.end("internal error\n");
-    return;
-  }
+    const response = await handler(request);
 
-  const outHeaders: Record<string, string> = {};
-  response.headers.forEach((value, key) => {
-    outHeaders[key] = value;
-  });
-  res.writeHead(response.status, outHeaders);
+    const outHeaders: Record<string, string> = {};
+    response.headers.forEach((value, key) => {
+      outHeaders[key] = value;
+    });
+    res.writeHead(response.status, outHeaders);
 
-  if (!response.body) {
-    res.end();
-    return;
-  }
-  const reader = response.body.getReader();
-  res.on("close", () => void reader.cancel());
-  try {
+    if (!response.body) {
+      res.end();
+      return;
+    }
+    const reader = response.body.getReader();
+    res.on("close", () => void reader.cancel());
     for (;;) {
       const { done, value } = await reader.read();
       if (done || res.destroyed) break;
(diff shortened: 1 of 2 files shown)


kid7st (commented):
本次改动核心:让 pump TOTAL —— 总是把响应驱动到终态、返回 promise 永不 reject,从而 void pump(...) by contract 安全。方向正确(关闭 pump 自身 totality,而非外部 .catch),流式循环的 catch + destroy + fail-visible log 都对,测试也 pin 住了 mid-flight stream error 一路。经 git 对象核验(本会话 stdout 有污染,只信任 git show 抽取)。

**一条 finding(位置在未改动的上下文行,GitHub 无法 inline,故写在这里):src/channels/http.ts:129-138 —— totality 只关了一半。**

docstring/PR 声称 'pump is TOTAL … its returned promise NEVER rejects',但这次只包住了流式循环。handler 的 try/catch(→500,约 123 行结束)与流式 try(140 行)之间,有三行在**任何 catch 之外**;在那里抛出会连流式的 finally 都跳过,于是没有任何东西 end 响应、pump reject → 正是要消除的 void pump(...) unhandled-rejection 崩溃:

- **129 `response.headers.forEach(...)`**:一个 non-conformant handler 返回(非 Response:undefined / 普通对象)使 `response.headers` 为 undefined → TypeError。这恰是 PR 自己点名的威胁('a non-conformant Agent that violates SPEC MUST 2'),而『忘了 return / 返回错类型』比 mid-flight stream error 更常见。
- **132 `res.writeHead(response.status, outHeaders)`**:自定义 channel 的 Response 带一个 Node 拒绝的 header 值时抛出。
- **138 `response.body.getReader()`**。

新测试只覆盖循环内(第 2 次 pull)的 stream error,这个 sibling 区域未测。与被修的 bug 同类、就在同一函数里 —— 修了上报的那条路径,没修根(pump totality)。

方向:把 guard 上移覆盖整个 post-handler 区域 —— try 从 129 行(handler catch 之后)开始;catch 按 `res.headersSent` 分支:未发送 → `writeHead(500)` + end;已发送 → log + destroy(现有 catch 行为)。这样 'NEVER rejects' 才真正成立。

kid7st:
Addressed in the amended commit `1ff3c66`. You were right — the first draft only closed half the totality (the stream loop), leaving the sibling gap between the handler catch and the stream try (`response.headers.forEach` on a non-Response, `writeHead`, `getReader`) — the *more* common path (a channel that forgot to `return`).

Fix: collapsed to a **single try/catch totality boundary** over the whole request→response→stream path, branching on `res.headersSent` (not sent → 500 + end; streaming → destroy). `pump`'s promise now never rejects *by construction*, so the docstring's 'NEVER rejects' is actually true. Added a non-Response→500 regression (the gap you named) plus the mid-flight one. This was patch-then-patch twice on the same class; the revision derives the fix from the total contract (one boundary = all exits) instead of chasing the next unguarded line.

src/channels/http.ts
```diff
@@ -139,17 +139,27 @@ async function pump(
...
     }
-  } finally {
-    if (!res.destroyed) res.end();
+    if (!res.destroyed) res.end(); // normal completion
+  } catch (error) {
+    // The ONE totality boundary: every failure above lands here, so pump never rejects (see the header
+    // doc). Don't leak the internal message to the client. Not-yet-streaming → a clean 500; already
+    // streaming (headers sent) → destroy the socket (truncated, not a hang).
+    log.error(`[host] request failed: ${String(error)}`);
+    if (!res.headersSent) {
```
kid7st: catch 的两个终止分支对 `res.destroyed` 处理不对称, 削弱了 totality 不变量.

"pump 永不 reject" 依赖 catch 自身也不 throw. streaming 分支写 `else if (!res.destroyed) res.destroy(...)` 显式守卫了 destroyed; 但这个 pre-header 500 分支只判 `!res.headersSent`, 没有配套的 `!res.destroyed`.

路径: 客户端在 handler await 期间断开 → `res` 'close' → `controller.abort()` → handler 抛 AbortError → 落入 catch, 此时 `res.headersSent === false` 且 `res.destroyed === true` → 对已销毁的 `res` 调 `writeHead(500)` + `end(...)`. 若这组写入在 destroyed socket 上同步 throw, pump 就 reject —— 正是本 PR 要消灭的 unhandled rejection; 即便不 throw, 也是往死 socket 的无效写入.

建议: 500 分支同样守卫 destroyed, 与 streaming 分支对称 —— `if (!res.headersSent && !res.destroyed) { writeHead(500); end(...) }`, 让 "catch 内绝不对 destroyed 的 res 动手" 成为显式规则, 而非隐含假设 writeHead/end 在 destroyed 上不会 throw.

kid7st (commented):
Thesis: pump 收敛为单一 try/catch totality 边界, 承诺返回的 promise 永不 reject, 使 `void pump(...)` by contract 安全. 一处 finding (见 inline), 落在 diff 内.

src/channels/http.ts
kid7st: Fixed in the amended commit (a35218c). Verified the path: client disconnect during the handler await → res close → controller.abort() → a signal-aware handler throws AbortError → catch with res.destroyed=true, headersSent=false → the old 500 branch would writeHead/end a dead socket → ERR_STREAM_DESTROYED thrown INSIDE the catch → pump rejects → the exact crash.

Rather than repeat `!res.destroyed` in both branches, I made the invariant one named gate at the top: `if (res.destroyed) return;` then headersSent decides destroy-vs-500. With it the catch is provably non-throwing (writeHead only when !headersSent && !destroyed; destroy is idempotent), so no nested try/catch is needed. Added a regression: res pre-destroyed + a writeHead/end stub that throws if touched → asserts the catch early-returns and never touches the dead socket (a regression re-surfaces as the unhandled rejection vitest fails on).

This was the 3rd round on this class — the first two fixed which business path throws; this one is the cleanup code itself throwing. Thanks for holding the totality invariant to "the catch must not throw either."

kid7st (commented):
No findings.

Thesis: pump 收敛为单一 try/catch totality 边界, 返回 promise 永不 reject, 使 `void pump(...)` by contract 安全.

上一轮 finding (catch 两个终止分支对 `res.destroyed` 不对称, pre-header 500 分支可能对已销毁 socket writeHead/end → catch 自身 throw → 重新引入 unhandled rejection) 已解决: catch 现以 `if (res.destroyed) return` 单一 gate 打头, writeHead/end 仅在 `!destroyed && !headersSent` 时执行, destroy 幂等, 故 catch 可证不 throw. 新增测试 `catch never writes to an already-destroyed res` 以 throwing stub 守住此不变量 (去掉 gate 即 fail, 非 tautological). 全量重扫无新问题.

PR #186, opened 2026-07-07
feat(schedule): agent self-scheduling — the wake tool (Phase 4a)

kid7st:
## What

Phase 4a of the scheduler: **agent self-scheduling** (opt-in). The built-in `wake` tool lets an agent schedule *itself* — a turn calls `wake({ in, prompt })` and, after the delay, a new turn runs **back in the same session**, so the agent resumes the conversation it was in.

```
The agent (mid-conversation):  wake({ in: "10m", prompt: "check whether the deploy finished" })
  → 10 min later, a new turn runs in THIS session with that prompt; the agent has the full context.
```

Second **producer** of scheduled invocations (the first, Phase 1, is the author's `schedules/` files). Both feed one scheduler; different reliability contracts, so both stay.

## Design

- **Opt-in, not default.** Enable with `selfSchedule: true` in `fastagent.config`. Self-scheduling is an *autonomy* capability (it changes what the agent IS, not just a tool it has), so it's an explicit choice — not silently given to every served agent, and never mounted on the one-shot `invoke`/`fire` (they exit and never poll → a promise nothing fulfills).
- **`ToolContext.session` is a public tool capability** (already exported). A `defineTool` tool is built once and reused across sessions, so the session can't be a closure — it rides `AsyncLocalStorage` set around the harness turn (`invoke`) and read in `execute`. Any tool can use it (which conversation am I in — partition data, tag logs); `wake` is one consumer. Proven to propagate through pi's tool execution (`test/tool-context.test.ts`).
- **Replay-safe re-fire.** A wake into a **busy** session (a channel mid-turn on the same id — the common "wake me in 10 min while I'm still chatting" case) is deferred + retried, since the turn **never started**. Every other failure is terminal: re-running a turn that *did* start could duplicate its tools' side effects. Busy is told apart by the failed event's **`code`**, not a `details` text match.
- **The woken turn's reply isn't delivered** — the `wake` description tells the model it must use a channel send tool to reach the user (same as any scheduled job; output is the agent's tools' job).
- **Guardrails** — minimum delay, **per-session** pending cap (one conversation can't DoS others), bounded busy-retry. Claim-one-fire-one; malformed entries / wrong store shape fail-visible. Total poller (chained `setTimeout`, state-IO fault doesn't kill the chain).

## Contract extension (SPEC §8, adopted — please review)

[...]

PR #188, opened 2026-07-07
fix(telegram): busy-wait a user turn colliding with an external lease holder (wake)

kid7st:
## What

Closes the **reverse collision** left open by #186: while a self-scheduled wake turn holds a session's lease, a user message on the same session collided and got '⚠️ Temporary problem — please try again' — a user-visible error in wake's flagship scenario ('wake me in 10 min while I'm still chatting'). #186 handled wake→busy (scheduler defers); this handles user→busy (channel waits).

## How

telegram's own turns never collide (the turn-queue serializes per session), so a busy reject in `invokeTurn` is always an **external** holder. A fail-fast `session_busy` **first** event (the turn never started — replay-safe, discriminated by `failed.code` from #186) is now retried boundedly (5 s pace, 2 min cap) instead of surfaced: attachments are already resolved (no re-download), the user sees the Thinking… placeholder, then the answer. Only a first-event busy retries — nothing that started is ever re-run. An exhausted wait surfaces the busy failure as before.

Scoped to telegram; github/http colliding with a wake still fail fast (noted in core.md — their sessions rarely mix with wake sessions; add the same wait if it bites). The structural alternative (one shared per-session serial seam) stays deferred: two bounded waits cover the real scenario at a fraction of the seam's cost.

## Test
4 new units on `invokeTurn` (injectable retry pace): busy→retry→answer (no error surfaced); exhausted wait surfaces busy; non-busy failure immediate (no replay); mid-stream busy passed through (started turn never re-run). 513 total.

### DIFF
--- a/src/channels/telegram/invoke-turn.ts
+++ b/src/channels/telegram/invoke-turn.ts
@@ -5,7 +5,7 @@
  * pure) because this half touches the Bot API + disk; split from telegram.ts so the factory keeps only
  * wiring and the per-turn lifecycle.
  */
-import type { Agent, AgentEvent, ImageRef } from "../../agent.ts";
+import { type Agent, type AgentEvent, type ImageRef, SESSION_BUSY_CODE } from "../../agent.ts";
 import { log } from "../../log.ts";
 import type { BufferedRef } from "./context-buffer.ts";
 import { type DownloadedFile, resolveFiles, resolveImages } from "./telegram-api.ts";
@@ -101,6 +101,16 @@ async function resolveTurnAttachments(t: TurnTransport, attachments: TurnAttachm
   return { images: allImages.length ? allImages : undefined, promptSuffix: `${bufferedNote}${manifest}` };
 }
 
+/** How the busy-wait paces: retry the invoke every `delayMs` while the session's lease is held by an
+ *  EXTERNAL turn (a self-scheduled wake, a concurrent embedder invoke), up to `maxWaitMs` total. The
+ *  channel's own turns never collide (the turn-queue serializes per session), so a busy reject here is
+ *  always an outside holder — wait for it like a queued turn, instead of erroring at the user. */
+export interface BusyRetry {
+  delayMs: number;
+  maxWaitMs: number;
+}
+const DEFAULT_BUSY_RETRY: BusyRetry = { delayMs: 5_000, maxWaitMs: 120_000 };
+
 /**
  * Run one turn: resolve its attachments, then stream agent.invoke. A primary-attachment failure surfaces
  * as a `failed` event (never a silent drop). `onCompleted` (if given) fires on the turn's `completed`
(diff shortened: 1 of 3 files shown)


kid7st (commented):
Thesis: `invokeTurn` gains a bounded busy-wait on a first-event `failed{code: session_busy}` (fail-fast reject, turn never started) — the mirror of #186's wake-defer, closing the user→busy direction for telegram. The core concept is clean: the `first` gate, the deadline arithmetic, and the commit point (`onCompleted`) inside the retry loop all check out, and the "Thinking… placeholder is visible while waiting" claim matches `preview.ts` (`touch()` sends it before any event).

One finding, anchored inline on `DEFAULT_BUSY_RETRY`.

src/channels/telegram/invoke-turn.ts
```diff
@@ -101,13 +101,30 @@ async function resolveTurnAttachments(t: TurnTransport, attachments: TurnAttachm
...
+/** How the busy-wait paces: retry the invoke every `delayMs` while the session's lease is held by an
+ *  EXTERNAL turn (a self-scheduled wake, a concurrent embedder invoke), up to `maxWaitMs` total. The
+ *  channel's own turns never collide (the turn-queue serializes per session), so a busy reject here is
+ *  always an outside holder — wait for it like a queued turn, instead of erroring at the user. */
+export interface BusyRetry {
+  delayMs: number;
+  maxWaitMs: number;
+}
+const DEFAULT_BUSY_RETRY: BusyRetry = { delayMs: 5_000, maxWaitMs: 120_000 };
+
```
kid7st, line 118: **`maxWaitMs: 120_000` does not cover the scenario the PR claims to close, and core.md overstates the coverage.**

The success criterion is "the user never sees the busy error in wake's flagship scenario", but a 2-minute cap only holds when the lease-holding turn finishes within 2 minutes. A tool-using wake turn routinely runs longer — and then the user gets exactly the "Temporary problem — please try again" this PR set out to eliminate. Meanwhile `docs/design/core.md` now reads "**Both collision directions are handled** for telegram", with no mention of this ceiling, and production has no knob (`busyRetry` is injectable only from tests; the `telegram.ts` call site uses the default).

Waiting is nearly free — each retry is a lease-check-level reject (`tryAcquire` runs before harness assembly), one per 5 s. Either size `maxWaitMs` to the same order as a real wake turn (e.g. 10 min), or scope the core.md claim to "handled for holders shorter than maxWaitMs" and name the ceiling + its exhausted-wait consequence in the `DEFAULT_BUSY_RETRY` comment.

kid7st: Fixed: raised maxWaitMs to 10 min (each retry is a lease-check-level reject — waiting is nearly free and ends within one delay of the holder finishing, so the cap is sized to outlast a real tool-using wake turn). Named the ceiling + its exhausted-wait consequence in the DEFAULT_BUSY_RETRY comment, and core.md now states the bound and why it exists (a stuck lease must not hang a chat turn forever) instead of the unqualified claim.

PR #189, opened 2026-07-07
feat(deploy): keep one machine running when time triggers exist (Phase 3)

kid7st:
## What

Phase 3 of the scheduler: `deploy` now enforces **keep-1-machine** when the agent has **time triggers** — `schedules/` files (under agentDir) or `config.selfSchedule` (the wake tool). Cron/wake has no external wake-up: unlike a webhook, nothing wakes a scaled-to-zero box for a cron instant, so it would sleep through every fire.

## How

- **preflight** detects `hasTimeTriggers` (host-neutral fact) + a conditional note.
- **fly**: generated `fly.toml` forces `min_machines_running = 1` (reason-tagged, same pattern as the github no-replay floor). **KEEP mode covered**: a kept fly.toml that still scales to zero — explicit `0` **or the line absent** (Fly's platform default is 0) — gets a warn with the fix (`set 1` / `--force`); the author who deployed first and added schedules later hits exactly that path.
- **railway**: the runbook's App Sleeping line flips to 'do NOT enable' (dashboard-only setting — stated, like the github floor).
- `hasTimeTriggers` is **required** on both plan inputs (a caller can't silently omit it and degrade the guidance).

## Test
preflight (none/schedules/selfSchedule matrix + note), fly plan (min=1 line), railway runbook (do-NOT-enable line). 514 total.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -661,15 +661,23 @@ async function runDeploy(): Promise<void> {
   // KEEP mode + time triggers: the kept fly.toml may still scale to zero — which would sleep through every
   // cron instant / wake-up. The generated plan can't fix a kept file, so surface it instead of the preflight
   // note silently not applying (the author who deployed FIRST and added schedules LATER hits exactly this).
+  // Under `--run` this is a GATE (same discipline as the model-travel gate): a full deploy whose schedules
+  // silently never fire is worse than a crash-loop — nothing fails visibly when a cron instant passes on a
+  // sleeping machine, and unlike github's min=0 there is no legitimate trade to accept here.
   if (flyTomlExists && !values.force && hasTimeTriggers) {
     const min = parseFlyMinMachines(await readFile(flyTomlPath, "utf8"));
     if ((min ?? 0) === 0) {
       // undefined = the line is absent — Fly's platform default for min_machines_running is 0, so a
       // hand-written fly.toml without the line scales to zero exactly like an explicit 0.
-      console.error(
-        `[fastagent] warn: your kept fly.toml has min_machines_running = 0, but schedules/self-scheduling ` +
-          `need a running machine (no external wake-up). Set it to 1, or pass --force to regenerate.`,
-      );
+      const msg =
+        `your kept fly.toml scales to zero (min_machines_running = ${min ?? "absent → platform default 0"}), but ` +
+        `schedules/self-scheduling need a running machine (no external wake-up). Set min_machines_running = 1, ` +
+        `or pass --force to regenerate.`;
+      if (values.run) {
+        console.error(`[fastagent] deploy stopped: ${msg}`);
+        process.exit(1);
+      }
+      console.error(`[fastagent] warn: ${msg}`);
     }
   }
   const plan = planFlyDeploy({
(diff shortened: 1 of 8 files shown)


src/cli.ts
```diff
@@ -652,12 +658,27 @@ async function runDeploy(): Promise<void> {
         `was kept. Edit auto_stop_machines/min_machines_running in fly.toml, or pass --force to regenerate.`,
     );
   }
+  // KEEP mode + time triggers: the kept fly.toml may still scale to zero — which would sleep through every
+  // cron instant / wake-up. The generated plan can't fix a kept file, so surface it instead of the preflight
+  // note silently not applying (the author who deployed FIRST and added schedules LATER hits exactly this).
+  if (flyTomlExists && !values.force && hasTimeTriggers) {
```
kid7st, line 667: `--run` + kept min=0 + time triggers: warn-not-gate contradicts the repo's own `--run` discipline, and the choice is undeclared.

Preflight's two known-broken deploys (model won't travel, agentDir with its own package.json) follow "known crash → gate `--run`, warn else". A kept fly.toml with min=0 plus time triggers, under `--run`, fully deploys a box whose schedules sleep through every fire — arguably worse than a crash-loop: a crash is visible, this is a production silent no-op (no failed event when the cron instant passes on a sleeping machine), which is exactly what "fail visibly" forbids. And unlike the github case, there is no legitimate trade here: github's min=0 has one ("accept occasionally losing an in-flight review"), but "accept that schedules never fire" is not a trade anyone means to take. The PR body never mentions the `--run` interaction, so this reads as a silent decision.

Direction: under `--run`, escalate this to a gate (exit with the fix: `set min_machines_running = 1`, or `--force`), mirroring the model-travel gate. If warn is deliberate (preserving operator override), state why this deviates from the `--run` gate discipline — in the code comment and the PR.

kid7st (commented):
Thesis: `hasTimeTriggers` (schedules/ or `config.selfSchedule`) as a host-neutral preflight fact → fly forces `min_machines_running=1`, railway runbook forbids App Sleeping, plus a KEEP-mode warn. Concept is clean; detection dir (agentDir) matches runtime (`startSchedules(agentDir)`); required plan inputs prevent silent omission.

One finding that can't anchor to a diff line:

**The PR's headline scenario (KEEP mode) has zero tests.** The body calls it "KEEP mode covered", but coverage exists only in code: `parseFlyMinMachines` (src/deploy/fly/plan.ts:218) is a new exported parser with three semantic branches — explicit `0`, absent line → `undefined` → treated as 0 (a judgment that encodes Fly's platform default, exactly the kind of external-behavior assumption a test should pin), and `>= 1` — and none are tested. Its sibling `parseFlyAppName` IS tested (test/deploy-fly.test.ts:151-153), so the new parser breaks the file's own pattern. All three new test blocks cover only the GENERATED path; the "deployed first, added schedules later" path the PR names as its motivating scenario is the one path with no protection. Direction: add sibling-style unit tests for `parseFlyMinMachines` (explicit 0 / `= 1` with trailing comment / absent line → undefined).

src/cli.ts
kid7st: Fixed: under --run this is now a GATE (deploy stopped + exit 1, same discipline as the model-travel gate — a full deploy whose schedules silently never fire is worse than a crash-loop, and unlike github's min=0 there's no legitimate trade). Warn-only remains for generate-only (runbook). Also added the sibling-style parseFlyMinMachines unit tests (explicit 0 / = 1 with trailing comment / absent line → undefined, with the caller-treats-absent-as-0 judgment pinned in a comment) — the body finding.

PR #191, opened 2026-07-07
feat(schedule): run audit — runs.jsonl + fastagent schedule history (Phase 2)

kid7st:
## What

Phase 2 of the scheduler: a **run audit**. Every fire (cron **and** self-scheduled wake) appends one line to `<stateRoot>/schedule/runs.jsonl`; `fastagent schedule history <name>` (or `wake`) answers cron's classic pain — **did last night's run silently fail?**

## Design

- **Append-only JSONL, full reply** (the session-store shape; storage is cheap). A run record is an immutable per-fire snapshot — the rolling session, where fires interleave with user turns, cannot give you that per-fire. Operational audit only; conversational truth stays in the session store.
- **Total**: an audit-write failure is logged, never breaks a fire.
- **Audit honesty**: a busy wake dropped at the retry ceiling is `failed` (a FINAL silent loss — exactly what the audit exists to catch), not `deferred`; `deferred` only when actually re-scheduled.
- **`wake` is a reserved schedule name** (discover rejects `schedules/wake.ts`) so the audit namespace can't be polluted.
- History text mode **tails the last 20** (the question is about last night, not month one); `--json` = full records incl. complete reply text.

## Test
audit roundtrip (full reply preserved), name filter, malformed-line skip; scheduler records completed/deferred; CLI usage branch. 518 total.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -352,6 +352,7 @@ function runScheduleCmd(): void {
     process.exit(2);
   }
   const target = resolve(positionals[3] ?? ".");
+  loadDotEnv(target); // FASTAGENT_STATE_DIR may live in .env — read the SAME state root the scheduler wrote
   const runs = readRuns(resolveStateRoot(target), name);
   if (values.json) {
     console.log(JSON.stringify(runs, null, 2));
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -73,6 +73,7 @@ import { spawnRunner } from "./deploy/runner.ts";
 import { assembleSecrets } from "./deploy/secrets.ts";
 import { registerTelegramWebhook } from "./channels/telegram/register-webhook.ts";
 import { discoverScheduleFiles, loadSchedules } from "./schedule/discover.ts";
+import { readRuns } from "./schedule/audit.ts";
 import { createScheduler, scheduleSession } from "./schedule/scheduler.ts";
 
 function usage(code: number): never {
@@ -83,6 +84,7 @@ function usage(code: number): never {
   fastagent tool   <name> '<json-args>' [dir]
   fastagent invoke <message> [dir] [--model provider/modelId] [--auth-path file]
   fastagent fire   <name> [dir] [--model provider/modelId] [--auth-path file]
+  fastagent schedule history <name> [dir] [--json]
   fastagent dev    [dir] [--port N] [--model provider/modelId] [--auth-path file] [--no-watch] [--tunnel]
   fastagent chat   [dir] [--model provider/modelId]
   fastagent start [dir] [--port N] [--model provider/modelId] [--sessions-dir dir] [--auth-path file] [--tunnel]
@@ -125,6 +127,9 @@ function usage(code: number): never {
          counterpart of tool, for CI smoke and quick checks. Same model resolution as dev.
   fire   run ONE schedule's turn immediately (authoring loop, like invoke) — fires schedules/<name>.ts
          now without waiting for its cron. Reply→stdout; does NOT advance the schedule's fire state.
+  schedule history <name>  print the run audit for a schedule (or "wake" for self-scheduled wake-ups):
+         when each run fired, completed/failed/deferred, duration, and the reply/error — the answer to
+         "did last night's run silently fail?". --json for the full records (complete reply text).
   start  run the agent in dir (default .) in production posture — the SAME assembly as dev
          (your directory is the agent), just no file-watching. No build step: start reads the
          definition directly; model/http come from fastagent.config.ts (frozen by git).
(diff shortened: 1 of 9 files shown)


src/cli.ts
```diff
@@ -333,6 +339,42 @@ async function runFire(): Promise<void> {
...
+ * to "did last night's run silently fail?". Text mode previews the reply/error; --json is the full record.
+ */
+function runScheduleCmd(): void {
+  const sub = positionals[1];
+  const name = positionals[2];
+  if (sub !== "history" || !name) {
+    console.error(`usage: fastagent schedule history <name> [dir] [--json]`);
+    process.exit(2);
+  }
+  const target = resolve(positionals[3] ?? ".");
```
kid7st, line 354: `runScheduleCmd` never calls `loadDotEnv`, so it can read the wrong state root. `resolveStateRoot` honors `FASTAGENT_STATE_DIR`, and every other state-touching command (`tool`/`invoke`/`fire`/`info`/`login`/`dev`/`start`) loads `.env` before resolving paths. If a project sets `FASTAGENT_STATE_DIR` in `.env`, the serving scheduler (which loads `.env`) writes runs.jsonl to the env-configured root while `schedule history` reads `<dir>/.fastagent` and prints "no recorded runs" — the audit command silently reading the wrong ledger is exactly the failure class this audit exists to eliminate. Fix: `loadDotEnv(target)` before `resolveStateRoot(target)`.

src/schedule/scheduler.ts
```diff
@@ -60,26 +61,32 @@ export function createScheduler({ agent, stateRoot, schedules, now = () => new D
    *  failed specifically because the session was BUSY (the turn never started) — the ONLY replay-safe reason
    *  to re-fire a wake-up; every other outcome (completed, or a failure whose side effects may have run) is
    *  terminal. */
-  async function runTurn(label: string, session: string, prompt: string): Promise<{ busy: boolean }> {
+  async function runTurn(
```
kid7st, line 64: The docstring above still says "Returns whether the turn failed specifically because the session was BUSY", but this change widened the return to `{ busy, failed?, reply, ms }`. The doc describes the old contract and misleads the next reader — update it to name the audit-carrying fields too.

test/scheduler.test.ts
```diff
@@ -122,6 +126,8 @@ describe("schedule/scheduler: fire algorithm", () => {
     // NOT dropped: re-scheduled (deferred) with a bumped attempt count — a one-shot wake must not vanish.
     await vi.waitFor(() => expect(listWakeups(root)).toHaveLength(1));
     expect(listWakeups(root)[0]).toMatchObject({ session: "busy", attempts: 1 });
+    // Audited as deferred (not failed): honest — the wake was re-scheduled, not finally lost.
+    expect(readRuns(root, "wake")[0]).toMatchObject({ outcome: "deferred", session: "busy" });
```
kid7st, line 130: Test gap: the `failed` outcome is never asserted through the scheduler path — only `completed` (cron) and `deferred` (busy wake) are. Untested branches: (1) a failed cron fire → `{ outcome: "failed", error, reply: undefined }` — the existing "failed turn still claims the slot" test already constructs this scenario, it just lacks one `readRuns` assertion; (2) a busy wake dropped at the retry ceiling → `failed` + `"dropped after too many busy retries"` — the PR's headline audit-honesty decision, and the only branch with no test protection. The audit exists to answer "did it fail?"; its fail path is the untested one.

kid7st (commented):
Thesis: run audit — one immutable per-fire record (full reply) appended to `<stateRoot>/schedule/runs.jsonl` for every cron fire and wake, read by `fastagent schedule history`. Core concept is clear; write path (total append) and read path (fail-visible skip) have correct responsibility assignment. 3 findings, all inline.

src/cli.ts
kid7st: Fixed in 185580b: loadDotEnv before resolveStateRoot (same-root guarantee); runTurn docstring names the audit-carrying return fields; added the two fail-path audit tests (failed cron fire → outcome failed + error, no reply copy; busy wake dropped at the retry ceiling → audited FAILED with 'dropped after too many busy retries', store empty).

src/schedule/scheduler.ts
kid7st: Fixed in 185580b: loadDotEnv before resolveStateRoot (same-root guarantee); runTurn docstring names the audit-carrying return fields; added the two fail-path audit tests (failed cron fire → outcome failed + error, no reply copy; busy wake dropped at the retry ceiling → audited FAILED with 'dropped after too many busy retries', store empty).

test/scheduler.test.ts
kid7st: Fixed in 185580b: loadDotEnv before resolveStateRoot (same-root guarantee); runTurn docstring names the audit-carrying return fields; added the two fail-path audit tests (failed cron fire → outcome failed + error, no reply copy; busy wake dropped at the retry ceiling → audited FAILED with 'dropped after too many busy retries', store empty).

PR #193, opened 2026-07-08
fix(schedule): make the cron fire path total; move loader.ts to the neutral top level

kid7st:
## What

Follow-ups from a full design/implementation review of the scheduler.

1. **Fire-path totality (the real bug).** `fireThenReArm` is void-scheduled from a timer, but `fire()`'s `loadFires` throws by design on an unreadable `fires.json` (non-ENOENT IO fault) — an escaping throw was an **unhandled rejection**: one broken state file crashing the whole service, channels included. It now has the same totality boundary `pumpWakeups` already had (the same file had one void-scheduled path guarded and one bare): log, audit, skip the run, keep the schedule armed. Boot-time faults still fail `start()` loudly. Both skip paths (fire-state fault, unpersistable claim) now **append a failed run record** — `runs.jsonl` is a different file than the broken `fires.json`, so a skipped fire shows up in `schedule history`, not only stderr.
2. **`loader.ts` moves `engines/pi/` → `src/`.** Pure node stdlib (ESM discovery/loading for tools/channels/schedules/config); `schedule/discover.ts` importing it from `engines/` made the layering read as an engine coupling it never was. Import-path-only change for 7 files.
3. **Companion text**: `deferWakeup` documents the accepted microtask-scale one-shot resurrection window (cousin of the recurring one the advance-in-place claim eliminated); core.md states `runs.jsonl` is unbounded by design.

## Test

Fault-injection test: a directory at `fires.json` (EISDIR — the unreadable-state class) at fire time → the run is skipped **and audited**, no unhandled rejection, and the schedule stays armed (fires normally once the operator fixes the state). 526 total.

### DIFF
--- a/src/schedule/scheduler.ts
+++ b/src/schedule/scheduler.ts
@@ -91,27 +91,14 @@ export function createScheduler({ agent, stateRoot, schedules, now = () => new D
   }
 
   /** Fire one schedule's turn: claim the slot (persist lastFired BEFORE invoking) so a crash mid-turn
-   *  does not re-fire this slot on restart, then run the turn. */
+   *  does not re-fire this slot on restart, then run the turn. A state fault while claiming (loadFires or
+   *  saveFires — both before the invoke, so nothing ran) THROWS to the single skip+audit boundary in
+   *  {@link fireThenReArm}; skipping rather than firing unclaimed also avoids an infinite catch-up loop
+   *  on restart. */
   async function fire(s: LoadedSchedule): Promise<void> {
     const fires = loadFires(stateRoot);
     fires[s.name] = now().toISOString();
-    try {
-      saveFires(stateRoot, fires);
-    } catch (e) {
-      // Can't persist the claim → skip this fire rather than risk an infinite catch-up loop on restart.
-      // Still audited: `appendRun` is total and writes a DIFFERENT file (runs.jsonl), so a broken
-      // fires.json must not also make the skip invisible to `schedule history`.
-      log.error(`[schedule] ${s.name}: cannot persist fire state, skipping this run: ${String(e)}`);
-      appendRun(stateRoot, {
-        name: s.name,
-        session: scheduleSession(s.name),
-        firedAt: now().toISOString(),
-        ms: 0,
-        outcome: "failed",
-        error: `run skipped — cannot persist fire state: ${String(e)}`,
-      });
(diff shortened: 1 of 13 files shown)


kid7st (commented):
Thesis: `fireThenReArm` gains the totality boundary `pumpWakeups` already had (state-IO fault at fire time → log + audit + skip, schedule stays armed), and `loader.ts` moves to the neutral top level. One finding (inline).

src/schedule/scheduler.ts
```diff
@@ -99,7 +99,17 @@ export function createScheduler({ agent, stateRoot, schedules, now = () => new D
       saveFires(stateRoot, fires);
     } catch (e) {
       // Can't persist the claim → skip this fire rather than risk an infinite catch-up loop on restart.
+      // Still audited: `appendRun` is total and writes a DIFFERENT file (runs.jsonl), so a broken
+      // fires.json must not also make the skip invisible to `schedule history`.
       log.error(`[schedule] ${s.name}: cannot persist fire state, skipping this run: ${String(e)}`);
+      appendRun(stateRoot, {
```
kid7st: `fire()` has exactly one caller — `fireThenReArm`, which now owns an equivalent totality boundary. With this PR, the inner `saveFires` catch and the outer catch handle the same semantic event (state fault → skip this run, audit failed, stay armed) with bit-identical behavior: the throw happens before `runTurn`, so letting it propagate also prevents an unclaimed invoke, and the outer catch audits + re-arms the same way. Instead of consolidating, the PR duplicated the 9-line `appendRun` block — two near-identical audit sites in one file, a double miss point next time `RunRecord` changes shape.

Direction: delete the inner try/catch and let the single boundary in `fireThenReArm` own skip+audit; if the more specific diagnostic matters, rethrow with context (`cannot persist fire state: ...`) rather than duplicating the audit.

kid7st: Fixed in 776b8da: the inner try/catch is gone — fire() lets a claim-time state fault (loadFires or saveFires, both before the invoke) propagate to the single skip+audit boundary in fireThenReArm; the duplicated appendRun block is deleted and fire()'s docstring names the boundary. No rethrow-with-context needed — the fs error already carries syscall+path, and the outer audit line prefixes 'run skipped — fire failed'.

PR #196, opened 2026-07-08
feat(schedule): envelope the woken turn; actionable guardrails; info shows the full time-trigger surface

kid7st:
## What

Agent-experience follow-ups from the cron/scheduler DX review — the served agent's side of the product.

1. **The woken turn's prompt arrives enveloped**: `[wake-up <id> fired — YOUR self-scheduled turn, not a user message] …` (a recurring adds its cron + an `unwake` hint). Without it the model sees its own instruction as a bare user message — in a chat session it may answer a "user" who said nothing — and has no way to know the wake-up's id (buried in a long-past tool result), which `unwake` needs. Static cron fires stay raw on purpose: their prompt is the author's instruction in a dedicated `schedule:<name>` session where every turn is a fire.
2. **Actionable guardrails**: the pending-cap rejection lists the pending ids ("unwake one" is only actionable if the model HAS them); the one-shot success message carries the unwake hint too (recurring already did).
3. **`fastagent info` shows the full time-trigger surface**: schedules are loaded (imported + validated, like tools) instead of filename-listed — a broken schedule file is reported at info time, not first at `dev`; each schedule prints its next fire instant; a `selfSchedule` line shows whether the wake tool mounts.

**Note**: `info --json` `schedules` changes shape (`string[]` → `{name, cron, tz, next}[]`) and gains `scheduleFailures` + `selfSchedule` keys. `info --json` is a diagnostic surface, not a stable API — called out here so the choice is conscious.

## Test

Envelope format (one-shot regex; recurring contains id + unwake hint), cap rejection contains every pending id. 531 green.

### DIFF
--- a/src/schedule/wakeups.ts
+++ b/src/schedule/wakeups.ts
@@ -136,9 +136,17 @@ export function addWakeup(
   const all = load(stateRoot);
   const mine = all.filter((w) => w.session === input.session);
   if (mine.length >= MAX_PENDING_WAKEUPS) {
-    // List what's pending: "unwake one" is only actionable if the model HAS the ids (they were returned
-    // when set, but that may be buried far back in the conversation).
-    const pending = mine.map((w) => `${w.id}${w.cron ? ` (recurring "${w.cron}")` : ""} at ${w.fireAt}`).join("; ");
+    // List what's pending WITH a prompt preview: "unwake one" is only actionable if the model has the
+    // ids AND can choose by meaning — both were returned when set, but that may be buried far back in
+    // the conversation.
+    const pending = mine
+      .map(
+        (w) =>
+          `${w.id}${w.cron ? ` (recurring "${w.cron}")` : ""} at ${w.fireAt}: ${
+            w.prompt.length > 60 ? `${w.prompt.slice(0, 60)}…` : w.prompt
+          }`,
+      )
+      .join("; ");
     return {
       ok: false,
       error: `too many pending wake-ups for this conversation (${MAX_PENDING_WAKEUPS}) — wait for some to fire, or unwake one. Pending: ${pending}`,
--- a/src/schedule/wakeups.ts
+++ b/src/schedule/wakeups.ts
@@ -134,10 +134,14 @@ export function addWakeup(
     fireAtDate = input.fireAt;
   }
   const all = load(stateRoot);
-  if (all.filter((w) => w.session === input.session).length >= MAX_PENDING_WAKEUPS) {
+  const mine = all.filter((w) => w.session === input.session);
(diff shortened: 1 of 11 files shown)


src/schedule/wakeups.ts
```diff
@@ -134,10 +134,14 @@ export function addWakeup(
     fireAtDate = input.fireAt;
   }
   const all = load(stateRoot);
-  if (all.filter((w) => w.session === input.session).length >= MAX_PENDING_WAKEUPS) {
+  const mine = all.filter((w) => w.session === input.session);
+  if (mine.length >= MAX_PENDING_WAKEUPS) {
+    // List what's pending: "unwake one" is only actionable if the model HAS the ids (they were returned
+    // when set, but that may be buried far back in the conversation).
+    const pending = mine.map((w) => `${w.id}${w.cron ? ` (recurring "${w.cron}")` : ""} at ${w.fireAt}`).join("; ");
```
kid7st: The pending list omits each wake-up's **prompt**, so "unwake one" is still not fully actionable. This change's own thesis is "only actionable if the model HAS the ids" — but by the same logic, *choosing which one* to unwake requires knowing what each does, and the prompt is exactly as "buried far back in the conversation" as the ids were. With only `id + fireAt (+ cron)` the model can only pick blindly by time or recency. Suggestion: append a truncated prompt per entry (e.g. first ~60 chars) so the model can choose by meaning.

kid7st (commented):
Thesis: envelope the woken turn (id + origin), make guardrail messages actionable, and have `info` load+validate schedules instead of filename-listing. Semantics are coherent; two findings.

**Not anchorable to the diff — missing tests for `info`'s new behavior:** the PR's third feature ships untested. `--json` `schedules` changes shape (`string[]` → `{name, cron, tz, next}[]`), `scheduleFailures`/`selfSchedule` are new keys, and the headline claim — "a broken schedule file is reported at info time, not first at `dev`" — is unverified. `test/cli.test.ts` has the exact isomorphic case for tools ("info degrades when a tool can't load — reports it, still shows the surface, exits 0"); schedules take the same G2 isolation path with no counterpart. Add one: a workspace with a bad-cron schedule file → assert `info --json` exits 0, `scheduleFailures` contains it, and a good schedule carries `next`.

src/schedule/wakeups.ts
kid7st: Fixed in 27a6861: each pending entry now carries a ~60-char prompt preview (id + fireAt + recurring cron + prompt) so the model can choose by meaning, not just by time. Also added the missing info test from the review body — G2 parity with the tools case: bad-cron schedule file → info exits 0, scheduleFailures surfaces it (json + stderr), the good schedule carries its next instant, selfSchedule line asserted.

PR #198, opened 2026-07-09
Fix agent-dir init gitignore layout

Sukitly:
## Summary
- Keep run-root ignores focused on `.env` and `.fastagent` for `--agent-dir` workspaces
- Add a kit-local `.gitignore` template so `<agentDir>/node_modules` is ignored inside the agent kit
- Document and test the layout-specific ignore behavior

## Tests
- npm run typecheck
- npm test


src/scaffold/init.ts
```diff
@@ -242,29 +243,42 @@ export async function scaffoldWorkspace(dir: string, options: ScaffoldOptions =
...
       if (still.length > 0) {
         warnings.push(
           `your .gitignore/.fastagentignore does not exclude ${still.map((s) => `"${s}"`).join(", ")} — a deploy that copies the directory may ship secrets/state`,
         );
       }
     }
+    if (!minimal && options.agentDir) {
+      const kitRel = options.agentDir.replace(/^\.[/\\]/, "").replace(/\\/g, "/");
+      const kitIgnore = await loadRootIgnore(join(dir, kitRel));
+      if (!coveredBy(kitIgnore, "node_modules")) {
```
kid7st, line 280: Kit coverage check is asymmetric with the root check — the "appends only what's missing" invariant doesn't hold on the kit side. The root path dedups against the effective matcher; here only the kit-local ignore is consulted (`loadRootIgnore(join(dir, kitRel))`), never the root. A root-level unanchored `node_modules/` (which almost every host repo has) matches at any depth in git, so an adopted kit `.gitignore` gets an effectively redundant `node_modules/` appended, and `patched` reports an update that wasn't needed. If this is a deliberate jurisdiction stance ("the kit self-contains its dependency ignore regardless of root"), say so in the comment; otherwise: `coveredBy(rootIgnore, `${kit}/node_modules`) || coveredBy(kitIgnore, "node_modules")`.

src/scaffold/init.ts
```diff
@@ -242,29 +243,42 @@ export async function scaffoldWorkspace(dir: string, options: ScaffoldOptions =
...
-      const still = need.filter((p) => !stillCovered(p));
+      const still = rootNeed.filter((e) => !coveredBy(after, e.path)).map((e) => e.path);
       if (still.length > 0) {
         warnings.push(
           `your .gitignore/.fastagentignore does not exclude ${still.map((s) => `"${s}"`).join(", ")} — a deploy that copies the directory may ship secrets/state`,
         );
       }
     }
+    if (!minimal && options.agentDir) {
+      const kitRel = options.agentDir.replace(/^\.[/\\]/, "").replace(/\\/g, "/");
```
kid7st: The manual `kitRel` normalization is dead code and inconsistent with the rest of the function. `kit = options.agentDir ?? "."` already exists and every other path (occupants check, `files`, `keptPkg`) goes straight through `join(kit, ...)` — `join` eats the `./` prefix and normalizes separators per platform. Both regexes feed immediately into `join`, so their net effect is zero, while implying `join` isn't enough. Drop `kitRel` and use `kit` directly.

kid7st (commented):
Thesis: ignore-file ownership now follows layout jurisdiction — root `.gitignore` owns run-root `.env`/`.fastagent`, `node_modules/` lives where the kit lives (flat → root, agentDir → `<agentDir>/.gitignore`). The core semantics are clear.

One finding that doesn't anchor to a diff line:

**src/scaffold/init.ts:58 — `ScaffoldResult.patched` doc is stale.** It still reads "(a kept .gitignore missing the .env/.fastagent excludes)", but `patched` now also covers the flat-layout `node_modules/` append and the kit-side `<agentDir>/.gitignore` append. The consumer (cli.ts "missing fastagent excludes appended") was updated; the type doc wasn't. Reword to the current semantics, e.g. "kept ignore files appended with missing fastagent excludes (root: env/state; kit: node_modules)".

src/scaffold/init.ts
kid7st: Addressed in the review commit: the kit-only check is DELIBERATE and now documented in place — the kit is the portable agent directory, so its dependency ignore must travel with it (a fresh kit .gitignore carries node_modules/ unconditionally for the same reason; skipping the adopted-kit append when the root happens to cover it would make adopted kits less self-contained than fresh ones). The dead kitRel normalization is dropped (plain `kit` + `join`, like every other path in the function), and ScaffoldResult.patched's docstring now states the current semantics (root: env/state + flat node_modules; kit: node_modules). Also rebased onto main (branch was behind). — handled by the maintainer to move the PR along; shout if the jurisdiction call reads wrong.

src/scaffold/init.ts
kid7st: Addressed in the review commit: the kit-only check is DELIBERATE and now documented in place — the kit is the portable agent directory, so its dependency ignore must travel with it (a fresh kit .gitignore carries node_modules/ unconditionally for the same reason; skipping the adopted-kit append when the root happens to cover it would make adopted kits less self-contained than fresh ones). The dead kitRel normalization is dropped (plain `kit` + `join`, like every other path in the function), and ScaffoldResult.patched's docstring now states the current semantics (root: env/state + flat node_modules; kit: node_modules). Also rebased onto main (branch was behind). — handled by the maintainer to move the PR along; shout if the jurisdiction call reads wrong.

PR #200, opened 2026-07-09
Use project auth for first-run model picker

Sukitly:
## Summary
- Resolve the first-run model picker against the workspace project auth file
- Keep --auth-path / FASTAGENT_AUTH_PATH precedence intact
- Add coverage for the effective workspace auth path fallback

## Tests
- npx biome check src/cli.ts src/engines/pi/config.ts test/config.test.ts
- npm run typecheck
- npm test


kid7st (approved):
LGTM. Verified the picker now resolves the same auth file as the opener (`workspace.ts` anchors to `resolveStateRoot(dir)` identically), the flag > env > project-default precedence is intact, and `loadDotEnv` precedes the picker at every call site so a `.env` `FASTAGENT_AUTH_PATH` is still honored. Typecheck + full test suite green locally.

Optional nit (non-blocking): `cli.ts:1299` (`maybeSeedAuth` in `start`) still inlines `authPathOverride ?? defaultAuthPath(resolveStateRoot(dir))` — exactly what `resolveAuthPath` now expresses; fine to reuse it there in a follow-up.

PR #229, opened 2026-07-14
fix(deploy): honor proxy for webhook registration

Sukitly:
## Summary

- initialize the Node proxy fetch dispatcher immediately after loading the workspace `.env` in the shared deploy command, covering both Fly and Railway post-deploy channel registration
- include the final transient Telegram transport error when `setWebhook` retries are exhausted
- add a real CLI subprocess regression test using an `.env`-only local proxy, plus coverage for the final retry diagnostic

Node's native `fetch` does not honor `HTTP_PROXY` or `HTTPS_PROXY`. Without this initialization, `fastagent deploy fly --run` and `fastagent deploy railway --run` could deploy successfully but fail to register webhooks on networks that require a proxy.

## Validation

- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test` (683 tests)

## Checklist

- [x] Docs were updated when behavior or public APIs changed (not applicable: no public API or documented contract changed)

### DIFF
--- a/src/channels/telegram/register-webhook.ts
+++ b/src/channels/telegram/register-webhook.ts
@@ -44,6 +44,7 @@ export async function registerTelegramWebhook(
 
   // Reachable → register. A short retry backstops Telegram's resolver lagging /health by a moment; only
   // network-transient errors retry (a permanent "bad webhook" config error is reported, not retried).
+  let lastTransientError = "unknown transport error";
   for (let attempt = 0; attempt < 3; attempt++) {
     if (attempt > 0) await sleep(opts.retryMs ?? 2000);
     try {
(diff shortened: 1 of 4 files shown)


kid7st (approved):
Correct and minimal: `runDeploy` was the one network entry point in cli.ts missing `installProxyFetch()`, and this uses the existing standard solution (src/proxy.ts) at the right spot — after `loadDotEnv` so a workspace-.env proxy is honored, before `loadConfig` so config-time fetches are covered too. The last-transient-error in the retry-exhausted warning is a good fail-visibly improvement, and the .invalid-host subprocess test proves the install ordering end to end. LGTM.

PR #230, opened 2026-07-14
feat(deploy): add local Docker Compose target

Sukitly:
## Summary

- add `fastagent deploy docker` as a first-class local deployment target using the existing portable Dockerfile generator
- generate a user-owned `fastagent.compose.yml` with a loopback port, persistent `/data` state volume, exact secret names, and safe keep/`--force` semantics
- add `--run` support that validates Docker/Compose, carries API-key or OAuth credentials without putting values in argv, reconciles the service, and verifies readiness
- add optional `--tunnel` generation for a pinned Cloudflare Quick Tunnel sidecar; `--run` reads its ephemeral URL and reuses the existing Telegram/Feishu/Lark webhook registration flow
- keep the experimental `agentDir + --run` gate unchanged for Docker, Fly, and Railway

## Details

The Docker target reuses `containerArtifacts()` from the Fly/Railway deployment path; only the Compose topology and local CLI sequence are target-specific. Existing Dockerfiles and Compose files remain authoritative unless `--force` is explicit.

The generated tunnel service stays separate from the FastAgent image. It bypasses Docker Desktop's injected proxy for the internal `http://agent:<port>` origin while host-side webhook registration continues to honor `HTTPS_PROXY`.

## Validation

- `npm run lint`
- `npm run typecheck`
- `npm test` — 65 files, 697 tests
- `npm run build`
- Docker Desktop smoke test: local health, OAuth auth seeding, Quick Tunnel creation, public health, and Telegram webhook registration


src/deploy/docker/plan.ts
```diff
@@ -0,0 +1,218 @@
...
+export const CLOUDFLARED_IMAGE = "cloudflare/cloudflared:2026.7.1";
+
+/** Whether a Compose file still opts into generated-drift warnings (drop the marker after taking ownership). */
+export function isGeneratedCompose(content: string): boolean {
+  return content.startsWith(GENERATED_COMPOSE_MARKER);
+}
+
+/** Whether an authoritative Compose file already carries the generated tunnel service contract. */
+export function composeHasTunnelService(content: string): boolean {
+  return /^\s{2}tunnel:\s*(?:#.*)?$/m.test(content);
```
kid7st: nit: this requires exactly 2-space indentation. A user who re-indents their kept Compose file (4 spaces, tabs) silently falls out of the preserve-tunnel branch in `runDeploy`, producing a spurious drift warning and a no-tunnel runbook — while the `--run` path (which uses `docker compose config --services`) still sees the tunnel. `/^[ \t]+tunnel:\s*(?:#.*)?$/m` would keep the two detections consistent for the common re-indent case.

src/cli.ts
```diff
@@ -814,6 +831,44 @@ async function runDeploy(): Promise<void> {
...
+    const projectName = toDockerProjectName(basename(target));
+    const dockerPlan = (tunnel: boolean) =>
+      planDockerDeploy({ projectName, port, modelAuth, channels, tunnel, extraSecrets, ...container });
+    let plan = dockerPlan(!!values.tunnel);
+    // Absence of --tunnel never means "remove it": an existing Compose file is authoritative. Preserve
+    // its generated tunnel shape in the comparison/runbook so a later plain `--run` starts + registers
+    // the topology the operator already chose. `--force` is the explicit reset to a no-tunnel default.
+    const composeFile = join(target, plan.composePath);
+    if (!values.tunnel && !values.force && (await exists(composeFile))) {
+      if (composeHasTunnelService(await readFile(composeFile, "utf8"))) plan = dockerPlan(true);
```
kid7st: `--tunnel` without `--run` against a kept no-tunnel Compose file: the file is correctly kept, but the printed runbook is the tunnel-plan runbook — it describes a `tunnel` service that isn't on disk (`compose logs -f tunnel`, auto-registration). The `--run` path gates on this; the generation-only path prints instructions that won't work. Worth mirroring the gate's hint (edit / delete+regenerate / `--force`) when the kept file lacks the service.

src/deploy/docker/run.ts
```diff
@@ -0,0 +1,183 @@
...
+  const compose = ["compose", "-f", plan.composeFile];
+  const env = plan.secrets;
+
+  // CLI/plugin gate first: unlike a daemon error, spawn ENOENT becomes 127 at the shared runner seam.
+  const version = await docker(["compose", "version"], { capture: true });
+  if (version.code === 127) {
+    return gate("Docker CLI not found — install Docker Engine/Desktop, then re-run");
+  }
+  if (version.code !== 0) {
+    return gate("Docker Compose plugin is unavailable — install/enable `docker compose`, then re-run");
```
kid7st, line 92: nit: the generated Compose file uses top-level `name:` (Compose >= v2.3.3). An older Compose plugin passes this gate, then fails at `config --services` with "invalid Compose file — fix fastagent.compose.yml", which points the user at a file that is fine. Either mention the minimum Compose version in this gate message or in the runbook prereq line — no need to parse the version.

kid7st (commented):
Solid target addition — reuses the container/secrets/webhook seams exactly as the deploy/ layout intends, gates precede the first side effect, and the stale-tunnel-log `rm -s -f` detail is right. Three nits inline, none blocking.

kid7st (approved):
All three review threads addressed in 2da4a6a — the kept-topology fix went beyond the suggestion (plan now shaped from the on-disk Compose file, so runbook/drift comparison always match reality). CI green across Node 22/24/26 + Bun + packed smoke. Ready to merge.

PR #231, opened 2026-07-14
fix(channels): report terminal webhook-registration failures at ERROR

kid7st:
## Summary

Two layers of the same fail-visibly gap in webhook registration, found while reviewing #229:

**L1 — log level.** A registration run that ends with the webhook NOT registered (exhausted transient retries, /health never up) left the operator in the same state as a permanent config error but was reported at `warn` while the permanent path used `error`. Terminal failures now log at ERROR in both the telegram and feishu registrars.

**L2 — exit code.** `deploy --run`'s contract is the gate model (every step that leaves the operator with work stops with one actionable line and a non-zero exit), but step 7 (post-deploy webhook registration) was outside it: a terminal registration failure returned `{ ok: true }` → exit 0 — a coding agent reports "deployed" while the bot cannot receive messages.

Design (registrars report facts, deploy owns policy):

- Registrars return `RegistrationOutcome = "registered" | "manual" | "failed"` (`src/channels/registration.ts`) — a fact at their own layer. The tunnel (long-running dev) ignores it.
- The shared host-neutral kernel `src/deploy/registration-gate.ts` owns the policy: `"failed"` gates (steps 1-6 are idempotent, so a re-run just retries registration); `"manual"` does not gate — re-running can never change it (an unclearable gate would spin a coding agent forever: the Lark cloud-lag 404 has no API path on that cloud) — but every remaining manual step (including github's repo webhook and unwired feishu/lark) is re-surfaced after all registrar output so it cannot be buried.
- The gate message composes with cli.ts's `deploy stopped:` prefix (leads with the failure; notes the app itself deployed).

Note: one-line overlap with #229 on the telegram exhausted-retries line (it enriches the message, this changes the level). Whichever lands second rebases; the merged form is `log.error(...last error: ...)`.

## Validation

- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test` (686 tests)
- [x] Tests: ERROR level on terminal registrar paths; three-state outcomes per path (incl. "manual" for missing creds / cloud-lag 404); runners gate on "failed" after attempting all channels; mixed manual+failed composition; "manual" surfaces without gating

## Checklist

- [x] Docs were updated when behavior or public APIs changed (AGENTS.md repo map: registration.ts + registration-gate.ts)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -46,6 +46,7 @@ src/
 │   ├── turn-store.ts        # SHARED: generic durable turn intent (L1) — record shape/validator/order injected per channel
 │   ├── state.ts             # SHARED: atomic state files under <stateRoot>/channels/<kind>/
 │   ├── wait-health.ts       # SHARED: readiness probe for the webhook registrars (both platforms verify the URL)
+│   ├── registration.ts      # SHARED: registrar outcome type (registered|manual|failed) — registrars report facts, deploy owns gate policy
 │   ├── github/              # github channel (+ scaffold/ bundle)
 │   ├── telegram/            # telegram channel — see docs/design/core.md §9.2
 │   │   ├── telegram.ts      # Telegram wiring: ingress + per-turn lifecycle + composition (pure parsing → parse.ts, run one turn → invoke-turn.ts)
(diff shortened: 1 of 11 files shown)


Sukitly (approved):
LGTM

PR #233, opened 2026-07-14
fix(feishu): reuse queued reply cards in place

Sukitly:
## Summary

- Mount Feishu/Lark queue feedback as the turn's streaming card, reply-quoted to the exact source message in both group and p2p chats.
- Reuse that card for the thinking preview and final Markdown answer instead of recalling a text notice and posting a second message.
- Show queue feedback immediately by default while retaining `queueNoticeDelayMs` for explicit delayed behavior.
- Keep text fallback, deferred-turn, and poison-turn terminal updates on the same mounted preview when available.

## Validation

- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test`

## Checklist


kid7st (approved):
LGTM

PR #234, opened 2026-07-14
fix(pi): restore the session's active-tool set when reopening (stateless invoke)

kid7st:
## Summary

> Scope note: this PR originally carried the pi 0.80.7 bump + the no-date prompt-cache fix. Those changes reached `main` as part of #235's merge (its commit message does not mention them — a working-tree mixup between two concurrent sessions; content verified correct on main). After rebasing, this PR is only what main does not have yet.

**Restore the session's recorded active-tool set on reopen.** pi's harness writes active-tool changes to the session (`setActiveTools` → `active_tools_change`) but its constructor never reads them back — pi's own TUI harness is long-lived and keeps the set in memory. fastagent builds a FRESH harness per invoke, so a session's recorded active set silently reset to "all tools" on every turn.

`piHarnessFactory` now restores it from `session.buildContext()`:

- recorded names are **filtered to the mounted tools** — the constructor throws on unknown names, so a tool removed from the workspace would otherwise brick every future invoke of that session
- an intact **empty set is restored as-is** (pi allows `setActiveTools([])`; a deliberate recorded state, not a degradation)
- a non-empty set that **filters down to empty falls back to the default** — the recorded intent cannot be honored, and honoring its empty shadow would be a silent capability loss
- filter degradations **warn once per session+missing set** (fresh harness per invoke × weeks-long channel sessions would otherwise repeat the same warn every turn)

Nothing in fastagent writes `active_tools_change` yet — this makes reopen faithful to whatever the session records, and is the load-bearing wiring for deferred/dynamic tool loading (pi 0.80.7) on the serving path.

Docs: `docs/design/core.md` §3 — the "continuity comes from PiSessionStore" claim now names the active-tool restore.

## Validation

- [x] `npm run lint` / `npm run typecheck` / `npm test`
- [x] Tests: write→reopen restores the set; a removed tool is dropped instead of bricking the session; unit three-state semantics (null → default, intact [] faithful, filter-to-empty → default)

## Checklist

### DIFF
--- a/docs/design/core.md
+++ b/docs/design/core.md
@@ -94,7 +94,11 @@ sessions, and state paths. `dev`, `start`, `invoke`, and `fire` share this assem
 parallel implementations.
 
 Each invocation builds a fresh harness for its session and discards it after the turn. Conversation
-continuity comes from `PiSessionStore`, not a resident harness.
+continuity comes from `PiSessionStore`, not a resident harness. Reopening is faithful to the whole
+record, not just the messages: pi's harness writes active-tool changes to the session but never reads
+them back (its own TUI harness is resident), so `piHarnessFactory` restores the recorded active-tool
+set itself — filtered to the mounted tools, since a recorded-but-removed tool would fail construction
+(`harness.ts` `restoreActiveToolNames`).
 
 ## 4. Event translation and terminal discipline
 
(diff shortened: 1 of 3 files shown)


Sukitly (approved):
LGTM

PR #237, opened 2026-07-14
feat(config): thinkingLevel — the serving-side counterpart of pi's thinking selector

kid7st:
## Summary

Stacked on #234 (needs the 0.80.7 chat-parity test from it; retarget to `main` after #234 merges).

Authors tune reasoning effort in the pi TUI while vibing, but the served agent had no way to set it: `piHarnessFactory` never passed `thinkingLevel`, so every deployment ran at pi's default — a fidelity gap (what you iterate is not what you serve). `AgentHarnessOptions.thinkingLevel` has existed all along; this is a pure pass-through that was missing.

- `config.thinkingLevel`: `off | minimal | low | medium | high | xhigh | max` (pi's scale — includes Fable 5's new `xhigh`/`max` from 0.80.7). Validated at config load (fail visibly on a typo); levels a model doesn't support are clamped by pi per model. Unset = pi's default.
- Threads workspace → L2 `createPiAgentFromDefinition` / L1 `createPiAgent` → `piHarnessFactory` → `AgentHarness`.
- Surfaced in the `dev`/`start` banner (`model: … (thinking: high)`) and `fastagent info` (text + json).
- Scaffold config template + `docs/configuration.md` updated.

## Validation

- [x] `npm run lint` / `npm run typecheck` / `npm test` (688)
- [x] Tests: config validation (valid + typo + non-string), harness threading (`getThinkingLevel() === "high"`), alongside the existing maxRetries wiring guard

## Checklist

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -38,6 +38,7 @@ Supported keys:
 | Key | Meaning |
 |---|---|
 | `model` | Default model spec, in `provider/modelId` form. |
+| `thinkingLevel` | Reasoning effort for the model, on pi's scale: `off` \| `minimal` \| `low` \| `medium` \| `high` \| `xhigh` \| `max`. Unset = the engine default. Levels a model doesn't support are clamped by the engine. The serving-side counterpart of the thinking selector authors use in the pi TUI. |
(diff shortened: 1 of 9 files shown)


docs/configuration.md
```diff
@@ -38,6 +38,7 @@ Supported keys:
 | Key | Meaning |
 |---|---|
 | `model` | Default model spec, in `provider/modelId` form. |
+| `thinkingLevel` | Reasoning effort for the model, on pi's scale: `off` \| `minimal` \| `low` \| `medium` \| `high` \| `xhigh` \| `max`. Unset = the engine default. Levels a model doesn't support are clamped by the engine. The serving-side counterpart of the thinking selector authors use in the pi TUI. |
```
Sukitly: Which one is pi's default? Should we explicitly specify one as the default? If pi changes its default behavior, fastagent's program could be affected without our knowledge.

kid7st: Great catch — this uncovered a real fidelity gap, not just a docs gap. Facts: pi's TUI defaults to `medium` (`DEFAULT_THINKING_LEVEL`), but the bare harness (what fastagent serving uses) falls back to `off` — so an author who vibed at `medium` was silently served at `off`. Fixed in 11eda14: fastagent now pins its own default `medium` in `piHarnessFactory` (pi TUI parity), so an upstream default change in either place can no longer silently alter deployments. Docs + scaffold template state the default explicitly, and a test pins unset → `medium`.

PR #241, opened 2026-07-15
feat(pi): deferred tools — dynamic tool loading on the serving path

kid7st:
## Summary

For tool-heavy agents, every turn paid every tool schema and the model picked from all of them. This lands pi 0.80.7's dynamic tool loading on the serving path, with fastagent's own authoring surface (the load-bearing wiring — session active-set restore — landed in #234):

- **`defineTool({ deferred: true })`** registers a tool without activating it: its schema stays out of every request (and the model's sight) until discovered. Fresh sessions start with every non-deferred tool active; tool-sets without deferral behave exactly as before.
- **Built-in `search_tools` loader**, auto-mounted only when a deferred tool exists (an authored `search_tools` wins — the wake-pair rule; one marked `deferred` keeps its loader active with a warning). AND keyword matching (adding a word narrows) over the **whole** catalog — an already-active match says "call directly", never the capability-missing trap. Per-search **activation cap (5)** with a guaranteed escape (an exact tool name activates that one tool); every listing (miss / over-cap / active-note) is capped too — no answer pours the catalog back into the context deferral protects.
- **`ToolContext.tools`** (`active/registered/activate`) is the per-turn activation bridge (same ALS seam as `ctx.session`); `search_tools` is its first consumer. Activations are **serialized locally** (per-turn chain — not a bet on pi's setter timing); `addedToolNames` — the native deferred-loading load point — comes from each execute's OWN `activate()` calls (a snapshot diff would misattribute a parallel sibling's activation) and is stamped on a **copy** (an author's frozen result survives).
- **Session semantics: a record contributes activations, not a frozen snapshot.** Reopen rebuilds the active set as initial (current non-deferred tools) ∪ recorded activations — a tool added to the workspace later joins old sessions; the corollary (durable narrowing via `setActiveTools` is not honored) is documented in core.md as a constraint on future writers.
- **`chat` emulates deferral** (what you iterate is what you serve): fresh sessions narrow by subtracting deferred names (robust to pi mounting its own tools), the same loader activates through a session-lifetime bridge with the same serialization, prompt identical to serving. One documented divergence (pinned by test): chat activations don't survive `/new`/`/resume` — pi's chat session records none, so a resumed conversation re-discovers.
- Prompt stability: the tools list comes from the static mounted set + a discovery note; activation never rewrites the prompt. Reporting: `deferred:` line in dev/start banner + `fastagent info` (text + json), one report slot per name.

Docs: api-reference (Deferred tools incl. L1 verbatim-instructions contract), core.md §3/§5, AGENTS.md repo map.

## Validation

- [x] `npm run lint` / `npm run typecheck` / `npm test` (707)
- [x] E2E through invoke (faux model): activation → `addedToolNames` on the toolResult → next turn's fresh harness keeps it active; chat E2E drives the real loader through pi's tool surface, incl. /new re-narrowing
- [x] Failure paths: no-match, over-cap (zero activations) + exact-name escape, outside-turn degradation, parallel-batch stamp attribution, frozen result, deferred authored loader, later-added tool joins old sessions

## Checklist

### DIFF
--- a/src/engines/pi/chat.ts
+++ b/src/engines/pi/chat.ts
@@ -207,17 +207,15 @@ export async function buildChatRuntime(
       customTools: customToolDefs,
     });
     sessionRef.current = result.session;
-    // Deferral emulation: pi's TUI session starts with everything active — narrow it to the initial
-    // active set (non-deferred), like a fresh served session. Only when the session still has the
-    // all-active default: a /resume of a session where the loader already activated tools keeps its
-    // narrower recorded set (the corner where a resumed session had ALL tools deliberately active is
-    // re-deferred — chat-local, and search_tools re-activates in one call).
+    // Deferral emulation: pi's TUI session starts with everything active — narrow a FRESH session by
+    // SUBTRACTING the deferred names from whatever is active (robust to pi mounting tools of its own;
+    // an exact-set-equality gate would silently stop narrowing the day pi adds one). A resumed session
+    // (has messages) keeps its recorded set — the loader's earlier activations survive, like serving.
     const deferredNames = customTools.filter(isDeferredTool).map((t) => t.name);
-    if (deferredNames.length > 0) {
-      const allNames = [...defaultNames, ...customTools.map((t) => t.name)];
+    if (deferredNames.length > 0 && result.session.messages.length === 0) {
       const active = result.session.getActiveToolNames();
-      if (active.length === allNames.length && allNames.every((n) => active.includes(n))) {
-        result.session.setActiveToolsByName(allNames.filter((n) => !deferredNames.includes(n)));
+      if (deferredNames.some((n) => active.includes(n))) {
+        result.session.setActiveToolsByName(active.filter((n) => !deferredNames.includes(n)));
       }
     }
     return { ...result, services, diagnostics: services.diagnostics };
--- a/src/engines/pi/chat.ts
+++ b/src/engines/pi/chat.ts
@@ -41,7 +41,7 @@ import { createPiModels } from "./models.ts";
 import { canonicalPath, loadAgentDefinition } from "./definition.ts";
 import { isDeferredTool, loadTools, mergeDiscoveredTools } from "./tool.ts";
 import { withSearchTool } from "./search-tools.ts";
-import { type ToolActivation, turnContext } from "./tool-context.ts";
+import { type ToolActivation, additiveActivation, turnContext } from "./tool-context.ts";
 import { reportDefinitionWarnings, reportModuleLoadFailures, reportToolCollisions } from "./report.ts";
 
 export interface RunPiChatOptions {
@@ -69,9 +69,12 @@ export async function buildChatRuntime(
       active: () => session.getActiveToolNames(),
       registered: () => session.getAllTools().map((t) => ({ name: t.name, description: t.description ?? "" })),
       async activate(names) {
-        const registered = new Set(session.getAllTools().map((t) => t.name));
         const current = session.getActiveToolNames();
-        const added = [...new Set(names)].filter((n) => registered.has(n) && !current.includes(n));
+        const added = additiveActivation(
+          session.getAllTools().map((t) => t.name),
+          current,
+          names,
+        );
         if (added.length > 0) session.setActiveToolsByName([...current, ...added]);
         return added;
       },
--- a/src/engines/pi/chat.ts
+++ b/src/engines/pi/chat.ts
@@ -135,7 +135,10 @@ export async function buildChatRuntime(
       parameters: t.parameters,
       execute: (id: string, params: unknown, signal: AbortSignal | undefined) => {
         const bound = sessionRef.current;
-        if (!bound) return t.execute(id, params, signal);
+        // Unreachable by construction (createRuntime sets sessionRef before any turn can run a tool).
+        // Throw rather than silently run outside the turn context — that would disguise a broken
+        // session-lifecycle invariant as a normal out-of-turn call (fail visibly).
+        if (!bound) throw new Error("chat tool executed before its session was built (lifecycle invariant broken)");
         return turnContext.run(
           { session: bound.session.sessionId, tools: bound.activation },
           () => t.execute(id, params, signal) as Promise<unknown>,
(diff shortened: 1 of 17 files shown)


src/engines/pi/chat.ts
```diff
@@ -77,18 +112,38 @@ export async function buildChatRuntime(
...
+    // it). Each execute runs inside the turn context with the CURRENT session's activation bridge — the
+    // assembly is memoized across /new//resume/fork rebuilds while the session changes, so the bridge
+    // resolves through sessionRef at call time, exactly like the serving path resolves its harness.
     const customToolDefs = customTools.map((t) => ({
       name: t.name,
       label: t.name,
       description: t.description ?? "",
       parameters: t.parameters,
-      execute: (id: string, params: unknown, signal: AbortSignal | undefined) => t.execute(id, params, signal),
+      execute: (id: string, params: unknown, signal: AbortSignal | undefined) => {
```
Sukitly, line 139: Could we address the second attribution layer on the chat path? These definitions are passed to Pi as SDK `customTools`, and Pi's `wrapRegisteredTool` also derives `addedToolNames` from the session-wide active set before/after `execute`. With two parallel `search_tools` calls, both outer wrappers capture the same initial set; the first activates `lookup_weather`, while the second reports it as already active, but both results are stamped with `addedToolNames: ["lookup_weather"]`. That defeats the per-execute attribution this bridge is trying to guarantee and can anchor the native definition at the wrong result. I reproduced this with `Promise.all` against `buildChatRuntime`. We should either fix/disable the outer Pi diffing for SDK tools or serialize loader execution, and add a parallel chat regression test.

src/engines/pi/harness.ts
```diff
@@ -75,43 +76,45 @@ const PROVIDER_MAX_RETRIES = 2;
...
-    );
-    return undefined;
+  if (missing.length > 0) {
+    const emit = warnedRestores.has(`${sessionId}\u0000${missing.join(",")}`) ? log.debug : log.warn;
+    warnedRestores.add(`${sessionId}\u0000${missing.join(",")}`);
+    emit(`[fastagent] session ${sessionId}: dropping recorded active tool(s) no longer mounted: ${missing.join(", ")}`);
   }
-  emit(`[fastagent] session ${sessionId}: dropping recorded active tool(s) no longer mounted: ${missing.join(", ")}`);
-  return known;
+  return [...new Set([...initial, ...known])];
```
Sukitly, line 137: This record is a full active-set snapshot, not an activation delta: `setActiveTools([...current, ...added])` persists all currently active non-deferred names too. Reinterpreting every recorded name as an activation means a tool that was ordinary when the snapshot was written remains active if it is later changed to `deferred`, even though the session never discovered it; this also discards the previous durable-narrowing semantics. Can we persist the added names separately (for example, in a dedicated session record) rather than unioning a snapshot with today's initial set?

src/engines/pi/tool.ts
```diff
@@ -14,26 +14,54 @@ import { join } from "node:path";
...
+   *  exists — activates it mid-turn. For tool-heavy agents: fewer schemas per turn, and on providers
+   *  with native deferred loading the activation preserves the prompt-cache prefix. The trade-off:
+   *  discovery rides entirely on this description — write it for the search. Default: false.
+   */
+  deferred?: boolean;
   execute: (input: z.infer<I>, ctx: ToolContext) => unknown | Promise<unknown>;
 }
 
+/** Read the {@link DefineToolOptions.deferred} marker off a mounted tool (extra property on the
+ *  AgentTool object — pi ignores it; config.tools authors can set it on a raw tool too). */
```
Sukitly: The claim that raw `config.tools` can set this marker is not type-safe today: `FastagentConfig.tools` is `AgentTool[]`, and upstream `AgentTool` has no `deferred` property, so a TypeScript `defineConfig({ tools: [{ ..., deferred: true }] })` fails excess-property checking. Please introduce and export a fastagent tool type such as `AgentTool & { deferred?: boolean }` and use it in `FastagentConfig` and the relevant L1 options, or narrow this contract to `defineTool`-created tools. Also, `ToolContext.tools` publicly exposes `ToolActivation`, but that type is not re-exported from `src/pi.ts`, which makes typed custom-loader helpers unnecessarily difficult.

src/engines/pi/chat.ts
kid7st: Confirmed and fixed — thanks for the repro. Root cause verified: pi wraps SDK customTools in wrapRegisteredTools' before/after diff, and the snapshots are taken before any inner serialization can help, so serializing our execute wasn't enough. The fix is pi's own mechanism: the builtin loader now declares `executionMode: "sequential"` (now a first-class `defineTool` option) — pi serializes any batch containing it, so the outer diff sees correct sequential snapshots. The chat adapter propagates the mode; an authored `search_tools` missing it gets the mode forced with a warning (a non-sequential loader would silently revive this bug); docs tell custom-loader authors to declare it. Regression test asserts the contract (marker on the registered tool) + single-stamp behavior under the serialization pi guarantees.

src/engines/pi/harness.ts
kid7st: Agreed — reinterpreting the snapshot was the weak point, and your flipped-to-deferred case is exactly the hole. Activations are now persisted as dedicated DELTA entries (`fastagent:tool-activation` custom session entries carrying only the names that call activated, written by the bridge through a harness→session seam); the resolve reads only those deltas and deliberately ignores pi's `active_tools_change` snapshots. A tool later flipped to deferred drops out of sessions that never discovered it (pinned by test), later-added tools still join old sessions. Durable narrowing remains unrepresentable — core.md records that as an explicit constraint on future writers.

src/engines/pi/tool.ts
kid7st: Both fixed: `FastagentTool` (`AgentTool & { deferred?: boolean }`) is exported and used by `config.tools` and the L1/L2 `tools` options, so a raw object literal with `deferred: true` passes excess-property checking; `ToolActivation` is re-exported from the pi surface. (`defineTool` also gained `executionMode` while closing the thread above.)

Sukitly (approved):
lgtm

PR #239, opened 2026-07-15
refactor(feishu): normalize webhook channel semantics

Sukitly:
## Summary

- introduce canonical Feishu/Lark raw-event and normalized message/resource models while preserving the existing public `route(event)` surface
- route verified webhook messages through one normalizer so turn assembly retains cloud, app, tenant, conversation, identity, mention, and message-scoped resource data
- add representative Feishu and Lark webhook fixtures that lock the shared wire model and normalization behavior
- align Feishu/Lark L1 persistence with Telegram by removing the channel-specific completed-message seen ring; `message_id` remains the delivery identity and `seq` preserves recovery order
- keep the implementation webhook-only and document the shared at-least-once duplicate-delivery limitation

## Validation

- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test`
- [x] Added or updated the smallest relevant tests

## Checklist

- [x] Public-facing text is in English
- [x] Errors fail visibly; no silent fallbacks or swallowed exceptions
- [x] No secrets, local paths, or machine-specific state were committed
- [x] No process-only notes were added (`*_PLAN.md`, `HANDOFF.md`, session notes, etc.)
- [x] Docs were updated when behavior or public APIs changed

PR #242, opened 2026-07-16
feat(feishu): add managed threaded conversations

Sukitly:
## Summary

- normalize Feishu/Lark webhook messages and align unfinished-turn persistence with the shared channel lifecycle
- make top-level p2p messages and summoned group messages independent root sessions backed by platform threads, with explicit continuous-mode opt-outs
- persist Agent-managed group roots and treat every later user message in those roots as a normal required turn, including queue feedback, streaming delivery, and user-visible failures
- document the `im:message.group_msg` requirement and cover root identity, restart recovery, routing boundaries, attachments, and deterministic crypto failure behavior

This supersedes #239: the normalization and turn-store groundwork is included here because stable root identity and restart-safe managed-thread routing depend on it.

## Validation

- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test`

## Checklist


kid7st:
Reviewed the full branch (8 commits). The direction is right — root-`message_id` session identity, pre-ACK persistence discipline (`owned-threads.json` / `buffers.json`, remove-then-commit on `completed`), the primary/buffered attachment tiers, and the test coverage are all solid. Four findings, in priority order:

**1. Removing `seen.ts` is not "alignment with Telegram" — it reopens a platform-specific failure mode.** The deleted code's own comment cites the reason it existed: the platform documents duplicate pushes *after a successful 200* and recommends deduping on `message_id`. Telegram's delivery protocol does not redeliver after an ACK, so the two channels were never symmetric here. Without the ring:

- a duplicate push while the turn is pending lands on the same session's FIFO chain and runs the turn twice → double reply;
- the pre-ACK crash window now double-runs too: recovery replays **and** the never-ACKed platform redelivers — exactly the case the turn-store header calls out with "unless the channel dedups accepted ids, the turn can run twice" (Feishu *was* that dedup).

The ring was ~48 lines of cheap insurance against a documented behavior. Please restore it (or an equivalent accepted-id check), or make the case for deletion on Feishu's actual delivery semantics rather than consistency with Telegram.

**2. The default flip is a breaking behavior change, on a partially unverified path.** Existing deployments upgrade into "every top-level DM opens a platform thread + a fresh session". The opt-outs and docs are complete, but (a) per-message DM threads as the *default* UX deserves an explicit product sign-off, and (b) the docs themselves note Lark thread creation has not been smoke-tested on a real tenant — shipping an unverified path as the default is risky. Consider defaulting Lark to `continuous` until it's verified, or verifying before merge.

**3. Unbounded growth (acknowledged).** `owned-threads.json`, `buffers.json`, and one pi session per root all grow without TTL/GC — already listed in Limits, fine for now, but an active group will grow these quickly; worth a follow-up issue.

**4. Minor: concurrent double-fold of a shared buffer key.** In threaded group mode, two concurrent roots in the same main chat each peek the same buffer snapshot, so the same background discussion folds into both turns (first `completed` commits). Harmless semantically, just duplicated context — a `ponytail:`-style note at the peek site would do.

1 is the one I'd consider blocking; 2 needs an explicit decision on the default.

Sukitly:
Keep `threaded` as default behavior, which is better UX in lark/feishu

kid7st (approved):
LGTM

kid7st (commented):
Review thesis: 本次改动把 Feishu/Lark 重塑为 per-root threaded 会话 + 平台 thread, 由 owned-threads.json / buffers.json / model+normalize 三个新概念支撑, 核心清晰. 4 条 findings 全部 inline.

src/channels/feishu/feishu.ts
```diff
@@ -437,42 +491,122 @@ export function buildFeishuChannel(
...
+          .filter((resource) => resource.kind === "file" || resource.kind === "audio" || resource.kind === "video")
+          .map((resource) => ({ msg: resource.messageId, key: resource.key, name: resource.name }));
         const baseText = r.text ?? cloudEnvelope(event, kind);
-        if (baseText.trim() !== "" || content.imageKeys.length > 0 || content.fileRefs.length > 0) {
+        if (baseText.trim() !== "" || images.length > 0 || files.length > 0) {
+          // Persist ownership before ACK. The platform thread does not exist until the first reply lands,
+          // but a failed reply has no continuation to misroute; pre-ACK ownership closes the opposite,
+          // worse window (thread created, process dies, then its unmentioned continuation is forgotten).
+          if (threadedGroup && m.thread_id === undefined && sameTarget && replyInThread === true) {
+            ownedThreads.add(m.chat_id, m.message_id);
```
kid7st, line 597: `ownedThreads.add` 不 gate 在 `route === undefined`, 写而不读: ownership 记录的语义是 "这个 root 的 bare continuation 会被 admit", 但唯一消费点 (line 513 的 `managedThread` admission) 要求 `route === undefined`. 自定义 route 的部署仍为每个 top-level group summon 写一条永不被读的记录 — 且这是 pre-ACK 同步持久化, 写失败 → webhook 500 → 平台 redelivery: 为一个对该部署禁用的功能承担 IO 失败面 + `owned-threads.json` 无限增长. 建议: add 的 gate 加 `route === undefined`, 与消费端对齐.

src/channels/feishu/model.ts
```diff
@@ -0,0 +1,128 @@
+/**
+ * Canonical Feishu/Lark protocol and normalized-message models.
+ *
+ * Raw wire types stay at the webhook/policy boundary: authors' existing `route(event)` callbacks keep
+ * receiving the platform event unchanged. The turn engine consumes {@link NormalizedFeishuMessage}
+ * instead, so transport envelopes, JSON-string message bodies, and platform field naming do not leak
```
kid7st: 模块头 claim 与实现不符: "The turn engine consumes NormalizedFeishuMessage instead, so … platform field naming do not leak further into persistence, attachment resolution…" — 实际 feishu.ts 的 session 推导、routing、StoredFeishuTurn 持久化全部直接消费 raw `m.chat_id/thread_id/root_id/parent_id` 与 `event.sender.sender_type`; normalized 只被用于 content/resources/bufferKey. 同时 `normalized.source/delivery/sender`、`NormalizedFeishuMention.isBot/openId/name`、`FeishuResourceKind` 的 `"sticker"` 在 src 内零消费, 只有测试断言它们 — 为 "future ingress" 预铺的 dead surface. 建议二选一: 完成迁移 (wiring 消费 normalized.conversation/sender), 或把 claim 收窄到 "content decoding + message-scoped resource locators" 并删掉无消费者的字段.

src/channels/feishu/invoke-turn.ts
```diff
@@ -56,38 +58,92 @@ interface ResolvedInputs {
...
+    ...backgroundFiles.map(({ file, ref }) => ({
+      ...file,
+      name: `${file.name} (from ${ref.from}, msg ${ref.messageId}, earlier discussion)`,
+    })),
+  ];
+  const manifest = allFiles.length
+    ? `\n\n[attached files — read them with your tools:\n${allFiles.map((file) => `- ${file.name} (${file.size} bytes) → ${file.path}`).join("\n")}\n]`
     : "";
-  return { images: imageRefs.length ? imageRefs : undefined, promptSuffix: `${referentBlock}${manifest}` };
+  const allImages = [...imageRefs, ...backgroundImages];
```
kid7st: buffered image 混入 vision 后 primary/background 分层丢失: 输入分层是本文件的核心语义 (PRIMARY fail-fast / BUFFERED degrade), files 侧执行到了 prompt — manifest 里每个 background file 标注 `(from X, msg Y, earlier discussion)`; images 侧却直接 `[...imageRefs, ...backgroundImages]` 追加, prompt 里只有加载失败时才出现 note. 模型无法区分哪张图是用户当前指着的、哪张是背景, 也无法把 vision 图对应回 discussion 文本里的 `[image]` 行. 建议: promptSuffix 里声明 background image 的数量与来源, 对齐 file manifest 的标注.

src/scaffold/add-channel.ts
```diff
@@ -90,7 +90,7 @@ const CHANNEL_SCAFFOLDS: Record<ChannelKind, ChannelScaffold> = {
       },
     ],
     steps: [
-      "PUBLISH the app version on the page the CLI opened — the switch to webhook mode takes effect on publish (one click, once ever; no API for it)",
+      "add the sensitive im:message.group_msg permission (tenant-admin approval) and PUBLISH the app version — this lets managed group threads receive unmentioned continuations; without it, only @mentions arrive",
```
kid7st: 必选步骤与可选步骤捆绑, publish 的理由被删: publish 是 webhook 订阅模式生效的必要前提 (对所有 Feishu 部署), sensitive `im:message.group_msg` 是 group buffering 的可选前提 (需 tenant-admin 审批). 新 step 把二者写成一句 "add the sensitive permission and PUBLISH", 读起来像 scope 是必装项; 原文 "the switch to webhook mode takes effect on publish (one click, once ever; no API for it)" 这一关键理由消失. 建议拆回两条: publish 必选并保留原理由; scope 可选单列并注明它启用什么.

src/channels/feishu/feishu.ts
Sukitly: Fixed in e089e49. `ownedThreads.add` now requires `route === undefined`, matching the only managed-thread admission path. Added a regression test proving an accepted custom-routed group turn does not create `owned-threads.json`.

src/scaffold/add-channel.ts
Sukitly: Fixed in e089e49. Feishu guidance now separates the optional `im:message.group_msg` step from the mandatory `PUBLISH` step and restores the webhook-mode reason. Lark guidance separates the same optional scope from its required setup as well; a test locks the distinction.

src/channels/feishu/model.ts
Sukitly: Fixed in e089e49 using the narrow option. Normalization now owns only the currently consumed conversation-place fields, decoded text, mention presence, and message-scoped resource locators. Removed the unused source/delivery/sender projection, per-mention fields, resource metadata/kinds with no consumer, and normalization options; narrowed the module/design claims and tests accordingly.

src/channels/feishu/invoke-turn.ts
Sukitly: Fixed in e089e49. Buffered image resolution now retains each source ref and appends a prompt manifest after primary images, naming the vision-image ordinal, sender, and message id. The regression test covers one primary image plus one buffered image and locks ordering/attribution.

PR #249, opened 2026-07-17
fix(pi): restore compatibility with pi 0.80.10 by owning the auth file lock

Sukitly:
Every fresh install is currently broken: `npm i @fastagent-sh/fastagent` resolves the `^0.80.7` pi range to 0.80.10 and fails at import time with `SyntaxError: The requested module '@earendil-works/pi-coding-agent' does not provide an export named 'FileAuthStorageBackend'`. The same failure gates the "Packed package smoke" CI job on every open PR (first seen on #248, a docs-only change).

## Root cause

pi 0.80.8 shipped a documented Breaking Change in the ModelRuntime refactor: `AuthStorage` and its storage backends are no longer exported, and the stated migration path for SDK consumers is a custom pi-ai `CredentialStore`. fastagent already implements exactly that; the only borrowed piece was the backend's cross-process file lock. Two smaller changes ride the same range: pi-ai's `CredentialStore` gained a required `list()` member, and `AuthLoginCallbacks` was renamed to `AuthInteraction` (identical members).

## Changes

- **`src/engines/pi/auth.ts`**: vendor the locked read-modify-write on `proper-lockfile` directly (the same library and parameters pi's backend used: exponential backoff, 30s staleness, compromise detection, 0700 dir / 0600 file modes). The shared tolerant reader now backs both `read()` and the new `list()` implementation, which returns `{ providerId, type }` metadata only and filters foreign entries with the same validation as `read`.
- **`src/engines/pi/login.ts` + `test/login.test.ts`**: `AuthLoginCallbacks` renamed to `AuthInteraction` (verified member-identical in pi's source).
- **`package.json`**: pi packages move to `^0.80.10` (the code now uses the 0.80.10 surface); `proper-lockfile` becomes a direct dependency (it was already in the tree transitively), plus its types as a dev dependency.
- **`test/auth.test.ts`**: two new cases locking `list()` behavior (metadata only, foreign-entry filtering, missing file silent, corrupt file warns).

## Trade-off

Owning ~50 lines of lock code removes the dependence on a pi internal that upstream is free to reshape, and matches upstream's own migration guidance. Pinning 0.80.7 instead would still require a release to take effect, would block Kimi K3 and later fixes, and would keep the fragile import.

## Verification

- `npm run lint`, `npm run typecheck`, `npm test`: 755 passed (753 + 2 new).
- Local replica of the CI packed-package smoke: `npm pack`, fresh install (resolves pi 0.80.10), all five entrypoint imports OK, `fastagent --version` OK.
- Full test suite was also run against 0.80.10 before the fix to confirm the failure surface: all 89 failures traced to the single removed export; no other runtime regressions.

## Follow-up (maintainer)

After merge, a 0.13.1 patch release is needed so published installs stop resolving into the broken combination. #248 needs a rebase to pick up the green smoke.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,15 +10,16 @@
       "license": "MIT",
       "dependencies": {
         "@clack/prompts": "^1.6.0",
-        "@earendil-works/pi-agent-core": "^0.80.7",
-        "@earendil-works/pi-ai": "^0.80.7",
-        "@earendil-works/pi-coding-agent": "^0.80.7",
+        "@earendil-works/pi-agent-core": "^0.80.10",
+        "@earendil-works/pi-ai": "^0.80.10",
+        "@earendil-works/pi-coding-agent": "^0.80.10",
         "@octokit/webhooks-methods": "^6.0.0",
         "@octokit/webhooks-types": "^7.6.1",
         "chokidar": "^5.0.0",
         "croner": "^10.0.1",
         "giget": "^3.3.0",
         "ignore": "^7.0.5",
+        "proper-lockfile": "^4.1.2",
         "undici": "^8.6.0",
         "zod": "^4.4.3"
       },
@@ -27,6 +28,7 @@
       },
       "devDependencies": {
         "@biomejs/biome": "^2.5.2",
+        "@types/proper-lockfile": "^4.1.4",
         "typescript": "^7.0.2",
         "vitest": "^4.1.9"
       },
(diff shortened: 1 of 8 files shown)


Sukitly:
All three findings addressed in 05be0e9, with one refinement on item 2's suppression rule.

**1. Init write outside the lock (accepted).** The init is now an exclusive create (`flag: "wx"`) that tolerates only `EEXIST`; a stale existence check can no longer clobber a file another process already initialized and filled. Coverage: a deterministic simulation of the stale check (`test/auth-init-race.test.ts`, existsSync mocked to report missing once while the file holds another provider's credential) plus real cross-process coverage (two spawned node processes first-writing different providers into the same absent path; both survive).

**2. Swallowed release errors (accepted, one refinement).** `withLockedAuthFile` now tracks the primary outcome: a mutation failure propagates and unlock noise cannot mask it; a failed release after a successful operation rejects instead of resolving with a leftover `auth.json.lock`; a compromise detected after the last in-band check also surfaces. Refinement versus the suggestion: when a primary error exists, the release error is suppressed regardless of its class, because classifying proper-lockfile's release errors is brittle and the primary must win either way. Coverage: `test/auth-lock.test.ts` (mocked proper-lockfile; success + failed release rejects, mutation failure stays primary).

**3. Structural validation (accepted).** A shared `decodeCreds` requires a non-null, non-array object root. The read path treats `[]`/`null`/scalar roots as corrupt (warn + degrade, same as unparsable text); the write path refuses them before the mutation callback runs, closing the array-root case where a "successful" modify persisted nothing (`JSON.stringify` drops string keys on arrays). Coverage: all three roots crossed with read/list/modify in `test/auth.test.ts`.

762 tests pass locally (lint + typecheck + full suite).

kid7st (approved):
LGTM

PR #250, opened 2026-07-17
feat: first-run model picker shows the full catalog and logs in inline

kid7st:
## What

`fastagent dev` / `start` / `invoke` with no model set used to show only models of providers that already had credentials — on a fresh machine that meant a warn + exit + "run `fastagent login`, then come back". The first-run funnel is now one command with no dead ends:

- **Full catalog, annotated.** The picker lists every registered model; ready providers lead, each annotated with its credential source (`ready — OPENAI_API_KEY`), the rest with their remedy (`login required`, or `API key required — set the provider's env var` when the provider has no interactive login flow). A broken stored credential is annotated, not silently dropped.
- **Inline login.** Picking a model whose provider needs auth runs `loginFlow` right there (with `runLogin`'s state-root leak guard), then continues to serving. Cancel is a decision, not a failure (`LoginCancelled`) — silent fall-through to the existing `missing model` error; `fastagent login` cancel now reports "login cancelled" instead of a failure.
- **Env-key-only providers keep the choice.** Model validity is independent of credentials: the pick is persisted, the remedy is named, the startup auth report warns until the env var is set.
- **Proof of life.** Serving ends with a copy-paste `try it:` curl; its body is `INVOKE_EXAMPLE_BODY`, exported by the http channel next to the shape check and pinned by a test so the hint cannot drift from the protocol.

## Deleted

- `configuredModelSpecs` (the filtered-menu primitive; sole caller was the picker) → replaced by `providerAuthStatuses` (3-state data: `ready`/`unconfigured`/`broken`, plus `interactiveLogin`).
- The `no authenticated provider — run fastagent login` dead-end branch.
- The picker's `select`-vs-`autocomplete` size switch: **the picker is now always `autocomplete`** (the full catalog is always large) — small-list users see the filterable UI too; deliberate.

## Also

- `providerOf(spec)` centralized in `config.ts` (spec-domain home); `reportAuth`'s inline `slice(0, indexOf("/"))` (mangles a spec without `/`) routed through it.
- Docs updated: cli.md, quickstart.md, configuration.md, troubleshooting.md.
- Named trade-off: the post-pick branch policy (keep-and-warn vs inline-login vs cancel) stays in `cli.ts` untested — it is three lines of side-effect sequencing around `loginFlow`; extracting a pure action-enum function was judged ceremony. The pure surfaces (`providerAuthStatuses`, `buildModelPickerOptions`, `INVOKE_EXAMPLE_BODY`) are unit-tested.

## Verification

`npm run lint && npm run typecheck && npm test` green (753 tests); `rv.sh local` run three times, all findings fixed or explicitly declined above.

## Rebuild (2026-07-17)

Rebuilt on top of the commander CLI restructure (#243) + pi 0.80.10 compat (#249): the cli.ts changes now live in `src/cli/shared.ts` (picker + `reportAuth` + `terminalLoginIO`, now shared with the login command), `src/cli/serve.ts` (`routesFor` provenance + try-it hint), and `src/cli/commands/login.ts` (`LoginCancelled` handling). Two policy refinements from the post-rebuild `rv` pass:

- A **failed** inline login now keeps the model choice (same policy as the env-key-only branch: model validity is independent of credentials) — the pick persists, the startup auth report names the remedy, and a later `fastagent login` fixes auth without re-picking. Cancel still discards, deliberately.
- An unknown-provider spec (unreachable when specs/statuses share one `Models`) hints neutral `auth required` instead of promising a login flow.

### DIFF
--- a/src/cli/shared.ts
+++ b/src/cli/shared.ts
@@ -105,7 +105,7 @@ export async function resolveFirstRunModel(
   const chosen = r as string;
   const provider = providerOf(chosen);
   const status = statuses.get(provider);
-  if (status && status.state !== "ready" && !status.interactiveLogin) {
+  if (status && status.state !== "ready" && status.login === "none") {
     // No login flow exists for this provider — the model choice is still valid (its validity is
     // independent of credentials), so KEEP it and name the remedy; the startup auth report warns too.
     log.warn(`[fastagent] "${provider}" has no interactive login — set its API key env var; invokes fail until then`);
--- a/src/cli/shared.ts
+++ b/src/cli/shared.ts
@@ -1,22 +1,29 @@
 /**
  * Helpers shared across command modules: interactivity gates, port parsing, the startup auth report,
- * and first-run model resolution. Bodies moved verbatim from cli.ts; the module-scoped flag access
- * (`values.*`) became parameters.
+ * first-run model resolution, and the login terminal IO. Bodies moved verbatim from cli.ts; the
+ * module-scoped flag access (`values.*`) became parameters.
  */
 import { readFile, writeFile } from "node:fs/promises";
 import { relative } from "node:path";
-import { autocomplete, isCancel, select } from "@clack/prompts";
+import { autocomplete, isCancel, log as clackLog, password, select, text as clackText } from "@clack/prompts";
+import { buildModelPickerOptions } from "../cli-models.ts";
 import { fastagentCredentialStore } from "../engines/pi/auth.ts";
 import {
   isValidPort,
+  listModels,
   loadConfig,
+  providerOf,
   resolveAuthPath,
   resolveModelSpec,
+  resolveStateRoot,
   rewriteConfigModel,
 } from "../engines/pi/config.ts";
-import { configuredModelSpecs, createPiModels, probeAuthSource } from "../engines/pi/models.ts";
+import { ensureStateRootSelfIgnored, isUnderDir } from "../engines/pi/definition.ts";
+import { LoginCancelled, type LoginIO, loginFlow } from "../engines/pi/login.ts";
+import { createPiModels, probeAuthSource, providerAuthStatuses } from "../engines/pi/models.ts";
 import { formatAuthReport } from "../cli-auth.ts";
 import { log } from "../log.ts";
+import { openExternalUrl } from "../open-url.ts";
 import { failStartup, failUsage } from "./fail.ts";
 
 /** Both stdin and stdout are a terminal — the precondition for an interactive prompt. */
@@ -44,7 +51,7 @@ export function parsePort(value: string | undefined, source: string, from: "flag
 
 /** Report which source provides the model's credentials, surfacing a remediation hint at startup. Non-blocking. */
 export async function reportAuth(modelSpec: string, authPath: string): Promise<void> {
-  const provider = modelSpec.slice(0, modelSpec.indexOf("/"));
+  const provider = providerOf(modelSpec);
   const source = await probeAuthSource(createPiModels({ authPath }), modelSpec);
   // Only when nothing satisfies auth do we read the store (refresh-FREE) to tell "nothing stored" from
   // "stored but unusable" — see formatAuthReport for why. store.read warns on a corrupt file itself.
@@ -60,12 +67,14 @@ export async function reportAuth(modelSpec: string, authPath: string): Promise<v
 }
 
 /**
- * First-run model resolution for the serving commands. When no model is set (flag/env/config), and
- * we're on a TTY, pick one from the providers the user is logged into and persist the choice. A no-op
- * when a model is already set; on a non-TTY (CI/deploy), or with `--no-input`, or with nothing
- * configured it stays silent and lets the opener raise its clear "missing model" error. The pick is
- * exported to FASTAGENT_MODEL so a spawned `dev` worker inherits it, and best-effort written back to
- * the config so the next run is quiet.
+ * First-run model resolution for the serving commands: ONE funnel, no dead ends. When no model is set
+ * (flag/env/config) and we're on a TTY, show the FULL catalog annotated per provider — ready (with the
+ * credential source, so which account pays is visible at the decision point) or login-required — and,
+ * when the choice needs auth, run the login flow INLINE instead of exiting with "run `fastagent login`
+ * and come back". A no-op when a model is already set; on a non-TTY (CI/deploy), with `--no-input`, on
+ * cancel, or on a failed login it stays quiet and lets the opener raise its clear "missing model"
+ * error. The pick is exported to FASTAGENT_MODEL so a spawned `dev` worker inherits it, and
+ * best-effort written back to the config so the next run is quiet.
  */
 export async function resolveFirstRunModel(
   workspaceDir: string,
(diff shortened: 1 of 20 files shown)


test/models.test.ts
```diff
@@ -1,48 +1,41 @@
...
+type FakeProvider = { id: string; models: string[]; auth: "ok" | "none" | "reject"; oauth?: boolean };
 
-/** A minimal Models stub exposing only what configuredModelSpecs touches: getProviders + getAuth. */
+/** A minimal Models stub exposing only what providerAuthStatuses touches: getProviders + getAuth. */
 function fakeModels(providers: FakeProvider[]): Models {
   return {
     getProviders: () =>
       providers.map((p) => ({
         id: p.id,
+        auth: { oauth: p.oauth ? {} : undefined }, // hasInteractiveLogin probe
```
kid7st: `hasInteractiveLogin`'s second disjunct (`p.auth.apiKey?.login`) is never exercised: the fake only models `auth.oauth`. That predicate is one of this PR's core semantics (the honest "login required" vs "set the env var" annotation); if a regression drops or breaks the apiKey branch, every api-key-login provider gets mislabeled as env-only and no test fails. Direction: extend the fake's auth shape with `apiKey: { login: … }` and add an apiKey-login-only provider case asserting `interactiveLogin: true`.

src/cli.ts
```diff
@@ -1583,6 +1609,13 @@ function serve(routes: Routes, port: number, onListening?: (boundPort: number) =
       process.send?.({ type: "ready", port: boundPort }); // tell the dev supervisor we bound + on which port
       log.info(`[fastagent] http channel on :${boundPort}`);
       log.info(`[fastagent] routes: ${Object.keys(routes).join(", ") || "(none)"}`);
+      // Proof of life: end setup with an observable success, not an inferred one. The body ships
+      // from the http channel (INVOKE_EXAMPLE_BODY) so this hint can't drift from the protocol.
+      if ("POST /invoke" in routes) {
```
kid7st: Route-key presence (`"POST /invoke" in routes`) is used as a proxy for "the built-in invoke protocol is mounted", but the drift-proof guarantee (`INVOKE_EXAMPLE_BODY` pinned by test) only holds for the built-in `createInvokeHandler`. A user channel declared at the same key with a different body shape wins in routesFor — the printed curl then 400s while the comment still claims the hint can't drift. Direction: print the hint only when routesFor took the fallback (no user channels), e.g. by returning that provenance alongside the routes.

kid7st (commented):
Review thesis: the PR replaces the filtered first-run menu with a full-catalog picker backed by 3-state `ProviderAuthStatus` data plus inline `loginFlow`; the semantics are coherent. 3 findings — 2 inline, 1 here (outside the diff):

**`src/cli-models.ts:63` — the `providerOf` fix misses a caller in the very file that imports it.** `providerOf`'s stated reason to exist (its doc comment) is eliminating the `slice(0, indexOf("/"))` silent-truncation pattern, and this PR imports it into cli-models.ts — yet `providerMatch` at line 63 still uses `spec.slice(0, spec.indexOf("/"))`. This contradicts the PR's own centralization thesis. Direction: `providerOf(spec).toLowerCase().includes(q)`.

kid7st:
All three review findings addressed in 2e2d29f:

1. **test/models.test.ts** — fake's auth shape now models both `hasInteractiveLogin` disjuncts; added an apiKey-login-only provider asserting `interactiveLogin: true`.
2. **src/cli.ts try-it hint** — `routesFor` now returns `{ routes, builtinInvoke }`; the curl hint prints only when the built-in invoke handler was mounted (a user channel at the same key may speak a different body shape).
3. **cli-models.ts providerMatch** — routed through `providerOf` (the silent-truncation pattern the helper exists to eliminate).

kid7st:
Branch rebuilt on current main (commander CLI restructure #243 + pi 0.80.10 fix #249) — the old smoke failure is gone with it. Semantics unchanged from the reviewed version except two refinements noted in the PR body (failed-login keeps the model choice; neutral hint for the unreachable unknown-provider branch). Full local verification + a fresh rv pass are green.

Sukitly (approved):
LGTM

src/cli/shared.ts
```diff
@@ -77,36 +87,141 @@ export async function resolveFirstRunModel(
...
- * `model:` placeholder the scaffold writes (or an existing `model:` line); anything else (zero-config,
- * a hand-shaped config) is left untouched with a printed hint. Never throws — persistence is a convenience.
+ * Interactive login with the api_key quick-fail probe closed into a LOOP: a definitively rejected key
+ * (HTTP 401) deletes the bad credential and RE-PROMPTS immediately — the user's hands are on the
+ * keyboard NOW; parking the failure for a later `fastagent login` would waste that. The loop exits on
+ * a verified/unverifiable key (kept), an OAuth login (completing the flow already proved the
+ * credential), or cancel (LoginCancelled propagates to the caller's cancel policy). Used by both the
+ * `login` command and the first-run picker's inline login.
+ */
+export async function loginWithKeyCheck(
```
kid7st, line 153: `loginWithKeyCheck` 的删除-重问循环把所有依赖硬编码, 不可测且未测. `loginFlow` 的设计前提是 IO 可注入 ("The terminal IO and providers are injected, so the routing is testable without real stdin"), 但这里内部构造 `terminalLoginIO()`, `verifyApiKeyLogin` 内部构造 `createPiModels` 并直接 `fastagentCredentialStore(authPath).delete(provider)`. 结果是这条**销毁凭证状态**的策略链 (rejected → delete stored credential → pin `method:"api_key"` → 重跑 loginFlow) 只有底下的 `probeApiKey` 判定有单测, 策略本身零覆盖. PR body 的 declined-test 只豁免 cli.ts 里"三行 side-effect sequencing", 这是 ~60 行带循环、带 state 删除的逻辑, 不在那条豁免里.

建议: 让 `loginWithKeyCheck` 接受 io (及 verify 探针) 作为带 terminal 默认值的参数, 加一个 fake-io 测试覆盖 "rejected → 删除 → 只重问 key → ok 退出" 主循环.

src/cli/shared.ts
```diff
@@ -77,36 +87,141 @@ export async function resolveFirstRunModel(
...
+    options: buildModelPickerOptions(listModels(models), statuses),
   });
   if (isCancel(r)) return; // cancelled: let the opener report the missing model
   const chosen = r as string;
+  const provider = providerOf(chosen);
+  const status = statuses.get(provider);
+  if (status && status.state !== "ready" && status.login === "none") {
+    // No login flow exists for this provider — the model choice is still valid (its validity is
+    // independent of credentials), so KEEP it and name the remedy; the startup auth report warns too.
+    log.warn(`[fastagent] "${provider}" has no interactive login — set its API key env var; invokes fail until then`);
```
kid7st: `broken` + `login:"none"` 被并进 env-var 分支, 给出的 remedy 与 auth 解析顺序矛盾. 分支条件 `status.state !== "ready" && status.login === "none"` 让 `broken` (stored credential 解析抛错) 也落进 "set its API key env var; invokes fail until then" — 但按 `createPiModels` 的语义 "A stored credential owns the provider; env is consulted only when nothing is stored", broken 的 stored credential 仍然 owns provider, 设 env var 修不好它; 真正 remedy 是修/删 stored credential. 且 `status.message` 在这条 log 里被丢弃. 同源问题在 `src/cli-models.ts` 的 hint 拼接: broken+none 渲染成 "API key required — set the provider's env var — stored auth unusable: …", 前后两半 remedy 互相矛盾.

建议: broken 时不走 env-var 措辞, 输出指向 stored credential 的 remedy, 如 "stored auth unusable: ${message} — fix or remove it in <authPath>".

kid7st (commented):
Review thesis: 首跑 picker 改为全目录 + 三态 provider auth 标注 (`providerAuthStatuses`) + inline login (`loginWithKeyCheck` 的 401 验证-删除-重问循环), 外加 serve 后的 try-it 提示 (`INVOKE_EXAMPLE_BODY` + `builtinInvoke` provenance). 状态建模与责任分配 (rejected 定谳才删 state) 总体自洽. 2 findings, 均为 inline.

kid7st:
Both audit findings on Sukitly's commits addressed in 17a51c0:

1. **broken + no-flow remedy** — a broken stored credential owns the provider (env is consulted only when nothing is stored), so the env-var advice was unreachable-by-design wrong. Picker hint now reads `stored auth unusable: <msg> — fix or remove the stored credential`; the post-pick warn names the auth path. Pinned in cli-models tests.
2. **loginWithKeyCheck testability** — the loop takes `seams: { flow?, verify? }` (production callers omit both); new `test/cli-shared.test.ts` pins the policy: rejected → re-ask ONLY the key (provider+method pinned on retry), unknown/oauth exit without retry, cancel propagates.

PR #253, opened 2026-07-18
feat: extend the first-run funnel to chat and deploy

kid7st:
## What

Follow-up to #250: the first-run funnel ("no model set + a TTY → pick, auth inline, persist, serve") covered `dev`/`start`/`invoke`/`fire` but left the two remaining interactive commands as dead ends. This PR closes both:

- **`chat` — full picker.** The vibe-first entry (`init` → `chat`) used to die with `missing model`. It now runs the same credential-annotated picker (inline login, `--auth-path`) as every other assembly command and persists the choice. (An earlier revision gave chat a plain `pickOnly` catalog because chat then authenticated through pi's own `~/.pi` store; #255 moved chat onto the shared session builder with fastagent's credential store, so that divergence — and `pickOnly` — is gone.)
- **`deploy` — full picker, before `loadConfig`.** The picker's write-back lands the model in `fastagent.config.*` — exactly what the model-travel gate requires (`--model`/env don't reach the deployed box) — and an inline login stores the credential that `--run` then carries. `deploy` gains `--no-input`, matching the other assembly commands (plan mode still only warns on a missing model; `--run` gates — documented).

## The bug this surfaced

`loadConfig` imports the config via ESM `import()`, which caches by URL — so a config **rewritten in the same process** (the picker's write-back) read back stale: deploy's travel gate would reject the model one line after printing `saved model "X"`. Fixed at the root: `loadConfig` cache-busts on file mtime (unchanged files stay cached; mtime resolution is the named ceiling). A failed write-back (zero-config / hand-shaped config) still gates, correctly — freshness, not a bypass. Pinned by a test (rewrite → second load sees the new value, mtime pushed past fs granularity).

## Named trade-offs

- The deploy `--no-input` wiring and the prompt path are interactive glue, untested (same declared boundary as #250); the testable regression class (config freshness) is what got the test.
- `chat` has no `--no-input`: the TUI is interactive by definition.

## Verification

`npm run lint && npm run typecheck && npm test` green (813 tests); `rv.sh local` after the #255 rebase → findings fixed (stale cross-reference in cli.md, over-promising persist hint, `--no-input` table exception, mtime-granularity test pin), rerun → No findings.

### DIFF
--- a/docs/ai-start.md
+++ b/docs/ai-start.md
@@ -40,7 +40,7 @@ All three paths continue with the same steps below: inspect, authenticate, initi
 
 You are non-interactive, so do not rely on prompts you cannot answer.
 
-- A model must be explicit. Without one, `dev`, `start`, and `invoke` open an interactive picker and fail in a non-TTY with `missing model`.
+- A model must be explicit. Without one, `dev`, `start`, `invoke`, `fire`, and `chat` open a picker on a TTY and fail in a non-TTY with `missing model`. `deploy` also opens the picker on a TTY, but non-interactively it does NOT say `missing model`: plan mode warns, and `--run` stops at the model-travel gate ("no model in fastagent.config.ts").
 - `fastagent login` is also interactive. Ask the user to run it in a terminal inside the workspace; it writes project-level `.fastagent/auth.json`, and credentials written in another directory are not visible here.
 - Alternatively, ask the user for a provider API key and put it in `.env` only with permission.
 - List available specifications with `fastagent models`.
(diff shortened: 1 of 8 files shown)


Sukitly (approved):
lgtm

kid7st:
Rebased onto main (#255 landed underneath). #255 moved chat onto the shared session builder with fastagent's credential store, which removed the premise for chat's plain-catalog `pickOnly` mode — dropped it; chat now runs the same credential-annotated picker (+ inline login, `--auth-path`) as the other assembly commands. Also closed post-rebase rv findings: stale cross-reference in docs/cli.md, the persist hint's "for this run" over-promise on the deploy path, the `--no-input` table's deploy exception, and pinned the cache-bust test above fs mtime granularity. Full suite green (813). PR body updated to match.

PR #254, opened 2026-07-18
docs: ai-start covers chat, skill vendoring, and Feishu/Lark onboarding

kid7st:
Three pure additions to the agent-facing quickstart (recovered from an earlier working session; verified against the current code):

- **`fastagent chat`** as the interactive path — same assembled agent, model, tool, and skill resolution as `dev`, in pi's TUI.
- **`fastagent add skill <source>`** for vendoring an existing Agent Skills skill (git ref `owner/repo/path`, local path, or a bare name from `~/.agents/skills`) instead of writing one.
- **Feishu / Lark onboarding**: `add feishu` (scan-to-confirm app creation, credentials to `.env`, one manual version-publish click remains) and `add lark` (international console guidance). Both interactive; `dev --tunnel` for local webhook testing.

Docs-only; `npm run lint` (incl. doc-link check) green. Note: touches the same file as #253 (different regions, no conflict).


Sukitly (approved):
lgtm

PR #258, opened 2026-07-20
feat(session): Phase 3 — the HTTP+SSE transport (controlRoutes / connectSessionControl / fastagent attach)

kid7st:
## What

The session control plane's transport, per [design §13/§15 Phase 3](https://github.com/fastagent-sh/fastagent/blob/main/docs/design/session-control.md): **one transport serves every remote consumer** — Web panel, desktop app, CLI attach. The subprocess adapter stays demand-driven.

### Server — `controlRoutes` (engine-neutral, `/core`)

`GET /control/capabilities|state|entries|events` + `POST /control/dispatch`, bearer token REQUIRED (timing-safe compare; an unauthenticated remote-control endpoint is not constructible by omission). The envelope is born at the wire: `{sessionId, epoch, seq, event}` per SSE message. Parse-don't-validate on dispatch: every command variant is rebuilt field by field — malformed input answers protocol-level `invalid_command`, never a cast into the trusting inner layers. Eager subscription registration (subscribed before response headers), heartbeats, and teardown discipline on every path (cancel, errored pull, rejecting implementations).

### Client — `connectSessionControl` (`/core`)

The SAME `SessionControl` interface, isomorphic to local (conformance-tested over a real HTTP server). Envelope consumed internally: a seq gap ends the events iterator (logged) into the standard reconnect steps; restart rides the disconnect rule; epoch is informational across connections. Structured `ControlRequestError.status` for 401 branching; CRLF-tolerant SSE parsing; each iteration opens its own connection; detach from a quiet stream resolves promptly on both sides.

### Serve wiring + `fastagent attach`

`config.sessionControl: true` → dev/start mount the routes, mint a per-boot token, write `<stateRoot>/control.json` (0600, removed on clean exit; signal handlers re-raise after cleanup — a listener must not absorb termination or dev's watch restart hangs). `--tunnel` warns loudly that `/control/*` goes public. `fastagent attach <session>`: stream live events, steer by typing, `/abort`, reconnect rounds with rendered replay + state re-check; one round = one stream on every failure path; 401 exits with the re-attach hint.

### Hub hardening surfaced by the transport

Subscriber waiter queue (concurrent `next()` both settle), done-is-terminal (post-done `next()` never re-registers), backlog cap (10k — a stalled remote connection closes its stream instead of growing server memory; semantically lossless).

## Verification

- lint / typecheck / test: **848/848** (23 new: fail-closed auth, local/remote isomorphism, SSE round-trip + envelope hygiene, seq-gap termination, quiet-stream detach both sides, concurrent pulls, backlog cap, attachRound backfill/401/failure-discipline, mountSessionControl, wire command parsing).
- `rv.sh local`: **10 review rounds, 24 findings, all addressed** (subscription leaks, unhandled-rejection paths, signal semantics, timing-safe auth, empty-token fallback, waiter-queue hang…). Final round: one comment fix.

## Wire note: `/invoke` now emits SSE comment heartbeats

The public invoke channel's SSE stream gained `: ping` comment frames every 30s (the emitting half of the remote watchdog — a quiet long tool call must stay distinguishable from a dead connection, on every SSE surface). Spec-conforming SSE parsers are unaffected; a naive client that `JSON.parse`s every non-empty line (which happened to work on the old wire) will now see non-`data:` lines and must parse per the SSE spec (documented in api-reference). This is a deliberate, additive-but-observable change to the existing wire, chosen over gating heartbeats to `/control/invoke` so all SSE surfaces speak one protocol.

## Surface note: the neutral transport exports live on `/core`

Session-control layering across subpaths, stated as a decision: the contract (SessionControl types + error codes) stays behind `/session`, the pi hub (`createPiSessionControl`) behind `/pi`, and the engine-neutral transport — `controlRoutes` (fetch-shaped server routes) and `connectSessionControl`/`connectAgent` (contract-consuming clients) — joins the channel kit on `/core`: no pi import anywhere in their closure, same nature as `createInvokeHandler`/`router`. Design §1's "separate subpath" promise is about the minimal handler CONTRACT (`src/agent.ts`), which is untouched.


src/cli/commands/attach.ts
```diff
@@ -0,0 +1,287 @@
...
+  session: string,
+  cursor: string | undefined,
+  io: AttachIo,
+): Promise<string | undefined> {
+  // The round HOLDS its subscription's iterator: one round = one stream, on every path — a
+  // backfill failure must close it before propagating, or the caller's retry round would stack a
+  // second concurrent stream interleaving the same session's output.
+  const iterator = control.events(session)[Symbol.asyncIterator]();
+  let authError: unknown;
+  const draining = drainEvents(iterator, io).catch((error) => {
```
kid7st: **live drain 与 backfill 渲染并发写同一输出 — 主场景 (active run 中重连) 下 replay 块会被打花。** round 的顺序是: 启动 `drainEvents` (后台立即开始写 live 事件) → settle 等待 → 渲染 backfill (`[replaying...]` … `[end of replay]`) → state 行。`drainEvents` 在 backfill 渲染期间并发运行: 若重连时恰有 run 在跑 (attach 重连的**最常见**场景), live 的 `message_delta` 原始 chunk 会插进 replay 块中间, 输出行级交错。`[may overlap what you saw live]` 标注的是内容重复, 不是交错 — 后者没有被任何契约覆盖。

建议: 订阅时机与消费时机拆开 — 提前调用一次 `iterator.next()` 并持有该 promise (第一次 next 即建立连接/订阅, 服务端 eager registration 保证 headers 之后的事件不丢; 后续事件缓冲在 SSE/ReadableStream 与 hub Subscriber 里, 无损), 渲染完 backfill + state 行**之后**再进入 drain 循环 (从持有的首个 result 开始消费)。不改变订阅窗口, 消除交错。

src/channels/control.ts
```diff
@@ -0,0 +1,229 @@
...
+ *  the two Prompt-bearing wire surfaces reject oversized bodies at the same line. */
+const DISPATCH_BODY_LIMIT = 1 << 20;
+
+/**
+ * Parse-don't-validate at the wire: a remote client can send any JSON, and the hub's inner layers
+ * trust command shapes (a malformed `steer` would surface as an ENGINE failure misclassified as
+ * `run_command_failed`). Returns the typed command, or undefined for anything malformed — which
+ * answers protocol-level `invalid_command`, same responsibility as the hub's unknown-type default.
+ */
+function parseWireCommand(raw: unknown): SessionCommand | undefined {
```
kid7st, line 51: **parse-don't-validate 的核心分支零测试覆盖。** PR body 把 "every command variant is rebuilt field by field — malformed input answers protocol-level `invalid_command`" 列为传输层核心机制, 代码注释还专门点名 element 级校验 (`images: [42]`)。但测试只覆盖了 unknown `type` (`make_coffee`) 一个分支: malformed steer prompt (`prompt.text` 非 string / `prompt` 缺失)、`images` 元素非法、`compact.instructions` 非 string、`set_model.model` 非 string — 全部未测。删掉 `promptOk`/`imageOk` 的任何一行, 848 个测试照样全绿 — 声称的保护不存在于测试网里。

建议: 在 control-http.test.ts 加一组 table-driven 的 malformed dispatch 用例 (经真实 HTTP route 打 `/control/dispatch`), 每个 command variant 至少一个坏形状, 断言 `invalid_command` + ok:false 且非 500。

kid7st (commented):
Review thesis: Phase 3 传输层 — `controlRoutes` (bearer 强制, envelope 在 wire 出生, dispatch 逐字段重建)、`connectSessionControl` (同一 SessionControl, envelope 内部消化, seq gap → 终止 → 标准 reconnect)、serve 接线 (per-boot token + control.json 0600 + 信号清理再抛)、`fastagent attach`、hub 加固 (waiter 队列 / done-terminal / backlog cap)。安全叙事、teardown 纪律与真 HTTP conformance 测试粒度都到位。2 个 findings, 均为 inline。

src/session-remote.ts
```diff
@@ -0,0 +1,393 @@
+/**
+ * The remote `SessionControl` — the client half of the Phase 3 transport (design §13). Engine- and
+ * server-neutral: speaks only the wire protocol `controlRoutes` serves (HTTP JSON + SSE with the
+ * {sessionId, epoch, seq, event} envelope) and re-exposes the SAME `SessionControl` interface, so
+ * local and remote consumers are isomorphic — client code does not change when the agent moves out
+ * of process.
+ *
+ * Envelope consumption is internal: a seq gap (loss in transit on this connection) ENDS the events
```
kid7st: **seq-gap / 重启的语义已改为 throw, 但两处伴随文本仍说 "ends the iterator"。** 实现在 68a3047 里刻意把 seq gap 从 clean end 改成 **throw** ("the consumer's failure budget and diagnostics own it"), server restart 也是连接断 → 读错误 → generator **rethrow** — design §13 已如实写 "a gap throws in the client"。但本文件的模块头仍是旧语义: "a seq gap **ENDS** the events iterator — the consumer then runs the standard reconnect steps" / "a broken stream is visible as a **terminated iterator**"; `docs/api-reference.md:487` 同样: "a sequence gap or a server restart **ends** the events iterator"。按文档写 `for await` 不包 try/catch 的消费者, 在 gap/重启时拿到的不是 clean end 而是未捕获异常。三份文本 (代码头、api-reference、design) 两对一, 代码头描述的正是它自己文件里已不存在的行为。

建议: 两处改为 throw 语义并说明动机 — "a seq gap (and any mid-stream transport failure, a server restart included) **throws** from the events iterator so the consumer's failure handling / budget owns it; only the consumer's own detach reads as a clean end. Recovery is the standard reconnect steps." 与 design §13 措辞对齐。

kid7st (commented):
Re-review after 48 new commits. 上轮 2 条 findings 均已落实 (replay 缓冲 + flush 专测; wire parser delivery-through-transport + junk-strip + element-level 拒绝测试)。本轮 delta 把 Phase 3 扩成完整远程实例: `connectAgent` 数据面 (SPEC 全纪律)、`decideRound` 纯策略、compact accept-fast + abort door、latched cancellation door、Subscriber cap 冻结语义 — 机制对称性与测试覆盖扎实。全量重审发现 1 个问题 (inline)。

src/cli/commands/attach.ts
kid7st: Addressed (earlier round): live output is buffered while the replay block prints and flushed after `[end of replay]` — the failure path releases the buffer too. Buffering was chosen over holding the first `next()`: same subscription window, no interleaving, and events keep flowing into the buffer instead of parking in transport. Pinned by "attachRound buffers live output during the replay block and flushes it after, failure path included".

src/channels/control.ts
kid7st: Addressed: the delivery-through-transport test (junk-strip + element-level `images: [42]` rejection) landed earlier; 2a414ec completes the net with a table-driven malformed case per command variant (prompt missing / text non-string / prompt non-object / instructions / model / level non-string) over real HTTP — removing any parseWireCommand check line now turns one red.

src/session-remote.ts
kid7st: Fixed in 2a414ec: both texts (session-remote module header + api-reference) now state the throw semantics — a seq gap and any mid-stream transport failure (server restart included) throw from the events iterator so the consumer's failure handling/budget owns it; only the consumer's own detach reads as a clean end. Wording aligned with design §13.

kid7st (commented):
No findings. (Re-review after 2a414ec: the throw-semantics companion texts are now aligned across session-remote.ts, api-reference, and design §13; the malformed-dispatch table closes the remaining parseWireCommand coverage gap variant by variant. Full-diff re-pass found nothing new.)

PR #259, opened 2026-07-20
feat: add Feishu and Lark WebSocket channels

Sukitly:
## Summary

Add first-class Feishu/Lark WebSocket ingress without changing the existing HTTP route-channel contract.

- Keep `ChannelModule = (ctx) => Routes` for webhook/request channels.
- Add the explicit `LongConnectionChannelModule` shape: `{ name, connect(ctx, signal) }`, where `connect` returns `{ ready, closed }` and `AbortSignal` is the only shutdown command.
- Add transport-specific factories:
  - `feishuChannel()` / `larkChannel()` for webhooks.
  - `feishuWebSocketChannel()` / `larkWebSocketChannel()` for outbound WebSocket connections.
- Feed webhook and WebSocket delivery into the same Feishu event acceptance, durability, queue, session, context-buffer, attachment, preview-card, and reply engine.
- Make context-aware groups the recommended onboarding path, while preserving an explicit mention-only least-privilege choice.

The official Lark SDK owns handshake and transient reconnect behavior. FastAgent reports ready only after the first successful connection, keeps health stable across transient reconnects, turns terminal transport failure into a visible serving failure, and translates framework abort into `WSClient.close()`.

### Local setup

The standard path is now:

```bash
fastagent add feishu --ingress websocket --group-behavior context
fastagent dev
```

Interactive onboarding presents:

```text
Context-aware groups (recommended)
  Bare replies in Agent-created threads invoke the Agent.
  Other unsummoned group discussion is buffered for the next @Agent turn.
  Requires im:message.group_msg, which delivers all group messages to the app.

Mention-only (least privilege)
  Only explicit @Agent messages are delivered.
  No group-wide message permission is required.
```

The generated channel is ordinary authored TypeScript:

```ts
import { feishuWebSocketChannel } from "@fastagent-sh/fastagent/feishu";

export default feishuWebSocketChannel({
  appId: process.env.FEISHU_APP_ID ?? "",
  appSecret: process.env.FEISHU_APP_SECRET ?? "",
});
```

WebSocket mode needs no public URL, tunnel, Verification Token, or Encrypt Key. The serving process still exposes `/health` for hosting readiness.

### Group visibility is explicit instead of silently degraded

`add feishu|lark` now inspects the app's scope grant state. For context-aware groups it attempts to add `im:message.group_msg` to the app draft through application-v7 config, opens Permissions for tenant-admin approval, and does not send the user to version publishing until permission work is ready. Lark control-plane failures degrade to an explicit manual path.

Serving also reports the effective capability:

```text
[feishu] group visibility: context-aware — bare managed-thread replies + buffered discussion enabled
```

or:

```text
[feishu] group visibility: @mentions only — im:message.group_msg is not granted; ... unavailable
```

### Existing webhook users

Existing `feishuChannel()` and `larkChannel()` modules keep their current factory and route contract. There is no runtime transport switch or hidden fallback.

Migrating an existing app requires both sides to move together:

1. Replace the authored factory with `feishuWebSocketChannel()` / `larkWebSocketChannel()`.
2. Change Events & Callbacks to long connection.
3. Publish a new app version.
4. Remove webhook-only secrets if no longer used.

Use separate apps when development and production intentionally use different ingress modes.

## Serving and deployment

- Channel discovery classifies the module by its default-export structure: function = HTTP routes; object with `connect()` = long connection.
- `dev`/`start` bind HTTP health/routes and open all long connections, reporting ready only when both surfaces are usable.
- SIGINT/SIGTERM abort long connections, stop accepting HTTP work, force-close active HTTP streams, and retain the bounded non-draining shutdown contract.
- All structurally detected long-connection channels, including custom ones, force always-on deployment:
  - Fly: one running machine.
  - Railway: App Sleeping must remain disabled.
  - Webhook/tunnel registration is skipped for long-connection channels.
- Deploy preflight imports channels to inspect their shape. First-party route credentials are therefore validated at activation rather than import time, so plans remain generatable before production secrets are provisioned while `deploy --run` and serving still fail before side effects.

## Reliability and protocol behavior

- Pre-ACK persistence failures propagate through the real SDK `EventDispatcher`; the SDK returns a 500 ACK so Feishu can redeliver, and no phantom dedup state is written.
- Temporary disconnects do not settle `closed` or flap health; the SDK reconnects.
- Terminal setup/retry exhaustion rejects `closed` and fails serving visibly.
- Webhook and WebSocket paths share message dedup, durable turn intent, managed-thread ownership, context buffering, FIFO session queues, attachment handling, and streaming-card delivery.
- Deployment channel import/inspection failures block planning instead of silently producing an incorrect topology.

## Examples

Webhook remains unchanged:

```ts
import { feishuChannel } from "@fastagent-sh/fastagent/feishu";

export default feishuChannel({
  appId: process.env.FEISHU_APP_ID ?? "",
  appSecret: process.env.FEISHU_APP_SECRET ?? "",
  verificationToken: process.env.FEISHU_VERIFICATION_TOKEN ?? "",
});
```

A custom long-connection channel uses the same lifecycle contract:

```ts
export default {
  name: "custom stream",
  connect(ctx, signal) {
    // Start outbound transport; translate signal abort to the transport close primitive.
    return { ready, closed };
  },
};
```

## Validation

- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm run build`
- [x] `npm test` — 77 files / 862 tests
- [x] Real Feishu tenant smoke: WebSocket handshake/readiness, p2p turns, group @mention turns, threaded session continuity, and streaming-card delivery
- [x] Real Feishu scope-list response verified (`im:message.group_msg`, tenant scope, granted)

A real Lark-tenant smoke test is still outstanding; Lark compatibility behavior is covered by unit/integration tests and explicit control-plane fallbacks.

## Checklist


Sukitly:
Review note on the non-Feishu-looking changes:

- `src/channels/telegram/telegram.ts` and `src/channels/github/github.ts`: credential validation moved from factory construction to channel activation. Deploy preflight now imports channel modules to classify route vs long-connection shape before production secrets exist; serving and `deploy --run` still activate them and fail fast before side effects. No Telegram/GitHub routing behavior changed.
- `src/channels/state.ts` and `src/channels/turn-store.ts`: comment-only wording change from “webhook ACK” to transport ACK, because the same pre-ACK durability boundary now feeds HTTP responses and WebSocket ACK frames.
- `src/host/node.ts` / `src/cli/serve.ts`: long connections need one coordinated shutdown path. `closeAllConnections()` prevents an active HTTP/SSE request from making SIGTERM wait indefinitely; shutdown remains bounded and non-draining.
- `src/tunnel.ts` / `src/dev-supervisor.ts`: pass only structurally detected route channels to webhook registration. Otherwise a Feishu WebSocket module would incorrectly start/register a tunnel merely because `channels/feishu.ts` exists.
- `src/deploy/**` and `src/cli/commands/deploy.ts`: deployment must distinguish route channels from all long-connection channels (including custom ones) to keep one process running, disable sleeping/scale-to-zero, omit webhook-only secrets, and skip webhook registration.
- `src/core.ts`, `src/host/node.ts`, `src/engines/pi/channel.ts`, `src/feishu.ts`, and `src/lark.ts`: these are the minimal public/discovery/serving seams required for the explicit long-connection contract and transport-specific factories; the existing route `ChannelModule` contract remains unchanged.

So the cross-cutting diff is lifecycle/deploy plumbing required by the new channel shape, not a Telegram/GitHub feature rewrite.

src/channels/feishu/ws-ingress.ts
```diff
@@ -0,0 +1,161 @@
...
+  const fail = (error: unknown): void => {
+    if (signal.aborted || closedSettled) return;
+    const failure = error instanceof Error ? error : new Error(String(error));
+    if (!readySettled) {
+      readySettled = true;
+      rejectReady(failure);
+    }
+    closedSettled = true;
+    rejectClosed(failure);
+  };
```
kid7st, line 109: **Terminal failure never closes the client, and the subsequent abort can't either.** `fail()` only rejects `ready`/`closed`; it never calls `closeClient`. After it sets `closedSettled = true`, the abort listener's `close()` early-returns on the `closedSettled` guard — so `WSClient.close()` is never invoked on this path. The stated contract ("AbortSignal is the sole shutdown command") stops holding after a terminal failure. SDK 1.71.1 sets `terminalError` and stops reconnecting, but resources only `close()` destroys (e.g. the dataCache sweep interval) outlive the connection. In the CLI this is masked by `failStartup`'s `process.exit`; `feishuWebSocketChannel` is public surface, so embedders leak for real.

Suggested direction: in `fail`, best-effort `closeClient(client)` before rejecting — and add the missing test asserting the client is closed on the terminal-failure path (the current lifecycle test covers abort-close but not failure-close, which is exactly where this bug lives).

src/cli/commands/add.ts
```diff
@@ -120,14 +130,111 @@ export async function runAddChannel(
...
+
+async function resolveGroupBehavior(kind: ChannelKind, raw: string | undefined): Promise<FeishuGroupBehavior> {
+  if (raw !== undefined && raw !== "context" && raw !== "mentions") {
+    failUsage(`--group-behavior must be "context" or "mentions", got "${raw}"`);
+  }
+  if (kind !== "feishu" && kind !== "lark") return "context";
+  if (raw !== undefined) return raw;
+  if (!(process.stdin.isTTY && process.stdout.isTTY)) {
+    console.error(
+      `[fastagent] no interactive terminal — defaulting ${kind} group behavior to context-aware (use --group-behavior mentions for least privilege)`,
```
kid7st: **A defaulted (not chosen) value drives a remote sensitive-scope write.** Non-TTY `add feishu|lark` without `--group-behavior` defaults to `context`, after which `configureGroupBehavior` PATCHes the sensitive `im:message.group_msg` scope into the app draft (`addAppScopes`) and opens approval pages. Combined with the other behavior change — the complete-credentials webhook re-run path used to be a pure no-op ("already set — keeping them") and now runs `configureGroupBehavior` on every re-run — a scripted re-run against an app deliberately configured mention-only silently escalates its requested permission set, with only a stderr note.

Suggested direction: in non-TTY mode, require the explicit `--group-behavior context` flag before taking the scope-mutating path; the defaulted path should inspect + report only.

src/host/node.ts
```diff
@@ -27,11 +27,23 @@ export interface ChannelContext {
   stateRoot: string;
 }
 
-/** A `channels/<name>.ts` default export: receives the mount context, returns the routes it mounts.
- *  Adapters (`telegramChannel(opts)`, `githubChannel(opts)`) RETURN one of these, so user glue holds
- *  only policy — the framework pipes `agent`/`stateRoot` to the adapter without transiting user code. */
+/** A `channels/<name>.ts` route channel: receives mount context and returns its HTTP routes. */
 export type ChannelModule = (ctx: ChannelContext) => Routes;
 
+/** One logical long connection's lifecycle. `ready` settles after its first usable connection;
```
kid7st: **The `ready` doc contradicts the reference implementation on abort-before-ready.** This says "`ready` settles after its first usable connection", but `ws-ingress`'s `close()` resolves `ready` as success on a pre-ready abort with no connection ever made — and `serve()` then logs `long connection ready: <name>` and calls `markReady()` mid-shutdown. This is a freshly minted public contract; custom channel authors need to know what `ready` must do when the signal aborts before the first connection (hang? resolve? reject?).

Suggested direction: document the abort-before-ready semantics here (e.g. "on abort, `ready` must settle; resolution is treated as cancellation, not readiness"), and have `serve` skip the ready log/`markReady` when `abort.signal.aborted`.

kid7st (commented):
Review thesis: this PR introduces `LongConnectionChannelModule` — a second structural channel form `{ name, connect(ctx, signal) → { ready, closed } }` — with the Feishu/Lark WebSocket adapters as first implementations, and threads always-on semantics through serving and deploy. The core concept is clean; 3 findings, all inline.

kid7st:
Pushed bcca469 addressing the three review findings (threads resolved):

- **ws-ingress**: `fail()` now settles `closedSettled` first (re-entry-safe) and best-effort-closes the client — terminal failure no longer leaks the SDK transport for embedders. Added the missing failure-close assertion.
- **add feishu|lark**: a defaulted (non-TTY, no `--group-behavior`) context choice now inspects + reports only; the `im:message.group_msg` draft write requires an explicit flag or interactive selection. New `GroupBehaviorChoice` threads explicitness through onboarding.
- **LongConnection contract**: abort-before-ready semantics documented (resolution = cancellation); `serve()` skips `markReady`/ready logs when the signal is already aborted.

Validated locally: lint + typecheck + 863 tests green.

kid7st (approved):
LGTM

PR #260, opened 2026-07-21
refactor(pi): THINKING_LEVELS has one definition — the one with the drift guard

kid7st:
config.ts and harness.ts each defined pi's runtime thinking-level list, and each claimed to be the single source. Only harness.ts's Set carries the exhaustiveness anchor (`satisfies Record<ThinkingLevel, true>`): pi adding a level breaks the build instead of config validation silently falling behind. It wins; config validation imports it, following the existing config→harness type-import direction (AnyModel).

- No behavior change: Set insertion order keeps the error message's level order identical.
- Verified: typecheck / lint / 835 tests green; `rv.sh local`: no findings.

### DIFF
--- a/src/engines/pi/config.ts
+++ b/src/engines/pi/config.ts
@@ -13,20 +13,11 @@ import { pathToFileURL } from "node:url";
 import type { ThinkingLevel } from "@earendil-works/pi-agent-core";
 import type { FastagentTool } from "./tool.ts";
 import type { Models } from "@earendil-works/pi-ai";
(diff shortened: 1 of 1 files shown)


kid7st (commented):
No findings. (THINKING_LEVELS single-sourced to the anchor-bearing harness.ts Set; config→harness import direction follows the AnyModel precedent; Set insertion order keeps the error-message level order — the no-behavior-change claim holds; no stale consumers repo-wide.)

PR #261, opened 2026-07-21
refactor(cli): the presenter modules live under src/cli/

kid7st:
cli-models.ts, cli-auth.ts, cli-add-feishu.ts and invoke-stream.ts predate the commander rebuild (#243) and stayed at the package top level next to the public subpath entries, though every consumer is CLI-side. Moved under src/cli/ — the cli- prefix is redundant there: **models-view.ts / auth-view.ts / add-feishu.ts / invoke-stream.ts**. Tests renamed to match; AGENTS.md map and the package-boundary neutrality whitelist track the new paths.

Pure move — no behavior change. Verified: typecheck / lint / 835 tests green; `rv.sh local`: one stale-comment finding, fixed.


kid7st (commented):
No findings. (Pure move verified: zero stale references to the old paths repo-wide; none of the moved modules were on the public surface (no index/core/pi/package.json references, so nothing breaks); the package-boundary neutrality whitelist tracks cli/invoke-stream.ts; tests and AGENTS.md map follow.)

PR #263, opened 2026-07-21
feat(pi): add session-aware tool context

Sukitly:
## Summary

- replace the duplicated `ToolContext.session` string with a FastAgent-owned read-only session manager exposing session identity, header, and branch history
- adapt both the stateless serving session and the resident chat session manager to the same tool-runtime contract
- bind execution cwd for serving, chat, and sessionless `fastagent tool` calls, and migrate `wake`/`unwake` to the shared session identity
- document reusable package tools as ordinary `config.tools` integrations; no extension or package-specific runtime is introduced

## Validation

- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test`
- [x] `npm run build`

`npm run lint` retains the existing unused-suppression warning in `test/control-http.test.ts:650`; it reports no errors.

## Checklist

### DIFF
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -191,11 +191,28 @@ export default defineTool({
 
 `tools/<name>.ts` files are discovered with `loadTools(dir)`, and the filename becomes the tool name.
 
-The second `execute` argument is a `ToolContext`: `{ signal?, session?, tools? }`. `session` is the id of
-the current turn's conversation (undefined outside a turn, e.g. a bare `fastagent tool` run) — a per-turn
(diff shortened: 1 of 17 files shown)


kid7st (approved):
LGTM. Serving/chat converge on one FastAgent-owned read-only session port, cwd is bound on all three execution paths, and the no-plugin-contract decision for package tools is documented. Tests cover all three producers.

One note before merge: replacing `ToolContext.session` with `sessionManager` is a breaking change to the tool-authoring surface — TS authors get a compile error, but a JS tool reading `ctx.session` silently sees `undefined`. Worth flagging in the squash commit body / release notes.

PR #266, opened 2026-07-21
feat: pi 0.81.1 summarization retries — opt-in policy wired + retry_scheduled session event

kid7st:
## Why

pi 0.81.1 (#6901 upstream) makes compaction/branch-summary model calls retry transient provider failures — but the policy is **opt-in**: `retryAssistantCall` with an undefined policy is equivalent to a direct call. Without this PR, neither of fastagent's compaction paths gets the fix. And since the upstream event source now exists, the reserved `retry_*` slot in the session-event vocabulary graduates from "future vocabulary" to an emitted event — deferring it to "when a consumer appears" would mean re-deriving today's context later.

## What

**Enable retries (commit 1)**
- `SUMMARIZATION_RETRY_POLICY` (`{ enabled: true, maxRetries: 3, baseDelayMs: 2000 }` — mirrors pi's own app defaults) wired into:
  - the `AgentHarness` config (`retry` option) → auto-compaction and branch summaries inside a run
  - the manual `compact()` dispatch in session-control (new 7th parameter); the abort door's signal still cuts through backoff sleeps
- Dependency floor → `^0.81.1`: on 0.81.0 the `retry` option, the `retry_scheduled` event, and the extra `compact()` arguments would all be silently ignored.

**Surface retries (commit 2)**
- `RetryScheduledEvent { operation, attempt, maxAttempts, delayMs, error }` joins `KnownSessionEvent` — explains a quiet gap (up to ~14s of backoff) that would otherwise read as a hang. Deliberately unclosed (the next event is the closure) and scheduled-only: the engine's `retry_finished` carries no outcome to forward, `retry_attempt_start` adds nothing over `delayMs`.
- Two provenances, both emitting: in-run (harness event → `toSessionEvent`, `runId` present) and manual compaction (`RetryCallbacks` → `emitOwn`, no `runId`). Both construction sites are typed against `RetryScheduledEvent` so a payload change breaks both at compile time.
- `log.warn` at both sites stays: events reach attached observers, warns reach operators tailing logs.
- `docs/design/session-control.md` L2 table updated; `session.ts` future-vocabulary note now lists only `turn_*`.

## Non-changes (checked, deliberately skipped)

- `classifyRetryable` stays: upstream's newly-exported `isRetryableAssistantError` is prose-only on `errorMessage`; ours is structured-signal-first (diagnostics code / HTTP status / thrown errors) and serves SPEC `failed.retryable`.
- SPEC `AgentEvent` untouched — `projectAgentEvent` drops `retry_scheduled` (the SPEC vocabulary is locked; SPEC consumers see the outcome, not the mechanics).
- 0.81.1's release-archive, extension stream-fallback, and Kimi K3 items need no code (the K3 metadata fix flows in via pi builtins).

## Verification

- `npm run lint && npm run typecheck && npm test` — 896/896 green on 0.81.1
- Upstream signatures cross-checked against installed d.ts: `AgentHarnessConfig.retry`, `compact(…, retry?, callbacks?)`, `RetryPolicy`/`RetryCallbacks`, `RetryScheduledEvent` payload (`errorMessage` → `error` rename consistent at both sites)
- `rv.sh local` run to clean across both commits; all findings addressed (silent retries, naming, version floor, dual-construction drift risk)

### DIFF
--- a/docs/design/session-control.md
+++ b/docs/design/session-control.md
@@ -299,7 +299,7 @@ compare. The vocabulary, grouped by the client maturity level that needs it:
 | L1 | `queue_changed { steering, followUp }` | Normalized queue depths. |
 | L2 | `turn_started`, `turn_finished` | Group tool activity under one assistant turn. |
 | L2 | `compaction_started/finished` | Manual compaction bounds: between runs, no `runId`; every started is closed (`summary`, `error`, or `aborted` — a deliberate stop is not a failure). Automatic overflow compaction stays inside its run and does not emit these. |
-| L2 | `retry_scheduled/finished` | Explain why a run is alive with no tokens flowing (future vocabulary — not yet emitted). |
+| L2 | `retry_scheduled { operation, attempt, maxAttempts, delayMs, error }` | A transient provider failure scheduled a summarization retry backoff — explains a quiet gap that would otherwise read as a hang. Inside a run (auto-compaction / branch summary, `runId`) or during manual compaction (no `runId`). No closing event: the next event is the closure. |
 | L2 | `state_changed { model?, thinkingLevel? }` | Material state changes. |
 
 Consumers MUST forward or ignore unknown event types; the vocabulary is additive. The contract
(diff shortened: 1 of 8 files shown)


kid7st (commented):
No findings.

PR #264, opened 2026-07-21
feat(slack): add first-party channel and app onboarding

Sukitly:
## Summary

- add a native Slack Events API channel with v0 request verification, logical/delivery deduplication, durable at-least-once turns, per-session FIFO execution, threaded DM/group sessions, owned-thread continuation, context buffering, live previews, and inbound/outbound file support
- add `fastagent add slack` single-workspace internal-app onboarding through App Manifests and OAuth, with resumable irreversible boundaries and owner-only local Configuration Token rotation
- update `dev --tunnel` and Docker/Fly/Railway `deploy --run` to update onboarded Slack Request URLs from the builder machine while keeping configuration credentials off runtime hosts
- publish `@fastagent-sh/fastagent/slack`, scaffold `channels/slack.ts` plus `tools/slack-send.ts`, and document setup, permissions, operation, deployment, and failure modes

Slack stays on the engine-neutral `Agent`/`ChannelModule` contracts and uses native `fetch`; this does not add Bolt, Socket Mode, Marketplace distribution, or multi-workspace token storage.

## Validation

- [x] `npm run lint` (passes with the existing unused suppression warning in `test/control-http.test.ts:650`)
- [x] `npm run typecheck`
- [x] `npm test` — 83 files / 936 tests
- [x] Added or updated the smallest relevant tests
- [x] `npm run build`
- [x] `npm pack --dry-run`
- [x] Live single-workspace smoke: Manifest/OAuth onboarding, automatic Quick Tunnel registration, DM, group mention, threaded continuation, and context buffering

## Checklist

- [x] Public-facing text is in English
- [x] Errors fail visibly; no silent fallbacks or swallowed exceptions
- [x] No secrets, local paths, or machine-specific state were committed
- [x] No process-only notes were added (`*_PLAN.md`, `HANDOFF.md`, session notes, etc.)
- [x] Docs were updated when behavior or public APIs changed

PR #267, opened 2026-07-22
refactor(preview): humanize tool labels in Telegram/Feishu live previews

kid7st:
**Stacked on #264 — merge after it lands.** The diff vs main includes #264's commits until then; this PR's own change is the last commit (`fc40397`).

## Summary

- wire `humanizeToolName` (added to the shared `preview-kit` by #264) into the Telegram and Feishu live-preview renderers, so all three channels show the same plain-language tool labels (`read` → "Read", `mcp__github__create_issue` → "Github: create issue") instead of raw identifiers
- update the preview-kit header comment to list the humanized tool label among the shared pieces

This closes the "single consumer in a shared module" gap: the shared placement of `humanizeToolName` is now real, and tool-label wording cannot drift between channels.

## Validation

- [x] `npm run lint` (existing suppression warning only)
- [x] `npm run typecheck`
- [x] `npm test` — 981 tests
- [x] Updated the two preview assertions that pinned raw tool names

### DIFF
--- a/src/channels/feishu/preview.ts
+++ b/src/channels/feishu/preview.ts
@@ -30,7 +30,7 @@ import {
   streamingCardJson,
 } from "./card.ts";
 import { type FeishuApi, type FeishuTarget, chunkFeishuText, isCardStreamingClosed } from "./feishu-api.ts";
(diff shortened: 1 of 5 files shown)


Sukitly (approved):
lgtm

PR #268, opened 2026-07-22
feat(slack): add `add slack --replace-config` and point registrar fallbacks at it

kid7st:
**Stacked on #264 — merge after it lands.** The diff vs main includes #264's commits until then; this PR's own change is the last commit.

## Summary

- `fastagent add slack --replace-config`: a direct, discoverable spelling for replacing the local App Configuration token pair (previously buried in the interactive menu). It skips the menu on the installed path, also covers the created-but-not-installed resume state (where a revoked token would otherwise strand the resume — rotation fails and no menu offered replacement), errors clearly when combined with `--no-onboard` or when no local onboarding state exists, and never touches the installed app or runtime credentials.
- `registerSlackWebhook` fallback messages now name the real remediation split: without local onboarding state → set the Request URL manually in the console or re-run from the onboarding machine; on a registration failure → `add slack --replace-config` repairs the tokens.
- docs: `cli.md` + `slack.md` document the flag and the machine-locality of the config credential.

Deliberately **not** included: an "adopt existing app on a new machine" flow (reconstructing onboarding state from appId + fresh tokens). The Request URL is one-time per stable host (Fly/Railway), so the second-machine gap is narrow; the manual console path remains its documented answer.

## Validation

- [x] `npm run lint` (existing suppression warning only)
- [x] `npm run typecheck`
- [x] `npm test` — 981 tests

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -251,6 +251,10 @@ creates a new internal app with `agent_view`, native streams/tasks, and suggeste
 through OAuth, and writes rotating bot credentials + the Signing Secret to the gitignored `.env`. The configuration refresh token stays owner-readable
 under `<state root>/channels/slack/` and is used locally by `dev --tunnel` / `deploy --run` to update the
 Events API URL; it never travels to the host. `--no-onboard` preserves the manual scaffold-only path.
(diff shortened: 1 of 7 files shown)


Sukitly (approved):
lgtm

PR #269, opened 2026-07-23
docs(session-control): record the steer-vs-durability rejection; true up the builder header

kid7st:
## Summary

Two truth fixes from a review of session-control consumers:

- **session-control.md §15**: record *considered and rejected* — replacing the chat channels' queued-turn path with `steer`/`follow_up`. The stateful channels persist turn intents pre-ACK (L1 at-least-once); a steered message lives only in the run's in-memory queue, so adopting steer would trade the durability floor for latency. The existing demand-driven "opt-in follow_up/steer policy" sketch remains the only sanctioned path.
- **session-builder.ts header**: it still promised the serving control plane as a second consumer; Phases 1–3 were built on the invoke pipeline instead. The header now states the reality: chat TUI is the one consumer, the builder is the independently instantiable assembly-seam proof, and control-plane observation deliberately covers invoke-driven runs, not chat sessions.

Comment/docs only; no behavior change.

## Validation

- [x] `npm run typecheck`
- [x] `npm run lint`

### DIFF
--- a/docs/design/session-control.md
+++ b/docs/design/session-control.md
@@ -418,6 +418,14 @@ for message-boundary delivery (enabling an opt-in follow_up/steer policy for mid
 once interactions exist, an interaction responder (e.g. Telegram inline keyboards). The extension
 unlocks those options; it does not mandate them.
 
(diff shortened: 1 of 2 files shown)


Sukitly (approved):
lgtm

PR #270, opened 2026-07-23
feat(spec): add the advisory `retrying` event and surface retry backoff in live previews

kid7st:
## Summary

pi 0.81.1's summarization retries (#266) reach the session plane as `retry_scheduled` but were dropped by the `AgentEvent` projection — up to ~14s of exponential backoff mid-turn read as a hang in every channel's live preview.

**Design decision**: the turn's own renderer must not depend on the opt-in control plane (`sessionControl` defaults off and is a remote-control surface — a layering inversion for preview completeness). The data plane carries it instead, via exactly the extension mechanism SPEC MUST 4/5 anticipates: a **non-terminal advisory event**. The terminal set stays frozen, existing terminal consumers ignore it, and relays pass it through. No existing terminal semantics change; SPEC stays v0.1 with the additive event recorded.

- **SPEC §5 + `agent.ts`**: `{ type: "retrying", attempt, maxAttempts, delayMs, reason }` — advisory, deliberately unclosed (the next event is its closure); `reason` is human-facing prose
- **projection** (`projectAgentEvent`): `retry_scheduled` → `retrying`; `operation` (`compaction | branch_summary`) stays session-plane vocabulary
- **telegram / feishu / slack-classic**: render a shared neutral notice (`preview-kit.RETRY_NOTICE`, no leaked internals), closed by the next progress event
- **slack-native**: DM and channel threads override Agent View's visible `loading_messages` during retry, restore Slack's default working state on progress, serialize status transitions through final clear, and keep DM-only thread titles
- **slack-classic rate safety**: re-render after the three-second mutation wait so a closed retry cannot publish a stale notice over newer progress
- **CLI `invoke-stream` + `observe`**: operator-facing retry lines including the reason
- **docs**: SPEC / API reference / embedding event enumerations and Slack rendering behavior updated

## Validation

- [x] `npm run lint` (existing unused-suppression warning only)
- [x] `npm run typecheck`
- [x] `npm test` — 87 files, 986 tests
- [x] `npm run build`
- [x] Real Slack Agent View smoke: default pending → `Temporary problem — retrying…` → default pending → streamed answer

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -4,7 +4,7 @@ description: "Agent Handler protocol v0.1: one invoke function and a small set o
 type: spec
 status: locked
 version: 0.1
-updated: 2026-06-09
+updated: 2026-07-18
 ---
 
 # Agent Handler — Protocol Specification v0.1 (locked)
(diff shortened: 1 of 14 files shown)


Sukitly (approved):
lgtm

PR #279, opened 2026-07-24
fix(slack)!: stream inline tool traces in native streaming

Sukitly:
## Summary

Slack's native streaming previously hid all tool activity, so multi-tool turns (e.g. six consecutive
`bash` calls) looked like the agent was silent or stuck. An earlier iteration of this PR mapped tool
calls to native `task_update` cards; that approach is reverted here because tool calls are execution
facts, not user-level tasks — mapping them to Task Cards misrepresents semantics and locks rendering
into Slack's fixed Task Card UI.

Final approach: stream compact, factual inline tool traces as plain `markdown_text`.

- render each tool call as a bold tool name with a code-formatted argument summary, no emoji:

  ```
  **Bash** — `npm test`
  **Bash** failed — `npm test`
  ```

- **a policy reversal, stated plainly**: native rendering previously showed no tool arguments at all
  (a Task Card had nowhere to put them). It now shows the call's first primitive argument, and a
  native stream cannot be retracted, so that value persists in the delivered message. The channel
  knows no tool schemas, so it cannot tell a query from a credential — `docs/slack.md` discloses this
  and a test pins it. The other renderers show the same summary, but only in a live preview that
  settles into the answer alone
- reuse the shared 48-code-point argument bound (`summarizeToolArgs`) that telegram, feishu, and the
  Slack compatibility renderer already use, so every channel exposes exactly the same thing
- tool OUTPUT never leaves the process, failed or not: the failure line repeats the operation so the
  call is identifiable, and says nothing about what it printed. Reading that would mean guessing the
  engine's result shape inside `src/channels`
- a successful call is not separately confirmed: an append-only stream cannot flip an already-sent
  line to ✓, and the next trace — or the answer itself — is the progress signal
- preserve ordinary assistant `text` in event order; raw `thinking` stays hidden
- drop the `taskDisplay` config and the `timeline | plan | dense` modes introduced mid-PR, along with
  the whole `chunks` mode of the Slack stream API (no longer applicable without Task Cards)
- record the one real exception in `docs/design/core.md` and `docs/slack.md`: a native stream cannot
  be retracted, so its traces persist beside the answer, while every other renderer settles into the
  answer alone. `rendering: "classic"` remains the answer-alone option

## Breaking change

`taskDisplay` was published in 0.15.0 and only laid out Slack Task Cards, which no longer exist here.
The option is removed outright — no runtime guard, no deprecation window: TypeScript rejects it, and a
stale JavaScript config ignores it.

```diff
 export default slackChannel({
   rendering: "native",
-  taskDisplay: "plan",
 });
```

Tool activity is inline Markdown in the same stream; there is no layout knob to replace it.

Explicit high-level task/plan events (Slack Plan UI) are intentionally out of scope: they require task
data the Agent contract does not yet emit, and must not be inferred from `tool_started`/`tool_ended`.

## Also in this branch

Two backward-compatibility carriers in the Feishu channel, removed under the same rule as
`taskDisplay` — this project does not carry compatibility for published options or for state written
by an older release:

- the `baseUrl` channel option, an `@deprecated` alias kept alive for channel files scaffolded before
  the rename to `apiBaseUrl`;
- `bufferKey`'s optionality on a stored turn. It existed so a record written before context buffering
  could still be recovered, at the cost of an invented `:legacy-turn:` bucket on every restart. The
  field is required now, so a pre-buffer record degrades like any wrong-shape state file: the store
  warns and starts empty, once, visibly.

`parse.ts`'s `parseContent` is *described* as a compatibility shim "for existing helpers/tests" but
sits on the live path (the prompt envelope, and quoted-parent resolution). Only its companion text was
wrong; the projection itself is real and stays.

## Testing

- `npm test` (1003 tests / 89 files)
- `npm run typecheck`
- `npm run lint` (only the pre-existing warning)
- `npm run build`

### DIFF
--- a/src/channels/slack/preview.ts
+++ b/src/channels/slack/preview.ts
@@ -472,7 +472,7 @@ export async function streamSlackReply(
     initialPreviewTs,
     threadTitle,
     disclaimer,
-    taskDisplay = "plan",
+    taskDisplay = "timeline",
     label = "[slack]",
   } = options;
   if (rendering === "native" && target.threadTs) {
--- a/src/channels/slack/preview.ts
+++ b/src/channels/slack/preview.ts
@@ -1,5 +1,5 @@
 /** Slack reply rendering: native Agent streams first, rate-safe edited-message compatibility second. */
-import type { AgentEvent } from "../../agent.ts";
+import type { AgentEvent, Json } from "../../agent.ts";
 import { log } from "../../log.ts";
 import {
   RETRY_NOTICE,
@@ -11,8 +11,10 @@ import {
   defaultErrorMessage,
   humanizeToolName,
   revealedAnswer,
+  summarizeToolArgs,
   toolLines,
 } from "../preview-kit.ts";
+import { truncateCodePointPrefix } from "../text.ts";
 import {
   type SlackApi,
   type SlackStreamChunk,
@@ -31,6 +33,8 @@ const CLASSIC_UPDATE_INTERVAL_MS = 3_000;
 const NATIVE_APPEND_INTERVAL_MS = 750;
 const WORKING_STATUS = "is working on your request…";
 const GENERIC_FAILURE = "⚠️ The response stream stopped unexpectedly. Please try again.";
+/** Slack documents one 256-character budget across a task_update's human-readable text. */
+const NATIVE_TASK_TEXT_MAX = 256;
 
 const wait = (ms: number): Promise<void> => new Promise((resolve) => setTimeout(resolve, ms));
 
@@ -42,6 +46,49 @@ const wait = (ms: number): Promise<void> => new Promise((resolve) => setTimeout(
  * (real Slack controls never contain a `<`). */
 export function sanitizeSlackMarkdown(markdown: string): string {
   return markdown.replace(/<[@!][^<>]*>/g, (control) => `&lt;${control.slice(1)}`);
+}
+
+interface NativeTask {
+  title: string;
+  details?: string;
+}
+
+/** Prefer pi's AgentToolResult text blocks, then common custom-tool error fields, then compact JSON. */
+function toolResultText(value: Json): string {
+  if (typeof value === "string") return value;
+  if (value === null) return "";
+  if (typeof value !== "object" || Array.isArray(value)) return JSON.stringify(value);
+
+  const blocks = value.content;
+  if (Array.isArray(blocks)) {
+    const text = blocks
+      .map((block) =>
+        block && typeof block === "object" && !Array.isArray(block) && typeof block.text === "string" ? block.text : "",
+      )
+      .filter(Boolean)
+      .join("\n");
+    if (text) return text;
+  }
+  for (const key of ["error", "message", "result"] as const) {
+    const field = value[key];
+    if (typeof field === "string" && field.trim()) return field;
+  }
+  return JSON.stringify(value);
+}
+
+function nativeTaskDetails(args: Json): string | undefined {
+  const details = sanitizeSlackMarkdown(summarizeToolArgs(args));
+  return details || undefined;
+}
+
+/** Failed output is the only result content made customer-facing: one line, sanitized, and fitted
+ * into Slack's task text budget after the title and operation details. */
+function nativeTaskErrorOutput(content: Json, task: NativeTask): string | undefined {
+  const used = Array.from(task.title).length + Array.from(task.details ?? "").length;
+  const available = NATIVE_TASK_TEXT_MAX - used;
+  if (available <= 0) return undefined;
+  const raw = toolResultText(content).replace(/\s+/g, " ").trim() || "Tool failed without an error message.";
+  return truncateCodePointPrefix(sanitizeSlackMarkdown(raw), available);
 }
 
 function withDisclaimer(markdown: string, disclaimer: string | false | undefined): string {
@@ -253,7 +300,7 @@ async function streamNativeSlackReply(
         .catch((error) => log.warn(`${label} could not set Slack Agent status: ${String(error)}`)),
     );
   };
-  const toolNames = new Map<string, string>();
+  const toolTasks = new Map<string, NativeTask>();
   let pendingText = "";
   let fullAnswer = "";
   let textTimer: ReturnType<typeof setTimeout> | undefined;
@@ -354,15 +401,21 @@ async function streamNativeSlackReply(
           setStatus("hit a temporary problem — retrying…");
         }
       } else if (event.type === "tool_started") {
-        const title = humanizeToolName(event.name);
-        toolNames.set(event.id, title);
-        sendTask({ type: "task_update", id: event.id, title, status: "in_progress" });
+        const task: NativeTask = {
+          title: humanizeToolName(event.name),
+          details: nativeTaskDetails(event.args),
+        };
+        toolTasks.set(event.id, task);
+        sendTask({ type: "task_update", id: event.id, ...task, status: "in_progress" });
       } else if (event.type === "tool_ended") {
+        const task = toolTasks.get(event.id) ?? { title: "Tool" };
+        const output = event.isError ? nativeTaskErrorOutput(event.content, task) : undefined;
         sendTask({
           type: "task_update",
           id: event.id,
-          title: toolNames.get(event.id) ?? "Tool",
+          ...task,
           status: event.isError ? "error" : "complete",
+          ...(output ? { output } : {}),
         });
       } else if (event.type === "completed") {
         finalized = true;
--- a/src/channels/slack/preview.ts
+++ b/src/channels/slack/preview.ts
@@ -76,14 +76,9 @@ function toolResultText(value: Json): string {
   return JSON.stringify(value);
 }
 
-function nativeTask(name: string, args: Json, displayMode: SlackTaskDisplayMode): NativeTask {
-  const title = humanizeToolName(name);
-  const details = sanitizeSlackMarkdown(summarizeToolArgs(args)) || undefined;
-  // Slack's plan/dense blocks persist only task title + status; details/output are timeline-only in
-  // the rendered message. Keep compact modes useful by folding the operation into their title.
-  return displayMode === "timeline"
-    ? { title, details }
-    : { title: truncateCodePointPrefix([title, details].filter(Boolean).join(" "), NATIVE_TASK_TEXT_MAX) };
+function nativeTaskDetails(args: Json): string | undefined {
+  const details = sanitizeSlackMarkdown(summarizeToolArgs(args));
+  return details || undefined;
 }
 
 /** Failed output is the only result content made customer-facing: one line, sanitized, and fitted
@@ -406,7 +401,10 @@ async function streamNativeSlackReply(
           setStatus("hit a temporary problem — retrying…");
         }
       } else if (event.type === "tool_started") {
-        const task = nativeTask(event.name, event.args, taskDisplayMode);
+        const task: NativeTask = {
+          title: humanizeToolName(event.name),
+          details: nativeTaskDetails(event.args),
+        };
         toolTasks.set(event.id, task);
         sendTask({ type: "task_update", id: event.id, ...task, status: "in_progress" });
       } else if (event.type === "tool_ended") {
@@ -474,7 +472,7 @@ export async function streamSlackReply(
     initialPreviewTs,
     threadTitle,
     disclaimer,
-    taskDisplay = "timeline",
+    taskDisplay = "plan",
     label = "[slack]",
   } = options;
   if (rendering === "native" && target.threadTs) {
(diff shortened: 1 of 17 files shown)


src/channels/slack/preview.ts
```diff
@@ -44,6 +43,45 @@ export function sanitizeSlackMarkdown(markdown: string): string {
...
+
+function nativeToolInvocation(trace: NativeToolTrace): string {
+  return `${boldText(trace.label)}${trace.operation ? ` — ${inlineCode(trace.operation)}` : ""}`;
+}
+
+/** Tool OUTPUT never leaves the process, failed or not: the channel would have to guess the engine's
+ * result shape to read it, and the agent already explains a failure it recovered from in its answer.
+ * The trace states the fact only — which call failed. Operators get the detail from the logs. */
+function nativeToolFailure(trace: NativeToolTrace): string {
+  return `${boldText(trace.label)} failed`;
```
kid7st: `nativeToolFailure` discards `trace.operation`, so the failure line is `**Bash** failed` — but `docs/slack.md:223` says the errored `tool_ended` "appends one line naming the failed call". It names the *tool*, not the call. With the common case (several `Bash`/`Read` calls in one turn, one of them failing) the user cannot tell which of the identical `**Bash** — \`…\`` lines above it the failure refers to; `toolTraces` is holding `operation` and then throwing it away exactly where it would disambiguate.

Direction: render `**Bash** failed — \`npm test\`` (reuse `nativeToolInvocation`'s operation, which is already sanitized and bounded), or drop the claim from the docs and accept the line is tool-level only.

src/channels/slack/preview.ts
```diff
@@ -354,16 +398,13 @@ async function streamNativeSlackReply(
...
       } else if (event.type === "tool_ended") {
-        sendTask({
-          type: "task_update",
-          id: event.id,
-          title: toolNames.get(event.id) ?? "Tool",
-          status: event.isError ? "error" : "complete",
-        });
+        const trace = toolTraces.get(event.id) ?? { label: "Tool" };
+        toolTraces.delete(event.id);
+        if (event.isError) sendToolTrace(nativeToolFailure(trace));
```
kid7st: The justification comment above `nativeToolFailure` argues against the line it emits: "the agent already explains a failure it recovered from in its answer" is the reason given for hiding output, and it applies with equal force to the failure line itself. `isError` covers routine, agent-handled tool errors (grep with no match, a file that doesn't exist, a probe command) — each of those now leaves a permanent, unretractable `**X** failed` in the delivered message of a turn that succeeded.

This is currently the worst of both positions: alarming enough to worry a user, too thin to act on. Pick one — either the failure fact is user-relevant (then it needs the call and a reason, see the comment on `nativeToolFailure`), or it isn't (then treat it like success and emit nothing, leaving the answer to explain).

src/channels/slack/preview.ts
```diff
@@ -44,6 +43,45 @@ export function sanitizeSlackMarkdown(markdown: string): string {
...
+  return {
+    label: sanitizeSlackMarkdown(humanizeToolName(name)),
+    ...(operation ? { operation } : {}),
+  };
+}
+
+/** Bold a factual label. The escape set is `\` and `*` only because `humanizeToolName` already
+ * normalizes `_`/`-`/`.` separators to spaces, so no underscore emphasis can reach here. */
+function boldText(value: string): string {
+  return `**${value.replace(/[\\*]/g, "\\$&")}**`;
```
kid7st: The comment states an invariant `humanizeToolName` does not guarantee: it normalizes `_`/`-`/`.` only on the success path, and falls back to `name.trim()` when normalization empties the string (`preview-kit.ts:255`, `base.trim() || name.trim() || "Tool"`). A tool named `_` or `__` therefore reaches `boldText` with underscores intact, and the escape set (`\\`, `*`) does not cover them.

Direction: either escape `_` too (one character in the class, no cost) or replace the comment with what is actually true — code that contradicts its own comment is worse than either.

src/channels/slack/preview.ts
```diff
@@ -309,9 +352,10 @@ async function streamNativeSlackReply(
       flushText();
     }, delay);
   };
-  const sendTask = (chunk: SlackStreamChunk): void => {
+  const sendToolTrace = (line: string): void => {
     flushText();
-    enqueue(() => sendContent({ chunks: [chunk] }));
+    const separator = streamHasContent && !streamEndsWithBlankLine ? "\n\n" : "";
```
kid7st, line 364: This separator branch — the only new stateful logic in the renderer (`streamHasContent` / `streamEndsWithBlankLine`) — is never exercised. Both native-trace tests (`test/slack-preview.test.ts:90`, `test/slack.test.ts:360`) emit `tool_started` before any `text`, so `streamHasContent` is `false` at the first trace and the previous trace already ends with `\n\n` at the second: `separator` is `""` in every assertion in the suite. A regression that dropped the separator (trace glued onto the tail of a streamed sentence) would pass the whole suite.

Direction: one ordering case — `text` → `tool_started` → `text` — asserting the appended trace begins with a blank line.

src/channels/slack/preview.ts
```diff
@@ -309,9 +352,10 @@ async function streamNativeSlackReply(
...
+    queueMarkdown(`${separator}${line}\n\n`);
```
kid7st, line 365: Moving traces from `task_update` chunks into the markdown text stream puts them inside the answer's Markdown structure, which the previous chunk type kept them out of. The stream is one document assembled from appends: if the agent opens a fenced code block, calls a tool, then continues, the trace is emitted *inside* the open fence — and `inlineCode`'s backticks can close it early, corrupting the rendering of everything after. This repo already treats fence balancing as real (`chunkSlackMarkdown`: "Split standard Markdown while balancing fenced code blocks"); the trace path has no equivalent.

Direction: track fence parity over the appended text (the same state `chunkSlackMarkdown` computes) and close/reopen around a trace, or state the limitation next to `sendToolTrace` as a known ceiling.

test/slack.test.ts
```diff
@@ -353,33 +346,21 @@ describe("Slack sessions, context, and managed threads", () => {
...
     const fetchMock = okFetch();
     vi.stubGlobal("fetch", fetchMock);
     let prompt: Prompt | undefined;
     const agent: Agent = {
       async *invoke(_scope, value): AsyncIterable<AgentEvent> {
         prompt = value;
         yield { type: "thinking", delta: "private chain of thought: launch-code" };
-        yield { type: "tool_started", id: "t1", name: "search", args: { token: "tool-secret" } };
-        yield { type: "tool_ended", id: "t1", isError: false, content: { result: "internal" } };
+        yield { type: "tool_started", id: "t1", name: "search", args: { query: "public docs" } };
```
kid7st, line 362: This fixture change inverts what the test guarded. It was the one assertion in the suite that a tool argument (`token: "tool-secret"`) never reaches Slack; it now asserts the argument does reach Slack. Nothing wrong with changing the policy, but the test name still reads as a safety test ("without exposing reasoning or successful tool output") while it is now half an exposure test.

Direction: keep a fail-case with a credential-shaped argument so the new boundary is stated as a spec — the arg IS emitted, bounded to 48 code points and sanitized — instead of the suite silently losing the only leak guard it had.

kid7st (commented):
Thesis: tool lifecycle events stop being Slack Task Cards and become inline Markdown "tool traces" appended into the same native stream as the answer — the concept moves from "a task with a status that can be updated" to "an append-only factual line", and the whole `chunks` mode plus `taskDisplay` disappear with it.

Findings that don't anchor to a single diff line:

**Silent decision: tool arguments become permanently customer-facing.** Before this PR, `test/slack.test.ts` used `args: { token: "tool-secret" }` and asserted `not.toContain("tool-secret")` — "generic tool arguments are never customer-facing" was the encoded policy (old docs/slack.md said exactly that). The PR replaces the fixture with `{ query: "public docs" }` and asserts the argument IS emitted. The behavioural delta is not "tool traces are now visible", it is "the first primitive field of every tool call is now permanently visible in a message that cannot be retracted". `docs/slack.md` discloses it in a subordinate clause; the PR summary lists it as "reuse the shared 48-code-point argument bound", which reads as a bound change, not a policy reversal. Direction: state the reversal in the PR body / commit message as its own bullet so the maintainer approves the policy, not the constant.

**Two axes collapsed onto `rendering`.** `docs/slack.md` now instructs an author who does not want persisted process traces to choose `rendering: "classic"` — which also gives up native streaming, Agent status semantics, and the append transport. Transport (native stream vs one edited message) and process visibility (traces vs answer-only) are independent concepts; the PR removed the only native-side render knob (`taskDisplay`) and left the answer-only choice reachable only by changing transport. Either say plainly in the docs that answer-only is not available under native rendering, or keep the visibility choice on its own option.

src/channels/slack/preview.ts
kid7st: Fixed in 8556f3d: the failure line repeats the operation (`**Bash** failed — \`npm test\``), so repeated calls to one tool stay distinguishable, and it exposes nothing the start line did not already show. Both lines now come from one `nativeToolLine(trace, outcome)`.

src/channels/slack/preview.ts
kid7st: Kept the line, with the call attached (8556f3d). Every other renderer marks a failed call (`🔧 Bash npm test ✗`), so emitting nothing would make native the only place where a failure is invisible. The comment's argument is about tool OUTPUT, not about the fact of failure — reworded to say that.

src/channels/slack/preview.ts
kid7st: Right, `humanizeToolName` falls back to `name.trim()` when normalization empties the label. Escaped `_` alongside `\\` and `*` in 8556f3d rather than keeping an assumption the callee does not guarantee.

src/channels/slack/preview.ts
kid7st: Covered in 8556f3d with an exact-string case: `text → tool_started → text` must render `Checking.\n\n**Read** — \`AGENTS.md\`\n\nDone.`. Dropping the separator now fails.

src/channels/slack/preview.ts
kid7st: Recorded the ceiling next to `sendToolTrace` (8556f3d) instead of tracking fence parity: it needs a model that opens a `\`\`\`` fence and calls a tool before closing it. The note names `chunkSlackMarkdown` as the fix if that is ever observed.

test/slack.test.ts
kid7st: Restored as a spec rather than a guard (8556f3d): a new test asserts `{ token: "sk-live-secret", env: "prod" }` renders as `**Deploy** — \`sk-live-secret\``, with a comment saying this is the disclosed boundary — a channel knows no tool schemas. Changing which field is picked now breaks a test. `docs/slack.md` and the PR summary state the reversal explicitly.

kid7st (approved):
Approving the design call and the shipped behavior. **Disclosure: five of the ten commits on this branch are mine** (the `review*` follow-ups), so this approval covers the author's original decision and the audit trail rather than being an independent read of every line — a second maintainer skim is cheap insurance.

**The decision that matters** — reverting Task Cards. A `task_update` is a user-level task with a mutable status; a tool call is an execution fact. Mapping one to the other misreports semantics and hands rendering to a fixed Slack UI. Inline traces are the honest shape, and they cost less API surface: `chunks` mode, `task_display_mode`, and the `SlackStreamContent` wrapper are all gone.

**What the two audit passes changed:**

| | before | after |
|---|---|---|
| argument bound | Slack-only 96 cp via a new shared-code parameter | the shared 48 cp every channel already uses (`preview-kit` diff is one `export`) |
| failure line | 256 cp of raw tool output, read by guessing the engine's result shape | `**Bash** failed — \`npm test\`` — which call, never what it printed |
| removed `taskDisplay` | silently ignored | throws on a supplied value, with a docs removal note |
| escaping / separator logic | untested | pinned, including a tool named `_` and backtick-edged arguments |
| the disclosure | a clause in `docs/slack.md` | stated in the `rendering` docstring, the scaffold comment, `preview-kit`, and the PR summary |

**Declined, on the record:** a `toolTrace: "operation" | "name" | "off"` knob. This PR deleted one speculative render knob; adding another for a use case nobody has reported repeats it. The option stays open.

**Known ceilings**, both recorded in code/docs rather than fixed: a trace appended while the answer has an unclosed fence can land inside it (needs a model that calls a tool mid-fence), and a successful call gets no completion mark, because an append-only stream cannot flip an emitted line.

**Merge order — this PR first, then #282.** They conflict in `docs/design/core.md`, `src/channels/slack/slack.ts`, and `test/slack.test.ts`; #282 still carries five `taskDisplay` references that must go, and its `slackChannel(options)` restructure now matches this branch's, so resolving in this order is mechanical.

CI green (10/10). Merging remains a maintainer decision.

PR #280, opened 2026-07-24
refactor(placement)!: one marker, and the workspace is the directory you point at

Sukitly:
## What changed

Placement had **two markers**: a nested agent was the directory NAME `fastagent` plus a shallow evidence probe; a flat agent was its `fastagent.config.*`. The name being load-bearing forced that asymmetry — and it made the agent directory unrenameable.

This PR replaces the pair with **one marker and one rule**:

> `fastagent.config.*` declares an agent. The **workspace is the directory you point fastagent at**. The agent is the single config holder at it, or one level inside.

```
findPlacement(dir):
  dir 有 config                → { agentDir: dir,   workspace: dir }
  exactly one child has config → { agentDir: child, workspace: dir }
  several do                   → refuse, naming them
  none                         → not an agent
```

### Why the invariant had to go, not the marker

Custom directory names cannot coexist with the old rule, because **"a config-bearing directory inside a config-less directory" is literally both shapes at once**. Without a reserved name, nothing in the tree says whether such an agent's workspace is itself or its parent.

What demanded an answer was an invariant — "an agent has ONE true workspace regardless of where you stand" — and that is the constraint worth dropping, because *it is not even true where it matters most*: on a box shipped nothing but the agent directory, insisting on the parent hands the agent `/`.

So the same tree answers two ways by aim, and both answers are honest:

```
repo/            # `fastagent dev` here → agent = repo/agent, workspace = repo
├── AGENTS.md
├── src/
└── agent/       # `fastagent dev` here → agent = repo/agent, workspace = repo/agent
    ├── fastagent.config.mjs        ← THE marker (any directory name)
    ├── persona.md  skills/  tools/  channels/  schedules/
    └── .secrets/  .state/
```

Point at the project and its agent serves with the project as its workspace (what `init` sets up). Point at the agent directory and it works on itself — the "flat" case (`--flat`) is no longer a mode, just the one rule with the two directories equal.

## Deleted, because the rule no longer needs them

| Gone | Why it existed |
|---|---|
| `AGENT_SURFACE` evidence probe | the name alone was too weak a marker |
| `isNamedAgentDir` / `isFlatAgentDir` | two markers needed two predicates |
| the entry-point-invariance candidate | only knowable from a fixed name |
| the reserved-name refusal + its dead end | `fastagent/` was a trap only because the name resolved |
| `isNestedAgentDir` | re-derived from a name what `agentDir !== workspace` states |
| `--flat` as a placement MODE | now an alias for `--agent-dir .`, which only says where files land |
| `AGENT_DIR` as a rule | survives as `DEFAULT_AGENT_DIRNAME` — what `init` names a directory |

`src/paths.ts` resolution goes from 3 ordered candidates + 2 predicates + a probe list + reserved-name handling, to **2 branches and one shallow scan**.

## Added

- **`--agent-dir <name>`** — names the agent directory anything (this is the flag `234ff97` retired, restored with its original meaning, its `conflicts: ["agentDir"]` against `--flat`, and its exit-2 usage class).
- **`FASTAGENT_AGENT`** — several agents can share ONE workspace, which is the shape a repository with an engineer's, a PM's and a content owner's agent needs. The one-level scan already found them; nothing could say which was meant. Selection is per-PERSON, which decides where it lives: a committed workspace file is shared by construction and cannot express "mine", so it is the environment — and the per-repo, per-person file this wants already exists and is not ours to invent (`.envrc`). A workspace REGISTRY was considered and rejected twice over: the scan already answers "which agents are here", so a list could only drift from it, and it would centralize what the scan decentralizes. It **asserts** (a directory without that agent resolves to nothing, at any count — serving a different agent than the one asked for is the silent wrong-target refused everywhere else here), the agent named `fastagent` breaks the tie, and `deploy` bakes the choice into the image so the container never re-picks. No per-command flag: placement has one resolution point, so all 15 commands got the capability without touching any of them.
- A **`hint:`** line under `agent:`/`workspace:` when you pointed at the agent and its parent carries an `AGENTS.md` or a `.git`. A hint may use that heuristic precisely because a *rule* may not.

## Also in this PR (the earlier, unchanged work)

- **`.secrets/` and `.state/` split by deploy lifecycle** — secrets ride the host's secret store, state rides a volume; `FASTAGENT_SECRETS_DIR` / `FASTAGENT_STATE_DIR` point both at it in a container.
- **No runtime git opinions.** `init` scaffolds two ignore files and no command ever reads, verifies or rewrites one. The single exception is one-directional: the directory fastagent writes secrets into carries its own `.gitignore` (`wx`, never over an existing file) — the risk is asymmetric (one unwanted file vs a published credential).
- **Ownership decides deploy overwrites**, not `--force` and not the path: every generated artifact opens with a marker, unmarked files are never touched, delete one to hand the path back.
- Placement moved from `engines/pi/config.ts` to the neutral **`src/paths.ts`** (the engine has no say in it), with `placementOrExit` replacing 16 hand-written try/catches.

## Breaking (pre-1.0, no migration layer)

1. **There is no zero-config agent.** The config may be `export default {}`, but the file must exist — a hand-made markdown-only agent (or one whose config was deleted) stops resolving. `init` always writes one.
2. **`cd agent && fastagent dev` narrows the WORKSPACE** to the agent's own directory: its cwd, its coding tools' root, deploy's build context. (② context is unaffected — that walk climbs ancestors either way.) Printed on every run, plus the `hint:` line.
3. **A workspace holding several agents** resolves only through `FASTAGENT_AGENT` or the `fastagent` default name; without either, every command refuses and names them.
4. `.fastagentignore` is gone; the credential path moved under `.secrets/`.

## Verification

`npm run lint` + `npm run typecheck` clean, **1054 tests**. Behavior smoke-tested end to end (custom name resolving from both ends, the hint, both refusals). One review pass (`rv.sh local`) run and all five findings addressed in `8e3b3f1`.


Sukitly:
## No-backward-compatibility ledger

Maintainer decision on this PR: the layout break ships with **no migration and no legacy detection**. This comment records every known way a workspace or deployment created by fastagent ≤ 0.15 fails under the new layout, so the decision can be re-audited and selectively reversed later. Where a row reads "loud", that is a side effect of a fail-visibly guard added on this PR, not compatibility code.

| # | What moved | Old | New | How a ≤0.15 workspace fails |
|---|---|---|---|---|
| 1 | Machine-state root | `<root>/.fastagent/{sessions,channels/<kind>,schedule}` | `<root>/.state/…` (`FASTAGENT_STATE_DIR`) | Usually **loud by accident**: an old state dir carrying `channels/` trips the config-less-workspace guard in `resolveWorkspace` — wrong story ("looks like a workspace… run `fastagent init`"), right remedy (move it away). An old dir with only `sessions/` + `auth.json` resolves **silently** → fresh empty state (conversations + Telegram turn-replay state reset). |
| 2 | Project credentials | `<root>/.fastagent/auth.json` | `<root>/.secrets/auth.json` | Quiet: reads as not-configured; the startup auth report names login as the remedy. OAuth users re-login (the old rotated token is ignored, never consumed). |
| 3 | Workspace `.env` | `<root>/.env` | `<resolved secrets dir>/.env` (default `<root>/.secrets/.env`) | **Silent**: the old root `.env` is never read → missing-secret / missing-model errors with the values sitting one directory away. |
| 4 | Global credentials | `~/.fastagent/auth.json` | `~/.fastagent/.secrets/auth.json` | **Silent**: the SDK's dir-less rungs and `login` from `$HOME` see not-configured; re-login. Explicit `FASTAGENT_AUTH_PATH` users are unaffected. |
| 5 | `config.agentDir` | supported (definition in a subdirectory) | removed — the embedded layout replaces it | Loud but generic: `unknown key "agentDir"` with no migration pointer (decided against a special-cased error). |
| 6 | `.fastagent/` gitignore convention | state dir — correctly gitignored by hosts (this repo included) | embedded workspace root — must be committable | **Silent**: a host `.gitignore` still carrying `.fastagent/` un-versions the whole new workspace; `init` reports success. Decided NOT to detect (`init` stays git-agnostic; the affected rule is the operator's own file). |
| 7 | Generated `.dockerignore` | carried `**/.fastagent` (excluding the old state dir) | must NOT exclude `.fastagent` (embedded ships it) | Loud on the embedded path: preflight gates `--run` / warns when a kept host `.dockerignore` still matches `.fastagent`. **Residual silent hole**: a flat workspace with a leftover old state dir that dodges row 1's guard (`sessions/` + `auth.json` only) gets that dir — auth.json included — baked into the image, since the new template no longer excludes it. |
| 8 | Generated host artifacts (fly.toml / compose / railway variables) | `FASTAGENT_STATE_DIR=/data` only — auth lived under the state root (`/data/auth.json`) | `/data/.state` + `FASTAGENT_SECRETS_DIR=/data/.secrets` | **Silent** on a deployed box that upgrades fastagent while keeping old artifacts (deploy never clobbers without `--force`): state still resolves fine (`/data`), but auth now resolves to the in-image `<root>/.secrets/` — the volume's rotated `/data/auth.json` is ignored, and the stale `FASTAGENT_AUTH_SEED` re-seeds that ephemeral path on every boot, breaking single-use OAuth refresh chains. Remedy: regenerate artifacts (`deploy --force`) or set `FASTAGENT_SECRETS_DIR=/data/.secrets` on the host, then re-login/re-seed once. |

Re-audit rule: when a support report matches a row above, point here first. If one class recurs, the candidate fix is a **one-shot detector** for that row (a startup message naming the move and the `mv`), never a compatibility shim — detection preserves the clean layout; a shim re-couples us to the old one.

kid7st:
Pushed `620fce5` — the review findings are fixed on this branch, no follow-ups.

| # | Fix | Where |
|---|---|---|
| 1 | `init --flat` into a directory named `fastagent` is now refused — any directory with that basename resolves as a *nested* root, so the workspace would have served with the PARENT as the workbench (the agent's cwd, its ② context, and deploy's build context, all one level out). Nesting into it stays legal. | `scaffold/init.ts`, `docs/cli.md`, `docs/ai-start.md` |
| 2 | `package.json` dropped from the config-less workspace signal. It was the one name in that list carrying no fastagent meaning, so an unrelated `fastagent/` (a vendored clone, a monorepo package) hard-blocked every command in its parent; the five definition names still catch a workspace that lost its config. | `engines/pi/config.ts` |
| 3 | The two **silent** ledger rows now speak once, at the moment the old location stops mattering: a pre-0.16 root `.env` (named when the new path has none) and an orphaned `.fastagent/` state dir (named *before* the new state root is created). Both point at the new path; neither reads nor moves anything. | `env.ts`, `engines/pi/workspace.ts` |
| 4 | `~/.fastagent/auth.json` → `~/.fastagent/.secrets/auth.json` in the fly/railway **runbooks** (user-facing — the old path no longer exists) and two docstrings; `resolveAuthPathOverride` now names `resolveSecretsDir` instead of the state root. | `deploy/{fly,railway}/plan.ts`, `engines/pi/{create,config}.ts` |
| 5 | deploy composes artifact paths from `WORKSPACE_DIR` instead of eight hardcoded `fastagent/` literals (the rename that just happened is the argument); `an nested` ×4; `info` writes through one padded-label helper. | `deploy/*`, `cli/commands/info.ts` |

**Ledger correction (row 1).** That comment predates the final `.fastagent` → `fastagent` rename, so its "loud by accident" no longer holds: `.fastagent` is no longer the workspace-directory name, so an old state dir does not trip the config-less guard — it was fully silent (sessions restart empty, channel dedup state resets → possible re-delivery). Fix #3 is what makes that row loud again, by naming it rather than by reading the old directory.

Tests: `lint` + `typecheck` clean, **1015/1015** pass (+4 covering the reserved name, the `package.json` non-signal, and both legacy notices).

kid7st:
Pushed `9e05f54` — **reverting fix #3 from the previous commit.** Maintainer call, and the right one: the two "legacy location" notices were back-compat code in a fail-visibly costume. They read nothing and moved nothing, but they were still two code paths, five imports and two tests existing only for a layout this branch deletes — and `workspace.ts` had to pull in `isUnderDir` + `GLOBAL_HOME_DIR` purely to keep the compatibility check itself correct under `$HOME`.

`env.ts`, `engines/pi/workspace.ts` and `test/env.test.ts` are now byte-identical to `3c522f2`. The no-back-compat decision here is "the old shape is gone", not "migrate quietly" — pre-1.0 with no install base, starting fresh IS the upgrade, and a signpost for it just makes every future reader ask what `.fastagent` was. My earlier appeal to fail-visibly was a misread: that rule governs the failures of the system as it is, not archaeology for the one it replaces. The ledger comment stays the record of what breaks.

What remains from the review (all about today's rules being right at their edges, none of it compatibility):

- reserved `fastagent` directory name → `--flat` into one is refused
- `package.json` dropped from the config-less workspace signal
- `~/.fastagent/.secrets/auth.json` in the fly/railway runbooks + two docstrings
- `WORKSPACE_DIR` instead of eight hardcoded `fastagent/` literals; `an nested` ×4; `info`'s padded-label report

`lint` + `typecheck` clean, **1013/1013** pass.

kid7st:
Pushed `a62ee16` — the terminology reversal, as scoped. **77 files, +701/−696, zero behavior change, 1013/1013 pass, CI green.**

## What changed

| | before | after |
|---|---|---|
| `<dir>/fastagent/` | `workspace` / `root` | **`agentDir`** |
| `<dir>` | `workbench` | **`workspace`** |

```
agent:     /host/fastagent      # definition, deps, machinery
workspace: /host                # its cwd; AGENTS.md here is context
```

The reason it was worth doing now: everywhere else in a developer's world — VS Code, IDEs, CI — a *workspace is where you work*. Calling the definition directory `workspace` and inventing `workbench` for the place the agent actually operates had the two nouns backwards, and taxed every reader with a new word for something they already had a name for. git settled this shape long ago: the repository sits at the root of the **working tree**, and the word for *work* belongs to the tree, never to the tool's own directory.

## It fixed a real bug

The scaffolded `persona.md` — the agent's own identity file — told every new agent that its tools *"operate at the workspace root you were started in"*. They don't; they operate in the workbench. We were writing a false statement about the agent's own location into the one file it is guaranteed to read every turn. Under the corrected nouns that sentence is simply true. `fastagent.config.mjs`'s template no longer greets first-time users with the word `workbench` either.

## Identifiers

`WORKSPACE_DIR`→`AGENT_DIR`, `WORKSPACE_CONFIG_NAMES`→`AGENT_CONFIG_NAMES`, `resolveWorkspace`→`resolvePlacement`, `ResolvedWorkspace{root,workbench}`→`ResolvedPlacement{agentDir,workspace}`, `WorkspaceAssembly`→`AgentAssembly`, `resolveWorkspaceTools`→`resolveAgentTools`, `assertInsideWorkspace`→`assertInsideAgentDir`, `src/workspace.ts`→`src/paths.ts`.

`createPiAgentFromWorkspace` **kept its name** — under the new nouns it got *more* accurate (the dir you hand it is the workspace), so the public API surface is untouched.

## Guard rails

- **`agentDir` is not the deleted `config.agentDir` key returning.** Placement stays structural and unconfigurable; this is only the internal name for the path resolution produces.
- **Slack's own "workspace"** (the platform's term) is untouched — zero diff under `src/channels/slack/`, `add-slack.ts`, `docs/slack.md`.
- **`docs/SPEC.md` never used either word** — the locked v0.1 contract is unaffected.
- Method: type definitions were reversed first and `tsc` drove every call site; the 1013 tests (255 of which touch these words) are the no-behavior-change proof.

kid7st:
**The description above is rewritten and is the current record** — squash-merge makes it the commit message on `main`, so three points where it supersedes the earlier text (and the comments below) are worth naming:

1. **`--flat` is back.** The branch deleted it, then restored it — deliberately, not by revert. What made the old two-placement design expensive was *how* the placements were told apart (config-as-structural-marker + a both-roots ambiguity refusal + a config-less probe); that part is redesigned. Nested is marked by the directory NAME, flat by its `fastagent.config.*`, and an overlap resolves nested by ORDER instead of being refused. Consequence to know: **there is no zero-config flat agent.**
2. **The no-back-compat ledger comment is written against superseded behavior.** Its rows still describe the runtime `.gitignore` machinery (since deleted) and the pre-`--flat` resolution. The breaking list in the description is the accurate one; the ledger's re-audit rule still stands, but read the rows as history.
3. **Two removals produce no message at all**, recorded here because they are what a support report will surface first: the undocumented `.fastagentignore` extension point (its only reader was the ignore matcher this change deleted), and the credential-path move (`~/.fastagent/auth.json` → `~/.fastagent/.secrets/auth.json`), which reads as not-configured and re-logs in. Both belong in the release notes as breaking.

PR #281, opened 2026-07-25
fix(feishu): derive managed threads from the thread's own root, not local state

kid7st:
## Problem

A managed group thread — one whose bare (un-mentioned) messages answer in the root session — was recognized only through `owned-threads.json`, written when *this process* reply-threaded a top-level summon. Losing the state disk (redeploy without a volume, a new machine) silently demoted every existing thread to @-mention-only, with no signal anywhere.

## Change

Management is a property of the thread itself: **its root message is a human summon of this bot.** The default route now derives that from the platform and keeps local state only as a cache.

- **Authoritative predicate** (`checkRootSummons`): on a cache miss, `GET /im/v1/messages/:root_id` and check `sender_type === "user"` + a mention of the bot's open_id + `chat_id` matching the event's chat. The GET response's `sender`/`mentions` wire shapes differ from the event push (string id + `id_type` discriminator); `fetchedMessageSummons` in `parse.ts` handles them, with a shape-drift warning if mentions become unreadable.
- **Positive cache**: `owned-threads.ts` → `managed-roots.ts`, now a write-through cache (on-disk name kept for state compat). A lost disk costs one re-check per thread; a failed cache write is a warning, never a failed delivery.
- **Negative cache**: in-process, `chat:root`-keyed, capped at 10k. Only *definitive* platform rejections are cached (recalled root, missing permission — including error codes carried on HTTP 200, classified by the new `isFeishuApiRejection`).
- **Transient failures** (network / 5xx / rate limit / timeout) fail the delivery (HTTP 500 / WS 500 frame) so the platform re-pushes and the next attempt re-checks — a genuine ask is never silently downgraded to background context. Trade-off documented in `docs/design/core.md`.
- **Pre-ACK budget**: identity + root reads share one 2s deadline (`ACK_CHECK_BUDGET_MS`); on expiry the underlying request is aborted (`getMessage` now accepts an `AbortSignal`).
- **Concurrency**: duplicate deliveries of one message share a single acceptance outcome; concurrent bare messages in one thread share a single root read; `seq` is assigned at arrival so recovery order is unaffected.

Known boundary (documented): a thread whose root was *recalled* cannot prove itself after cache loss and falls back to @-only.

## Testing

- 16 new/updated cases across `feishu.test.ts`, `feishu-api.test.ts`, `feishu-parse.test.ts`, `feishu-managed-roots.test.ts`: state-loss recovery, human/cross-chat/empty/recalled roots, transient re-push, both deadline branches, startup identity race, failed identity, duplicate-delivery sharing, rejection classifier table, wire-shape pinning.
- `npm run lint && npm run typecheck && npm test` green locally (89 files, 1016 tests).


src/channels/feishu/feishu.ts
```diff
@@ -442,29 +453,195 @@ function createFeishuRuntimeFactory(
...
+        // The platform answered OK but carried no message — an anomaly worth a visible line, not a
+        // silent branch. Unprovable → unmanaged, cached like a rejection.
+        log.warn(`${label} thread root ${rootId} read succeeded but returned no message — treating as unmanaged`);
+        rememberForeign(rootKey);
+        return false;
+      }
+      if (root.chat_id !== undefined && root.chat_id !== chatId) {
+        // The event claims a chat the root does not live in — an anomaly, and unmanaged FOR THAT CHAT.
+        // Cached under the chat-scoped key, so the mismatch cannot amplify into per-message reads.
+        log.warn(`${label} thread root ${rootId} lives in chat ${root.chat_id}, not ${chatId} — unmanaged`);
```
kid7st: The chat-scope check is fail-OPEN on an absent field (`root.chat_id !== undefined &&`), while the sibling anomalies in the same function (no message returned; unreadable mentions) are fail-closed-with-a-warning. If `chat_id` ever stops arriving in this response shape, chat scoping silently disappears and a root proves management for any chat that claims it — the exact silent-drift failure the mentions guard 13 lines below exists to prevent.

Direction: treat a missing `chat_id` like the `root === undefined` branch — unprovable → warn + unmanaged + negative-cache — so the scoping invariant fails visibly rather than evaporating.

src/channels/feishu/parse.ts
```diff
@@ -80,6 +80,26 @@ export function mentionsBot(message: Pick<FeishuMessage, "mentions">, botOpenId:
...
+ */
+export function fetchedMessageSummons(
+  message: { sender?: unknown; mentions?: unknown[] },
+  botOpenId: string | undefined,
+): boolean {
+  if (!botOpenId) return false;
+  if ((message.sender as { sender_type?: unknown } | undefined)?.sender_type !== "user") return false;
+  return (message.mentions ?? []).some((raw) => {
+    const mention = raw as { id?: unknown; id_type?: unknown };
+    return mention.id_type === "open_id" && mention.id === botOpenId;
```
kid7st, line 99: The predicate hard-requires `id_type === "open_id"`, but nothing in the request enforces that: `getMessage` sends no `user_id_type`, so the assumption rests on the platform default. The drift guard in `feishu.ts` only catches mentions with NO `id_type` — the more likely drift (`id_type` present but `union_id`/`user_id`) passes the guard, matches nothing here, and silently classifies every root as non-summoning: every managed thread degrades to @-only with zero signal.

Direction: pin the assumption at the request (`?user_id_type=open_id` on the message GET) so the response shape is guaranteed rather than hoped for; then the guard's "no id_type" case is the only remaining drift.

kid7st (commented):
Thesis: "managed thread" stops being a fact this process recorded and becomes a property of the thread's own root (root message is a human summon of this bot), platform-derivable; `owned-threads.json` demotes from source of truth to write-through cache, and acceptance becomes async so the derivation can happen pre-ACK.

**`src/channels/feishu/feishu-api.ts:345` (unchanged line, now on a new hot path) — the 429 retry loop is incompatible with the 2s pre-ACK budget.** `call` retries rate limiting up to `RETRIES=3` with `attempt * 1000` sleeps (up to 6s), so a rate-limited root read can never finish inside `ACK_CHECK_BUDGET_MS`: it always times out → abort → delivery fails → platform re-push → another root read. Under rate limiting the new read amplifies itself instead of shedding load, and the deadline's own comment ("must never ride the API pipeline's full … retry budget") is only half true — the abort caps it, but every rate-limited delivery still pays the full 2s and then fails. Direction: for the deadline-bounded pre-ACK read, skip the 429 retry loop (fail fast as transient) — e.g. an explicit `retry: false` alongside `signal` in the `getMessage` opts.

**`docs/feishu.md` was not updated with the semantics change (missing change; `docs/design/core.md` was).**
- `docs/feishu.md:330` still states `owned-threads.json` is "durable `root_id → chat_id` ownership for managed group threads, written before the transport ACK so restarts preserve continuation routing" — after this PR the file is a cache, its write no longer gates the ACK (a failed write is a warning), and restarts recover by re-reading the root. The line now describes the deleted module.
- `docs/feishu.md:236` likewise says continuations are admitted from "its durable managed-thread index".
- `docs/feishu.md:120-123` enumerates the scopes onboarding requests; none of them is the message-READ scope that `GET /im/v1/messages/:root_id` needs. The authoritative predicate now depends on that scope: without it every root read is a definitive rejection, every thread is negative-cached as unmanaged, and the operator gets exactly one WARN line for the whole process. If the scope is implied by the `PersonalAgent` archetype, say so; if not, it belongs in the list and in the startup capability report next to `im:message.group_msg`.

src/channels/feishu/feishu.ts
```diff
@@ -442,29 +453,195 @@ function createFeishuRuntimeFactory(
...
+          // re-push → re-check), because ACKing here would silently downgrade a GENUINE ask to
+          // background context. If every bounded re-push fails, both the turn and the buffered copy
+          // are lost (operator log only) — accepted; see docs/design/core.md.
+          throw new Error(`${label} could not read thread root ${rootId} pre-ACK: ${String(error)}`, {
+            cause: error,
+          });
+        }
+        // A definitive platform rejection (recalled root, missing read permission): management is not
+        // provable. Negative-cache it — per process only, so a permission grant plus restart recovers.
+        const line = `${label} thread root ${rootId} is not readable — treating the thread as unmanaged: ${String(error)}`;
```
kid7st, line 542: Two different failure classes share one rejection branch and one warn-once budget: a per-resource verdict (recalled root — correctly per-root and expected to repeat) and an app-wide verdict (missing message-read permission — a configuration error that disables EVERY bare continuation). The comment names this exact risk, then handles it with a single process-lifetime WARN that decays to debug — so the app-wide case produces the same one line as one recalled root, and an operator scanning logs a day later sees nothing.

Direction: split on the code class — permission/auth-shaped rejections keep warning (or surface once in the startup capability report next to the `im:message.group_msg` line, where the operator already looks for scope problems); resource-shaped rejections keep the warn-once-then-debug policy.

src/channels/feishu/feishu.ts
```diff
@@ -442,29 +453,195 @@ function createFeishuRuntimeFactory(
...
+        // budget) rather than misclassifying the ask; past the deadline the delivery fails → re-push.
+        await withAckDeadline(botIdentity, `${label} bot identity resolution`, deadline);
+        // Still undefined after settling = bot/v3/info FAILED: app-wide degraded mode (warned at
+        // startup, mirrors group summon staying off). Fail closed without caching.
+        if (botOpenId === undefined) return false;
+      }
+      // One platform read per root, shared by every delivery awaiting it (sharers resume in arrival
+      // order; a joiner inherits the FIRST caller's deadline, still within one budget of that start).
+      // A message that skips this check (an @mention, a stop) can still enter the live queue ahead of
+      // a bare message stuck in this read window; `seq` is assigned at arrival, so recovery order is
```
kid7st: The comment answers a question nobody asked and skips the one this change actually opens. `seq` protects RECOVERY order, but the live queue is fed by `queue.accept` in acceptance order (`turn-queue.ts`: "run after this session's previous turn, in arrival order" = accept-call order). Making acceptance async means a bare continuation parked in the root-read window can be overtaken, within the SAME session, by a later message that skips the check (@mention, stop) — the user's two messages get answered out of order, and `seq` does nothing about it.

Direction: either serialize acceptance per chat/session (one promise chain around `acceptNewEvent`, keyed by `chat_id`) so the pre-ACK await cannot reorder a session, or state in the comment that live answer order is best-effort and only recovery order is ordered — but don't cite `seq` as if it covered this.

test/feishu.test.ts
```diff
@@ -670,6 +670,463 @@ describe("turn flow", () => {
...
+      );
+
+    vi.useFakeTimers();
+    const pending = Promise.resolve(handler(ask())).then(
+      () => {
+        throw new Error("expected the delivery to fail");
+      },
+      (error: unknown) => error,
+    );
+    await vi.advanceTimersByTimeAsync(2_500); // past ROOT_CHECK_TIMEOUT_MS
```
kid7st: `ROOT_CHECK_TIMEOUT_MS` does not exist — the constant is `ACK_CHECK_BUDGET_MS` (`feishu.ts:73`), and it is deliberately a SHARED budget across the identity + root reads, not a per-read timeout. The stale name is repeated at line 913, where the wait being exceeded is the identity read, making the mismatch actively misleading. Direction: name the real constant in both comments.

kid7st:
On-PR audit addressed (commits `1265d09`, `51bf556`) — all five inline threads resolved, plus the two review-body findings:

| Finding | Fix |
|---|---|
| chat-scope check fails open on an absent `chat_id` | fails closed like its sibling anomalies, warn + chat-scoped negative cache |
| `id_type === "open_id"` not enforced at the request | `getMessage` pins `?user_id_type=open_id`; the drift guard now fires unless a mention carries an open_id |
| recalled-root and permission failures share one warn-once budget | one warning per platform error code (new `feishuApiErrorCode`) |
| `seq` comment doesn't cover the ordering this opens | comment states live order is best-effort across the read window; only recovery order is guaranteed |
| tests cite `ROOT_CHECK_TIMEOUT_MS` | renamed to the shared `ACK_CHECK_BUDGET_MS`; timeout text now reports this wait's actual share |
| 429 retry loop can't fit the 2s pre-ACK budget | the pre-ACK read passes `noRateLimitRetry` (one attempt) |
| `docs/feishu.md` not updated | cache semantics + message-read scope documented there, in both scaffold channel headers, and in the `add feishu\|lark` next steps |

Also from the follow-up local pass: the managed-root cache is now bounded, the new pipeline options have direct tests (pinned id type / no-retry / abort), and Slack's same-named `owned-threads.ts` notes the deliberate divergence.

Local: lint + typecheck + 1020 tests green.

kid7st:
Closing in favour of a redesign of the whole interaction model, not because this change was wrong.

**What this PR fixed:** managed-thread recognition depended on local state, so losing the state disk demoted every existing thread to @-mention-only. It replaced that with a predicate derived from the thread's own root.

**Why it is being superseded:** using the bot in a real Lark tenant surfaced that the underlying model is what is wrong. The `threaded` / `continuous` option pair couples three independent axes — session identity, reply placement, and the summon rule — so wanting room-level sessions forced giving up mention-free thread continuations. Fixing the predicate inside that model leaves the coupling in place.

The replacement derives the rules from one axiom (the agent is a participant in the room, not an endpoint) and drops both modes: [docs/design/participant-model.md](https://github.com/fastagent-sh/fastagent/blob/main/docs/design/participant-model.md) in the follow-up PR.

**What carries forward** (this branch is not discarded):

- the platform-derived participation predicate, extended — root summons the bot, or the root is the bot's own message, or the bot has spoken in the thread;
- `managed-roots.ts` as a write-through cache with re-derivation on miss;
- the hardening this PR's review rounds produced: definitive-vs-transient rejection classification (`isFeishuApiRejection`), the shared pre-ACK deadline with request abort, no rate-limit backoff on the pre-ACK read, concurrent-delivery sharing, and the wire-shape handling for `getMessage` (`user_id_type=open_id`, string `id` + `id_type`, `sender_type`).

The summon rule itself changes: participation stops being sufficient, because a thread with several humans requires an explicit mention again.

PR #282, opened 2026-07-25
feat(channels)!: participant model for Feishu/Lark and Slack — speak where you take part, and have not heard a second human

kid7st:
Redesigns how the Feishu/Lark and Slack channels decide when to answer without being named, and
replaces the `threaded`/`continuous` session-mode pair with a single model: **the agent is a
participant in the room, not an endpoint.**

Design note: `docs/design/participant-model.md`.

## The rule

> Speak without being addressed if, and only if, the agent **takes part** in this place and has **not
> heard a second human** in it. Otherwise require an explicit mention.

Direct messages are not a special case — they are the instance where only one human can be heard. A
thread where a second person speaks returns to requiring a mention, while the agent keeps listening
and folds the discussion into its next answered turn.

## What makes it work: observation, not membership

Both facts come from what the channel **heard**. Nothing is read back from the platform.

An earlier revision of this branch did read the thread back on the acceptance path to learn who was
really in it. That claim was never sound — a 50-message page cap left it incomplete — and it cost a
per-platform failure taxonomy, an ACK budget, request aborts, a completeness flag with a refusal
sibling, and a duplicate-delivery join to keep acceptance safe once it could suspend. Nearly every
defect in the feature lived in that machinery. Removing it takes **~1100 lines** out, returns
acceptance to synchronous, and leaves the weaker claim the rule actually needs.

The cost, stated plainly: a thread the agent joined before this deployment — or before a lost state
file — reads as unheard and takes one mention to re-enter. That is the same bootstrap every thread
starts with, it self-heals in one message, and it is visible. The failure it replaces was silently
mention-only, forever, with no signal.

Telegram already worked this way and needed no change; the weakest-claim channel turned out to be the
simplest and least buggy, which is the argument that removed the read.

## Store discipline (`src/channels/thread-participants.ts`)

- **Never gated on configuration.** `groupBehavior` and `route` are the tempting gate, but they change
  on a running deployment while a record outlives the change — gate on them, switch back, and
  `agentSpoke` is on disk with the intervening humans missing. Structural facts only.
- **Both halves in one merge**, so a record can never say "answered here, heard nobody" (which the
  rule would admit forever).
- **Observations only accumulate.** No platform signals that someone left, and the error directions
  are asymmetric: over-counting asks to be named, under-counting speaks into a crowd.
- **Bystander-first eviction**, never evicting the thread being written.
- Keyed by the **session id** it describes, since the record is a claim about that memory.

## Breaking changes

**Feishu/Lark** — DMs become one continuous conversation; group summons answer in place instead of
opening a thread; sessions are re-keyed to the place, so existing history is not migrated (this reads
to users as the agent forgetting) and the whole room becomes one concurrency queue. Buffered
discussion in threads does not survive the re-keying.

**Slack** — `directMessageSession` / `groupMessageSession` are removed and now fail at startup. A
deployment that set either to `continuous` changes both placement and session keying.

Both: `owned-threads.json` is removed on the next start. `test/migration-deadline.test.ts` fails at
the release that must delete the one-time migrations.

## Also in this branch

- Always load a quoted message. Skipping it inside a known thread needed a second fact — whether the
  channel *received* the messages in between — that depends on a permission changing over time while
  the record is durable; every way of gating it failed toward a silently missing quote.
- **A separate fix riding along** (deliberately, not derived from the model): the scaffolded
  `feishu-send` / `lark-send` / `telegram-send` tool descriptions gained the "do NOT use this to answer
  the current turn" boundary Slack's already had — without it the agent posts its reply twice. Scaffolded
  files are copies, so an existing workspace keeps the old text; the one-line fix is in each channel
  doc's upgrade section.
- `im:message:readonly` is documented and requested as optional (it affects quoted messages only, not
  the thread rule), and `im:message` is accepted as its superset.
- **The prompt-side text bounds are derived now, not picked.** Three different jobs shared the number
  280, and the referent bound had drifted 14x between two channels, because each number lived in the
  channel that used it rather than with what derives it. Split into two species, each next to its
  derivation: `REFERENT_MAX_CODE_POINTS` (a *fidelity* bound — a referent is the exact text the asker
  points at, so the cut must clear the largest message a chat platform accepts: 4096, Telegram's cap
  and the tightest of ours) and `BUFFER_LINE_MAX_CHARS` (a *fairness* quota — many messages compete
  for one 4000-char fold budget, so no single rambler may price out everyone after them). Display
  bounds stay in the renderers. Two rules now hold for every prompt-side cut, both previously
  violated: cut on code-point boundaries (Telegram used raw `slice()`, which tears an emoji), and mark
  the cut (a truncated referent used to reach the model as a stump with no signal). This is the one
  place Telegram changes beyond its scaffolded tool description.

## Rebase note (after #279)

#279 removes the `taskDisplay` option outright, along with `SlackTaskDisplayMode` and the `chunks`
mode of the Slack stream API. This branch still carries five `taskDisplay` references in
`src/channels/slack/slack.ts` (the interface field, the destructure, its validation, and the
`streamSlackReply` argument) — delete them while resolving. Nothing here needs to be remembered:
each residue is an independent typecheck failure after the rebase.

Verified live against a Lark tenant: upgrade cleanup on real data, and state-loss self-heal (deleted
`thread-participants.json` → the next mention re-enters the thread).

1038 tests / 90 files green. Audited to `No findings.`


src/channels/feishu/invoke-turn.ts
```diff
@@ -75,8 +81,18 @@ async function resolveTurnInputs(t: FeishuTurnTransport, attachments: FeishuTurn
...
+    // the user their answer — every first message of a thread carries one, so a hard failure here
+    // would turn an ordinary platform edge into a lost turn. Degrade visibly instead: the operator
+    // gets a warning, and the model is told the quote could not be read rather than being left to
+    // guess what "about that" refers to.
+    const parent = await t.api.getMessage(parentId).catch((error) => {
+      log.warn(`${t.label} could not read replied-to message ${parentId}: ${String(error)}`);
+      return undefined;
+    });
+    if (!parent) {
+      return { images: undefined, promptSuffix: `\n\n[replied-to message (msg ${parentId}) could not be read]` };
```
kid7st: The referent degradation drops the *ask's own* payload. This early return bypasses the entire primary/buffered resource block below: a user who replies to a deleted/restricted message *and attaches an image* gets `images: undefined` and no file manifest — the resources they explicitly pointed at vanish silently, along with the buffered context this turn was to fold. The stated intent ("a referent is CONTEXT, not the ask") is inverted by the implementation: losing the context costs the ask.

Set `referentBlock` to the marker and fall through to resource resolution instead of returning. No test covers the unreadable-referent path at all, so nothing catches this.

src/channels/feishu/feishu.ts
```diff
@@ -541,21 +719,29 @@ function createFeishuRuntimeFactory(
       const baseText = r.text ?? cloudEnvelope(event, kind);
       if (baseText.trim() === "" && images.length === 0 && files.length === 0) return;
 
-      if (route === undefined && threadedGroup && m.thread_id === undefined && sameTarget && replyInThread === true) {
-        ownedThreads.add(m.chat_id, m.message_id);
+      // Answering inside a thread makes the agent a participant of it, which is what lets the NEXT
+      // bare message address it without a mention (§3). Recorded at acceptance, before the reply is
+      // attempted: entering the conversation is the intent, and a delivery failure does not undo it.
+      if (replyInThread === true && m.thread_id !== undefined && sameTarget) {
+        threadParticipants.merge(m.chat_id, m.thread_id, { agentSpoke: true });
```
kid7st: `agentSpoke` is recorded before the intent is persisted, so a failed persist loses the referent anchor on the re-push. `submit(..., true)` can throw (`store.add` → disk failure → 500 → platform re-push), but this merge already happened; the re-push then samples `agentInThread === true` and sends `parentId: undefined`, so the thread's first message arrives with an empty session *and* no anchor.

This is the failure 85203bb fixed for observed humans ("observation writes one BEFORE acceptance can fail"), reintroduced through the other field. The comment's justification is about the *reply* failing, not about acceptance failing. Move the merge after `submit` returns.

src/channels/feishu/feishu.ts
```diff
@@ -442,29 +439,215 @@ function createFeishuRuntimeFactory(
...
+        }
+        threadParticipants.merge(chatId, threadId, { humans: speaker ? [speaker] : [], derived: true });
+        return;
+      }
+      threadParticipants.merge(chatId, threadId, {
+        // The window may not carry the message being accepted right now, but whoever is speaking is
+        // in the conversation by definition — a derived record must not drop them.
+        humans: [
+          ...senders.filter((sender) => sender.senderType === "user").map((sender) => sender.senderId),
+          ...(speaker ? [speaker] : []),
```
kid7st: A shared thread read compensates only for the FIRST caller's speaker, so a second human can be erased and then answered. A derived listing REPLACES the human set and its window may not yet contain a just-pushed message — that is patched here by appending `speaker`, but `readThreadParticipants` is memoized per thread (line ~572). When Alice's and Bob's bare messages race on a non-derived thread, Bob joins Alice's read, his just-observed presence is wiped by the replace, and his delivery then evaluates `humans.length === 1` → answered. That is exactly the multi-human barging §3 forbids.

Have each waiter merge its own speaker *after* awaiting the shared read (and drop the `speaker` parameter from the read), so the compensation is per delivery instead of per read.

docs/design/participant-model.md
```diff
@@ -0,0 +1,254 @@
...
+A side conversation loses two different things, and they have asymmetric costs:
+
+| Loss | Who suffers | Cost of fixing automatically |
+|---|---|---|
+| The room's session does not contain the thread's content | the agent (amnesiac when back in the room) | **none** — nobody sees it |
+| The people in the room do not know the thread's outcome | other humans | **high** — a message in the main timeline |
+
+> **Memory folds automatically; attention must be requested explicitly.**
+
+**Memory fold (automatic).** When a turn in a thread completes, it records that thread's *latest*
```
kid7st: §7's memory fold is written as implemented; it does not exist. Nothing in `src/` records a thread exchange into the room's bucket — `bufferKey` is always the message's own place and `commit` only clears it. §8 carefully labels its rungs ("Rung 2 is implemented", "not planned"); §7 (and the §5 table row "Thread → room | conclusions fold upward") does not, so the document advertised as the *why* behind the code states a mechanism the code lacks.

Mark §7's memory fold as planned, or implement it.

docs/design/core.md
```diff
@@ -334,25 +334,50 @@ a stable hand-authored surface. What is platform-different:
...
+  decisions:
+  - *Predicate.* Who takes part comes from the messages the channel observes, and — for a thread this
+    process has never completed a picture of — from `listThreadSenders` (`container_id_type=thread`,
+    pinned to `user_id_type=open_id`; a bot sender's `id` is the app id). The earlier design
+    recognized only threads this process had created, so a lost state disk silently demoted every
+    existing thread to @-mention-only.
+  - *Caching.* `thread-participants.json` is a bounded write-through cache carrying, per thread, the
+    humans seen, whether this agent has spoken, and whether the record is `derived` (complete).
+    Observation alone leaves `derived` false, so a failed derivation is retried rather than mistaken
+    for an answer; the agent's own reply sets it, since that settles the half of the rule that
```
kid7st: "the agent's own reply sets it" (`derived`) contradicts the code and its own test: `thread-participants.ts` states that only a platform listing sets `derived`, and `test/feishu-thread-participants.test.ts` asserts "only a platform listing completes the record: the agent's own reply does not". Stale from an earlier iteration, and it makes the failure-retry reasoning in the same paragraph unreadable. Say the listing sets it.

test/feishu.test.ts
```diff
@@ -1410,15 +1613,26 @@ describe("turn flow", () => {
     expect(existsSync(join(home, "buffers.json"))).toBe(false);
   });
 
-  it("does not persist managed-thread ownership when a custom route owns admission", async () => {
+  it("a custom route owns admission: no participation is derived and no thread state is written", async () => {
```
kid7st: The test name asserts the opposite of the body: "no thread state is written", while the assertions require `participants["oc_1:omt_custom"]` to be `{ agentSpoke: true, derived: false }` — state *is* written. The name is this test's spec; a reader trusting it would read a regression here as correct. Rename to what is verified: no derivation, but the agent's own participation is still recorded.

test/feishu-thread-participants.test.ts
```diff
@@ -0,0 +1,111 @@
...
+    store.merge("oc_1", "omt_a", { agentSpoke: true });
+    // The agent replying proves it takes part, but says nothing about who ELSE is in the thread —
+    // the half the summon rule counts.
+    expect(store.get("oc_1", "omt_a")).toEqual({ humans: ["ou_alex"], agentSpoke: true, derived: false });
+
+    store.merge("oc_1", "omt_a", { humans: ["ou_alex", "ou_bob"], derived: true });
+    expect(store.get("oc_1", "omt_a")).toEqual({ humans: ["ou_alex", "ou_bob"], agentSpoke: true, derived: true });
+  });
+
+  it("a derived listing REPLACES the human set: a thread can become two-party again", () => {
```
kid7st: This asserts a transition the channel cannot make. `derived` is monotone and the channel re-reads only while it is false, so the replace runs at most once per thread per process (the `ponytail:` comment in `merge` says so) — "a thread can become two-party again" is unreachable in production. The test validates a store capability under a scenario whose user-visible claim is false. Either state the reachable scenario (a re-derive after restart) or give the channel a way to re-derive.

kid7st (commented):
Thesis: participation becomes a property of the thread (observed per message, re-derived from the platform, cached in `thread-participants.json`), and the summon/placement/memory rules are derived from that instead of selected by a session mode.

**Terminology drift left in user-facing text.** `src/scaffold/add-channel.ts:133,157,171,181,206,217` still say "bare managed-thread replies"; `src/channels/feishu/scaffold/channel.ts:11,37` and `src/channels/lark/scaffold/channel.ts:11,35` still say "Agent-managed threads" / "bare continuations in Agent-managed …"; `src/channels/feishu/parse.ts:86` still documents "unmentioned continuations from its managed-root index" — the index this PR deletes. `docs/feishu.md` was converted, the onboarding output and the authoring surface's docstring were not, so the CLI teaches a model that no longer exists.

**Obsolete state cleanup is half-applied.** `feishu.ts:263` removes `owned-threads.json` on start, but the orphaned `<chat>:root:<id>` buckets in `buffers.json` are left indefinitely, holding chat content that can never be folded into a turn. The documented remedy is "delete the file", which also destroys live pending context. Prune keys containing `:root:` in the same startup step.

kid7st:
On-PR audit addressed in `d2b0d78` — all 7 inline threads resolved, plus both review-body findings.

**Real defects (all three now covered by tests):**

| Finding | Fix |
|---|---|
| An unreadable referent returned early, dropping the images/files attached to the message being accepted | falls through to resource resolution — the quote is context, the attachments are the ask |
| `agentSpoke` recorded before `submit` persisted the intent, so a re-pushed first message lost its referent anchor | recorded after the intent is durable (the same failure fixed for observed humans, reintroduced through the other field) |
| The shared thread read compensated only its first caller's speaker, so a second human racing it was erased and then answered | each waiter re-asserts its own speaker after the read |

**Accuracy and cleanup:** §7's memory fold is now marked planned (it is the next increment, not shipped); core.md no longer claims the agent's own reply sets `derived`; the two test names that specified the opposite of their body are renamed; retired `<chat>:root:<id>` buckets are dropped at load via a place-key predicate on the generic buffer, rather than leaving chat content nothing can fold; and the CLI/scaffold text no longer teaches "Agent-managed threads".

Local: lint + typecheck + 1024 tests green.

src/channels/feishu/thread-participants.ts
```diff
@@ -0,0 +1,146 @@
...
+      const key = keyOf(chatId, threadId);
+      const previous = records.get(key);
+      // A derived listing REPLACES the human set: it samples the thread's recent window, so it is the
+      // answer to "who is in this conversation now" rather than "who ever spoke". An EMPTY set never
+      // replaces: a listing that names nobody is an anomaly, not the news that the thread is empty,
+      // and dropping the humans already seen there could admit a message it must not.
+      //
+      // ponytail: `derived` is monotone and the channel re-reads only while it is false, so this
+      // branch runs at most ONCE per thread per process — a second human observed afterwards is
+      // terminal until a restart re-derives. Deliberate: re-reading on a timer would put a platform
```
kid7st: `derived` is persisted **and** monotone, and the channel re-reads only while it is `false` — so a thread is derived at most **once ever**, not once per process. "until a restart re-derives" is false: the restart loads `derived:true` from `thread-participants.json` and never reads again.

That invalidates every "recent window" claim in this change (this comment, the file header L13, `docs/design/participant-model.md:83`, `docs/feishu.md:264`, `docs/design/core.md:364`): a thread the agent has been silent in stays a participant forever, a second human is terminal forever, and "a thread that quietened back to two parties behaves like one again" never happens. Re-derivation is reachable only through state loss or `MAX_THREADS` eviction — which also makes the `replacing` branch's stated rationale (`test/feishu-thread-participants.test.ts:55`: "the shape a restart's re-derive relies on") a scenario the system cannot reach, since a re-derive after state loss has no previous record to replace.

`updatedAt` is written and shape-validated but never read — that is exactly where the missing invalidation belongs. Either give `derived` a staleness TTL on `updatedAt` (re-derive when older than the window), or stop persisting it; then make the five texts state whichever was chosen.

src/channels/feishu/feishu.ts
```diff
@@ -442,29 +444,205 @@ function createFeishuRuntimeFactory(
...
+          // addresses the agent, so most deliveries hitting it are ordinary thread chatter that would
+          // only have buffered — the trade-off is: unknown-addressing messages fail closed (HTTP 500 /
+          // WS 500 frame → platform re-push → re-check), because ACKing here would silently downgrade
+          // a GENUINE ask to background context. If every bounded re-push fails, both the turn and the
+          // buffered copy are lost (operator log only) — accepted; see docs/design/participant-model.md.
+          throw new Error(`${label} could not read thread ${threadId} pre-ACK: ${String(error)}`, { cause: error });
+        }
+        // A definitive platform rejection (deleted thread, missing read permission): participation is
+        // not derivable. Record the thread as seen-but-empty so the read is not retried per message;
+        // the agent then stays @-only there, and a permission grant plus restart recovers.
```
kid7st: This path writes `derived: true` for a **failed** read, so one field carries two meanings — "the human set is authoritative" and "do not retry" — and the failure becomes durable. Both claims on this line are then false:

- "a permission grant plus restart recovers" — it does not. The poisoned record survives the restart (see the `thread-participants.ts` comment), so a tenant that grants the read scope afterwards has to delete `thread-participants.json` by hand.
- "the agent then stays @-only there" — only when the record is empty. A thread where the agent already answered an @mention has `agentSpoke:true, derived:false`; the next bare message triggers the read, the rejection stamps `derived:true`, and the predicate now sees `agentSpoke && humans.length === 1` from a purely *observed* human set that the store's own doc says can under-count. The agent then answers unprompted in a thread that may hold several humans — the barge-in §3 exists to prevent. `test/feishu.test.ts:886` only covers the empty-record branch, so neither consequence is caught.

Keep the read failure out of `derived` (in-memory, or a short TTL) and treat a non-derived record as mention-only for the summon decision.

src/channels/feishu/feishu.ts
```diff
@@ -541,26 +714,36 @@ function createFeishuRuntimeFactory(
...
           chatId,
           replyTo,
           queueReplyTo,
           replyInThread,
-          parentId: threadedConversation && m.thread_id !== undefined ? undefined : m.parent_id,
+          // The referent anchor is what a thread starts from (§8 rung 2). Until the agent has answered
+          // in a thread its session is empty, so the replied-to message is the only context there is.
+          // Once it has, the session holds what later messages reply to and re-fetching would cost a
+          // call per turn and duplicate it.
+          parentId: agentInThread ? undefined : m.parent_id,
```
kid7st: The guard is justified as the thread ladder's rung 2 ("until the agent has answered in a thread its session is empty"), but `agentInThread` is false for every non-thread place, so this is unconditional outside threads: a quoted @mention in the main timeline and every quoted direct-message reply re-fetch the parent and re-inject up to 4000 code points that the place's continuous session (§5) already holds — one extra `getMessage` per turn plus duplicated prompt text, on the common path.

Either express the condition that is actually meant ("this place's session has no history yet") or state the main-timeline/DM policy explicitly rather than describing it as thread anchoring.

src/channels/feishu/feishu.ts
```diff
@@ -442,29 +444,205 @@ function createFeishuRuntimeFactory(
...
+          () => abort.abort(),
+        );
+      } catch (error) {
+        if (!isFeishuApiRejection(error)) {
+          // Transient (network/5xx/rate limit/timeout). This runs BEFORE we know whether the message
+          // addresses the agent, so most deliveries hitting it are ordinary thread chatter that would
+          // only have buffered — the trade-off is: unknown-addressing messages fail closed (HTTP 500 /
+          // WS 500 frame → platform re-push → re-check), because ACKing here would silently downgrade
+          // a GENUINE ask to background context. If every bounded re-push fails, both the turn and the
+          // buffered copy are lost (operator log only) — accepted; see docs/design/participant-model.md.
```
kid7st: Dangling reference: the accepted "both the turn and the buffered copy are lost" trade-off is argued in `docs/design/core.md` (§*Failure ownership*), not in `docs/design/participant-model.md`, which says nothing about delivery failure or the pre-ACK path. Point at core.md.

src/channels/context-buffer.ts
```diff
@@ -69,7 +73,12 @@ export function createContextBuffer<E>(options: {
       !Array.isArray(raw) &&
       Object.values(raw).every((entries) => Array.isArray(entries) && entries.every(isEntry))
     ) {
-      return new Map(Object.entries(raw as Record<string, E[]>));
+      const loaded = Object.entries(raw as Record<string, E[]>);
+      const live = isLivePlaceKey ? loaded.filter(([placeKey]) => isLivePlaceKey(placeKey)) : loaded;
```
kid7st: This drop and the `rmSync(owned-threads.json)` in `feishu.ts` are the only irreversible steps in the PR, and neither has a test: this one deletes persisted chat content at load, that one deletes operator state at start. No test asserts that live buckets survive the filter, that `<chat>:root:<id>` buckets are gone, or that the obsolete file is removed (`test/slack.test.ts:532` covers Slack's own file only). Add one test per deletion.

docs/feishu.md
```diff
@@ -226,61 +224,58 @@ type FeishuRoute = {
...
+message in that thread reaches it. When a second person speaks in the thread, addressing becomes
+ambiguous again and the Agent goes back to requiring a mention — while still listening, so the
+discussion is folded into its next answered turn there.
+
+Participation is a property of the thread, not of this process: for a thread it has no complete
+picture of, the channel reads the thread's recent senders back from the platform, so losing local
+state costs one lookup rather than the behaviour. Both halves are read from that recent window — a
+thread the Agent has been silent in for a long stretch reads as one it is no longer part of, and a
+single mention puts it back. That read happens before the event is acknowledged and shares a
+2s budget with the bot-identity lookup; if it fails transiently the delivery fails and the platform
```
kid7st: "2s budget" contradicts the code: `ACK_CHECK_BUDGET_MS` is 2500ms, with a comment explaining specifically why 2s left no headroom. The PR description repeats the stale number too.

docs/design/participant-model.md
```diff
@@ -0,0 +1,237 @@
...
+| Per-entry concurrency inside a session | divergence without convergence; stale reads |
+| Automatic summaries posted to the room | attention has a noise cost, so it needs consent |
+
+## 10. Mapping to platforms
+
+The model is platform-neutral; the primitives differ in strength.
+
+| Capability | Feishu / Lark | Slack |
+|---|---|---|
+| Side conversation | topic (`thread_id` + `root_id`) | thread (`thread_ts`) |
```
kid7st: The mapping table still names the Feishu side conversation `thread_id` + `root_id`, while this PR's central platform finding is that `root_id` is not thread-stable and identifies nothing. Drop `root_id` from the primitive.

kid7st (commented):
**Thesis** — this replaces the `threaded`/`continuous` mode pair with a *participant model*: a bare thread message is answered iff the agent takes part in the thread and exactly one human does, where participation is a per-thread cache refined by observation and re-derived once from a pre-ACK platform listing; sessions/context buckets re-key from `root_id` to `thread_id`.

Two findings cannot anchor to a diff line — both are omissions in files this PR did not touch.

**Removed options are silently ignored, and still advertised.** `directMessageSession` / `groupMessageSession` are deleted together with `validateSessionOptions`, so an existing `channels/feishu.ts` that sets them keeps running with the new behavior and no signal at all (excess properties are ignored at runtime; TypeScript catches it only if the author typechecks their workspace). Meanwhile the scaffold the CLI still emits advertises them: `src/channels/feishu/scaffold/channel.ts:28-30`, `src/channels/lark/scaffold/channel.ts:26-28`, and `docs/channels.md:178-180` (the Feishu example, right after the Slack one that legitimately keeps them). Delete those three references and warn at mount when either key is present.

**The new required capability is prose-only.** `listThreadSenders` makes a message-read scope (`im:message:readonly`) a hard precondition for the entire bare-reply half of the model, yet the onboarding automation requests only `im:message.group_msg` (`src/cli/add-feishu.ts:120`) and `application:application:patch` (`:298`), and the startup capability report probes only the group scope. A tenant that completes `add feishu` exactly as instructed can therefore have every thread demoted to mention-only — permanently, see the `derived` findings — announced by a single `warn` line. Add the read scope to `addAppScopes`/`addons` and report it next to the group-visibility line at startup.

kid7st:
Second full audit addressed in `583634f` — 7 inline threads resolved, plus both body findings.

**The one that mattered:** `derived` was persisted *and* monotone, so a thread was established once **ever** — a restart loaded the flag back and never re-read. That invalidated every "recent window" claim in the change, and made a refused read a durable *do not retry* that a later permission grant could not clear without deleting the file by hand. Observations stay durable (they only under-count); `derived` is now process-local, so each process re-establishes a thread once and that is what bounds staleness.

Its sibling: the summon predicate ignored `derived` entirely, so after a refused read a purely *observed* human set — which the store's own doc says can under-count — was treated as authoritative, and the agent could speak unprompted in a thread holding several humans. Speaking unprompted now requires an established record.

| Also fixed | |
|---|---|
| Retired context buckets were filtered in memory but left on disk | rewritten at load, so the deletion actually happens |
| Neither irreversible upgrade step had a test | one test covers both (obsolete file removed, live buckets survive) |
| The referent guard was documented as thread anchoring only | states what it does outside threads too: a quote there is the user pointing at something that may predate the session |
| `2s budget`, `root_id` as a thread primitive, the removed options in `docs/channels.md`, a dangling design reference | corrected |

Local: lint + typecheck + 1028 tests green.

kid7st (commented):
Thesis: the change replaces each channel's local "owned thread" index with a shared *thread participation* concept — who takes part in a thread, unioned from this process's observations and one platform listing — and derives the summon, placement and memory rules from a participant axiom instead of selecting them with session modes.

Finding that does not anchor to a diff line:

**The PR description is stale in the places that become the squash commit.** `CONTRIBUTING.md` makes the PR title and body the commit message, and both now describe a smaller change than the branch contains:

- the title is `feat(feishu)!` while the branch also lands a second breaking change to Slack (session options removed, summon rule replaced);
- "## Scope — Feishu/Lark. **Slack keeps its own session options** and Telegram its own threading; converging them is follow-up" states the opposite of commits `daed1e3` and `6b82103`;
- "read back from the platform (`container_id_type=thread`, **newest-first page**)" — the implementation reads `ByCreateTimeAsc`, and the design note now documents an opening-page ceiling that depends on that direction;
- "deadline-bounded (**2s** …)" — `ACK_CHECK_BUDGET_MS` is 2500, changed deliberately against measured round trips;
- the test count (1022) is now 1041;
- the BREAKING paragraph lists only Feishu's three changes and omits Slack's.

Direction: retitle to cover both channels, replace the Scope section with the per-platform reach the design note now carries, correct the read direction and the budget, and add Slack's breaking changes to the BREAKING paragraph.

src/channels/thread-participants.ts
```diff
@@ -0,0 +1,181 @@
...
+/** What a restart keeps: observations, which only under-count. */
+interface StoredParticipation {
+  humans: string[];
+  agentSpoke: boolean;
+}
+
+interface ThreadParticipation extends StoredParticipation {
+  /** Whether THIS process established the human set from a platform listing. Speaking unprompted
+   *  requires it: observation alone can only UNDER-count the humans in a thread. */
+  derived: boolean;
```
kid7st: `derived` names how the fact was obtained, not the guarantee it licenses.

This flag is the one that permits speaking unprompted, so what it must assert is "the human set is authoritative". `derived` says only "came from a derivation". The module then needs a dedicated paragraph plus a second flag (`unreadable`) whose entire docstring exists to stop a reader collapsing the two — and commit `81aab4f` in this same branch fixed a shipped bug that was exactly that collapse (a refusal wrote `derived`, and an observed-only record started passing the summon rule). A field whose name has to be defended in prose is the naming defect, not the prose.

Direction: rename to what it asserts — `humansEstablished` (or `authoritative`) — leaving `unreadable` as the "do not ask again" sibling, so the two names no longer occupy the same conceptual slot.

src/channels/feishu/feishu.ts
```diff
@@ -442,29 +467,177 @@ function createFeishuRuntimeFactory(
...
+      const seq = ++seqCounter; // arrival order, taken BEFORE any await below can reorder acceptance
       const bufferKey = feishuBufferPlaceKey(normalized.conversation);
       const isHumanGroup = event.sender?.sender_type === "user" && m.chat_type === "group";
-      const managedThread =
-        groupMessageSession === "threaded" &&
+      // Has THIS PROCESS answered in this thread before this delivery? That is the cheap proxy for
+      // "the thread's session already holds what this message replies to". It is deliberately read
+      // before the participation lookup below, and never from "a record exists": observation writes
+      // one before acceptance can fail, which would strip a re-pushed first message of its anchor.
+      // The cost of the proxy: after a restart, the first continuation in a long-running thread loads
```
kid7st: The stated cost of this proxy does not exist.

The comment owns a trade-off: "after a restart, the first continuation in a long-running thread loads its referent again — one extra read and some duplicated text". But `agentSpoke` is part of `StoredParticipation`, is persisted, and is restored by the loader, so a restart preserves it and the referent is *not* re-fetched. The cost only materialises on state **loss** (a new machine, or eviction under `MAX_THREADS`). As written, the comment invites a future reader to remove the proxy in exchange for a cost that was never being paid.

Direction: say "after state loss (a new machine, or eviction under the cap)" rather than "after a restart".

src/channels/slack/slack.ts
```diff
@@ -265,7 +279,17 @@ export function slackChannel({
...
     const seen = createSeenRing(join(stateHome, "seen.json"), label);
-    const ownedThreads = createOwnedSlackThreads(join(stateHome, "owned-threads.json"), label);
+    const threadParticipants = createThreadParticipants(join(stateHome, "thread-participants.json"), label);
+    /** This channel's place key for a thread (the shared store is key-agnostic). */
+    const threadKey = (teamId: string, channelId: string, threadTs: string): string =>
+      `${teamId}:${channelId}:${threadTs}`;
+    // The participant model replaced the owned-thread index; its file is dead weight on an upgraded
+    // deployment (a cache, so nothing is lost). Best-effort: a leftover file is untidy, not fatal.
+    try {
+      rmSync(join(stateHome, "owned-threads.json"), { force: true });
```
kid7st: An irreversible upgrade step with no test — and the one assertion that names the file is now vacuous.

This deletes operator state at construction. Feishu's identical step is covered (`test/feishu.test.ts`, "removes the obsolete owned-threads.json and drops only the retired context buckets"); Slack's is not. The only mention of the filename in Slack's suite is `test/slack.test.ts:644`, which asserts the file *cannot be read* — that now passes whether or not this deletion works, because nothing in the codebase can create it any more. So the half of this commit that touches disk is guarded by an assertion that has become a tautology.

Direction: mirror the Feishu test — write an `owned-threads.json` into the state home, construct the channel, assert it is gone; and either delete the vacuous line or repoint it at the real property.

kid7st:
All four findings addressed in `212d82f`, plus the description rewrite above.

| Finding | Fix |
|---|---|
| `derived` named its provenance, not the guarantee it licenses | renamed to `established`; `unreadable` now reads as the different claim it is. The docstrings say *authoritative human set* rather than defending the name |
| Slack's `rmSync(owned-threads.json)` untested, and the assertion naming it was vacuous | replaced with the Feishu-shaped test — write the file, construct the channel, assert it is gone. Verified it fails when the deletion is removed, which the old assertion did not |
| `agentInThread` claimed to be process-local and to cost a re-fetch after a restart | `agentSpoke` is persisted and restored, so neither was true: it survives restarts, and the cost lands on state loss |
| PR title and body described a smaller change than the branch | retitled to cover both channels; the Scope section is replaced by the per-platform mapping table; read direction, budget and test count corrected; both channels' breaking changes now listed |

One note on the rename: a blanket `derived` → `established` also hit sentences where *derived* was ordinary English (the state home is "derived, not an option"; "the rules are derived from one axiom"). Those were put back — only the flag changed name.

PR #293, opened 2026-07-28
feat: AWS Bedrock AgentCore deploy target (deploy agentcore)

Sukitly:
Adds **AWS Bedrock AgentCore Runtime** as the fourth deploy target, alongside `docker|fly|railway`:

```bash
fastagent deploy agentcore          # artifacts + runbook
fastagent deploy agentcore --run    # aws + docker CLIs driven to completion
```

## Why it differs in kind from the resident-box hosts

AgentCore has **no public URL** (ingress is the SigV4 `InvokeAgentRuntime` API only) and **no resident process** (per-session microVMs, reclaimed when idle). Serving fastagent there means fronting webhooks with a forwarder and replacing the resident cron/wake timers with an external clock.

## Deployment topology

One generated CloudFormation stack (`agentcore.template.yaml`) is the whole topology:

```mermaid
flowchart TB
    tg["Telegram / Feishu / Slack<br/>webhooks"]
    sdk["Programmatic callers<br/>(any AWS principal)"]

    subgraph acct["The operator's AWS account"]
        subgraph stack["One CloudFormation stack — agentcore.template.yaml"]
            fwd["<b>Forwarder Lambda</b><br/>Function URL only when something<br/>outside AWS must reach in"]
            eb["<b>EventBridge Scheduler</b><br/>cron rules + one-shot wake alarms"]
            subgraph rt["<b>AgentCore Runtime</b> — per-session microVM"]
                app["fastagent container (unchanged)<br/>POST /invocations · GET /ping"]
            end
        end
        s3[("S3 bucket — OUTSIDE the stack<br/>state snapshot + forwarder package")]
        ecr[("ECR — agent image")]
    end

    tg -->|"HTTPS POST, verbatim"| fwd
    eb -->|"fire / poke"| fwd
    fwd -->|"SigV4 + shared secret<br/>ONE ingress session"| app
    sdk -->|"SigV4, no secret →<br/>public plane: invoke only"| app
    app -.->|"wake alarm sync"| fwd
    fwd -.->|"one-shot at(fireAt)"| eb
    app -. "restore on first envelope,<br/>snapshot when work settles<br/>(presigned URLs)" .-> s3
    ecr -.-> rt
```

The image is the shared portable container, built locally (`docker buildx`, linux/arm64) and pushed to the account's ECR.

**Outside the stack, deliberately**: the ECR repository and the S3 bucket. The bucket holds the agent's memory, so `delete-stack` — or a rolled-back create — must not be able to take it.

**Stack resources**: Runtime + ExecutionRole; SessionStorage mount (`/mnt/state`) as the local disk; Forwarder Lambda + Function URL (+ the post-Oct-2025 **dual** invoke permissions) + ForwarderRole; `WakeSchedulerRole` + `FastagentWakeSecret` (NoEcho, minted by `--run`) when `selfSchedule`; one `AWS::Scheduler::Schedule` per translated cron.

**Traffic paths** (1–5 below; the deploy adds a sixth: `--run` sends a `checkpoint` envelope before stopping the ingress session, so a turn cut mid-flight has its durable intent in S3 and can be replayed):

1. **Webhook** — platform POST → Lambda wraps the request verbatim → the container reconstructs it and dispatches to the SAME channel routes (signature verification unchanged); the channel's real HTTP response rides back inside a transport-200 reply and is re-emitted byte-exact (Feishu URL-verification works). The ACK returns fast; the turn runs post-ACK while `/ping` reports `HealthyBusy` so idle reclaim can't kill it.
2. **Cron** — deploy-time-translated EventBridge rules fire the forwarder → `schedule-fire` envelope → slot-idempotent claim (delivery is at-least-once) → run + audit. No resident timers (`externalClock` mode).
3. **Programmatic invoke** — callers hit `InvokeAgentRuntime` directly; SSE streams back; per-session microVM isolation.
4. **Wake registration** — the agent's `wake` writes the store → a sink POSTs the pending set to the forwarder's reserved path (shared secret; bounded retry counted as in-flight work) → one-shot EventBridge schedules `at(fireAt)`, self-deleting. The container never calls AWS APIs itself.
5. **Wake fire** — the alarm pokes the forwarder → `wake-poke` envelope wakes the container → the existing boot-drain/30s wake pump fires the due entry; a recurring wake re-arms itself through the same store-save → sink loop.

**Key decisions**: ONE fixed ingress session for all webhook/cron/poke traffic (fastagent channel state is single-writer; a stopped session's id stays valid until the runtime is deleted); `--run` stops the ingress session after a successful update so the new image serves immediately (a live session would keep answering from the previous image).

## State durability (the one thing that is NOT obvious)

AgentCore's managed SessionStorage is **wiped on every runtime version update — i.e. on every deploy** — and after 14 idle days. `/mnt/state` is therefore a fast local disk, not the source of truth. Durability is one gzipped S3 object:

- the container **restores** the state root on its first ingress envelope and **pushes** a coalesced snapshot on the 0-in-flight edge (post-ACK turns finish long after the request returns, so "the request ended" is the wrong moment);
- the container holds **no AWS credentials** (verified on a live box), so the forwarder mints **SigV4-presigned GET/PUT URLs** and rides them on every envelope — the container stays AWS-SDK-free and credential-free;
- **fail-visible**: a snapshot that exists but will not restore 503s the request, because serving an empty agent would then overwrite the good copy with that emptiness. A 404 is first boot; `auth.json` restores absent-only (the deploy seeds a copy at least as fresh);
- the **bucket lives outside the stack** (like the ECR repo), so `delete-stack` cannot take the agent's memory with it.

A durable *mount* instead (EFS or S3 Files) requires VPC mode, which forces a NAT gateway for model/channel egress — a fixed ~$33/mo floor under a target whose whole appeal is pay-per-use. The template comments show the switch for anyone who wants it.

## Two planes, one entry point

`InvokeAgentRuntime` is an ordinary IAM action, so "the request arrived" proves nothing about who sent it. The adapter therefore splits its callers:

| | Reaches it how | May run | Carries internal fields |
|---|---|---|---|
| **Forwarder** (webhooks, cron, wake pokes) | shared secret minted per deploy | every kind | yes — state URLs, wake URL |
| **Anyone with the IAM action** | SigV4 only | `invoke` | no — dropped before dispatch |

Without that split, any principal holding the action could forge a `schedule-fire`, or ride a `state` URL on a public `invoke` to have the container upload its own `auth.json` to an address of their choosing.

Two matching decisions in the template: a **public Function URL is created only when something outside AWS must reach in** (a schedule-only deployment gets none — otherwise it would publish the builtin unauthenticated `POST /invoke` that a channel-less workspace mounts), and the Function URL's second permission carries `InvokedViaFunctionUrl: true` so the bare `*` principal cannot call the Lambda API directly and bypass the URL event shape.

## Verified on a real AWS deployment

An end-to-end Telegram deployment surfaced and fixed: **the SessionStorage wipe above** (caught by matching a conversation's truncation point to a deploy timestamp — the runtime ARN was unchanged, so this was the platform's documented version-update reset, not a CFN replacement), OAuth auth.json exceeding the 2048-char env cap (now chunked), CFN inline code being CJS-only (forwarder rewritten), the post-Oct-2025 dual Function-URL permission requirement, ROLLBACK_COMPLETE self-healing, and the stale-session-after-redeploy trap (auto-stop).

## Review-driven hardening

dow cron parsing (steps/ranges), query-string passthrough, alarm-sync failure propagation + bounded retry, github/schedule turns counted as in-flight work, template-drift + docker-daemon + identifier-collision + name-length gates, stderr-classified stop failures, an EXECUTED forwarder test harness (fake AWS clients, not substring assertions), and the user-docs pass (deploy guide section, CLI reference, README/overview/quickstart).

## Limits (gated/warned/documented)

Long-connection channels cannot run here (webhook mode is the path); a wake set inside a direct-invoke session has no alarm (its session storage is its own); the arm64 image builds locally (docker buildx required); deleting the deployment bucket starts the agent blank.

Verification: `npm run lint && npm run typecheck && npm test` (1182 passing; +~90 new cases incl. adapter/scheduler/plan/run/forwarder coverage).


kid7st (commented):
Review of the whole branch (checked out locally; `lint` + `typecheck` + `vitest` green, 1201 passing). The two-plane split, the "403 is not 404" restore rule, the parsed dow remapping and the deterministic zip are all the right calls. Nine findings, ordered by severity — file:line refer to the branch head.

### P1 — a `selfSchedule`-only deployment publishes the unauthenticated builtin `POST /invoke`
`src/deploy/agentcore/plan.ts:445,823` — `needsFunctionUrl = routeChannels.length > 0 || selfSchedule`, while `routesFor()` (`src/cli/serve.ts`) mounts the builtin `POST /invoke` exactly when there are **no** channels. selfSchedule + no channels satisfies both: the public Function URL is created, the forwarder forwards `event.rawPath` verbatim, and `/invoke` is on the internet — the precise hole the comment above that line says it is closing (an unguessable Lambda URL host is not authentication).

Changing the URL condition is not the fix (the wake-alarm callback genuinely needs a public URL, the container has no AWS credentials). The fix is to not mount the builtin `/invoke` under `FASTAGENT_AGENTCORE=1`: the `invoke` envelope already is that plane, behind IAM.

### P2 — a shrinking auth seed reassembles into a corrupt `auth.json`
`src/deploy/agentcore/run.ts:90` — `paramsFileContent` emits only the chunks it needs, and `aws cloudformation deploy` keeps `UsePreviousValue` for every parameter absent from `--parameter-overrides`. So a redeploy going 3 chunks → 2 leaves the previous `FastagentAuthSeed3` in place, and `collectAuthSeed` concatenates until the first empty var — tail of stale base64. Same retention hits a switch from OAuth to an API key (the old seed is still materialized on a wiped mount).

Fix: always emit all `AUTH_SEED_MAX_CHUNKS` parameters, empty string for the unused ones.

### P3 — the external-clock claim records `now()`, not the slot
`src/schedule/scheduler.ts:139` — the skip test is `lastFired >= slot` but the claim writes `now()`. Once delivery lag exceeds the cron interval (a cold Lambda, a retry), the *next distinct* slot is judged a duplicate and dropped silently. One line: write `(slot ?? now()).toISOString()` — a genuine duplicate still trips `>=`.

### P4 — the checkpoint puts the ingress secret on argv
`src/deploy/agentcore/run.ts:380` — `invoke-agent-runtime --payload {"kind":"checkpoint","auth":…}` places the shared secret in the operator machine's process listing, which this module's own header rules out ("`--parameter-overrides` on argv would put the values in the process listing"). `paramsDir` (0600) already exists — use `--payload file://…`.

### P5 — 12 h presigned URLs outlive the credentials that signed them
`src/deploy/agentcore/plan.ts:312` + `src/channels/agentcore-state.ts` — the URLs are signed with the Lambda execution role's temporary credentials and die with them, while the PUT may happen long after the envelope (a background turn settling). `save()` only logs on failure, and the very next thing is `/ping` reporting Healthy → reclaim. That is a silent state loss on the exact path this module exists to protect. At minimum classify an expired-signature failure explicitly in the log line; consider a shorter, re-minted window.

### P6 — non-constant-time secret comparison
`src/channels/agentcore.ts:167` — `envelope.auth === ingressSecret` on a publicly reachable path. `timingSafeEqual` costs nothing here.

### P7 — module-level mutable singletons
`setWakeupsSink` (`src/schedule/wakeups.ts`) turns a pure store module into one with a global callback, and `busy.ts` is a process-wide counter; both leak across tests and across any second mount. Passing the sink through the scheduler/adapter options would match the seam style used everywhere else in this repo. Not a merge blocker.

### P8 — regex-scraping the checkpoint result
`src/deploy/agentcore/run.ts:396` — `/"written"\s*:\s*true/` over stdout, in a PR whose description argues for executed tests rather than substring assertions. `JSON.parse` with a fallback is sturdier.

### P9 — `MAX_WEBHOOK_BODY_BYTES` enforces nothing
`src/channels/agentcore.ts:49` — the name reads like a limit, but its only consumer is a `console.error` note in `deploy.ts`; the adapter caps on `MAX_ENVELOPE_BYTES`. Either enforce it or rename it to say it is advisory.

P1–P3 are worth fixing before merge (all small). P4–P6 are cheap follow-ons in the same pass; P7–P9 can be separate.

kid7st (approved):
LGTM

PR #294, opened 2026-07-30
feat(serve): a bind address — containers keep all interfaces, a desktop gets loopback

### LINKED ISSUE #290: `fastagent serve` needs a bind address: containers want all interfaces, a desktop wants loopback
## What

A bind address for `fastagent serve` — config or flag — defaulting to today's behaviour.

## Why

The current bind is deliberate and says so:

```ts
// The serve binds ALL interfaces (containers require it), so /control/* is LAN-reachable
// with the bearer token as the only protection — the tunnel and deploy paths warn loudly,
// and the LAN path must not be the silent third way past the local trust story.
log.warn(
  "[fastagent] the port binds all interfaces: /control/* is reachable on your LAN, protected only by " +
    "the bearer token — firewall the port or wrap it for real exposure (docs: design §14)",
);
```

Nothing there is wrong. Containers do require it, and the warning is honest — which is exactly why
this is a request for an option rather than a change of default: the two deployments want opposite
things, and neither should have to be the other.

For a container, `/control/*` on all interfaces is the point. For a desktop app driving a local
agent, it is a session control plane — `steer`, `abort`, `set_model`, the transcript — sitting on
the LAN behind one bearer token, in a process the user started by opening a window. They did not
ask to serve anything, and the warning goes to a log they never see.

`0.0.0.0` and `127.0.0.1` are one parameter apart in `node:http`, which is what makes this worth
asking rather than working around.

## Why not a Unix socket instead

That would be the stronger answer — filesystem permissions rather than a token, no port, nothing to
firewall — and `serveNode` could listen on one, since `node:http` takes a path. It does not work for
this consumer, for a reason outside fastagent:

- a webview's `fetch` / `EventSource` has no URL form for a Unix socket;
- routing it through a native custom-protocol handler does not help either — Tauri's
  `UriSchemeResponder` takes `http::Response<Cow<'static, [u8]>>`, a complete body, so SSE cannot
  survive the trip;
- the remaining path puts the event stream on the app's JS↔native IPC bridge, which is where large
  streaming payloads compete with the paint thread.

So loopback is not the ideal, it is the strongest thing a webview can speak directly. Worth noting
that against a same-user process, a 0600 socket and a 0600 token file are equivalent anyway — the
gap that matters here is LAN reachability, and a bind address closes exactly that.

A Unix socket option would still be welcome for consumers that can use one; it is just not a
substitute for this.

## Shape

```
fastagent serve --bind 127.0.0.1
```

or the config equivalent, with `0.0.0.0` remaining the default so no deployment changes. The value
belongs in `control.json` next to the token if it is not already implied by `url`, so a client
connects to the address that was actually bound rather than assuming one.

## Context

duang.ai (see #284–#289) is moving from spawning `pi --mode rpc` to attaching to `fastagent serve`
directly, with no proxy in between — its current local runner binds loopback, and connecting
directly would otherwise trade that away silently.

Happy to send the PR.

kid7st:
Closes #290.

## Why

The serve binds all interfaces because containers require it, which leaves `/control/*` (steer/abort/set_model, the transcript) LAN-reachable behind one bearer token — honest, and warned about, but wrong for a desktop app that started the process by opening a window. The two deployments want opposite things, so this adds an option rather than changing the default.

Unset behaves exactly as today.

```txt
--bind <addr>  >  fastagent.config.* http.host  >  all interfaces
```

## One reading of a bind address, as the six questions it actually is

`src/bind.ts` exists because "is this local?" is not one question, and conflating any two of them produces a silent failure:

| | asks | why it is its own question |
|---|---|---|
| `isBindAddress` | can this be bound? | an IP literal or `localhost` — no other DNS name, since a bind address must be an address of *this* machine |
| `bindAddress` | as an ADDRESS, not a name | `localhost` → `127.0.0.1`, applied where a value **enters** |
| `classifyBind` | how far does it reach? | `wildcard` / `loopback` / `specific`; all of `127/8` is loopback |
| `answersLocalhost` | does a dial of the **name** `localhost` land here? | **not** the same as reach — `127.0.0.2` is loopback and a tunnel still cannot reach it |
| `bindLabel` | how does a message name it? | one rendering, so a reader never meets the same bind described two ways |
| `clientHost` | what does a client dial? | wildcard → loopback, otherwise itself; IPv6 gets its brackets |

The flag, `http.host` validation, `serveNode`, the ready lines, `control.json` and the deploy pre-flight all read a bind through this one module.

**A name never travels past the boundary.** Accepting `localhost` and passing it on are different things: `server.listen` would hand the name to `dns.lookup`, which picks one of `127.0.0.1`/`::1` by rules this module does not control — so what got bound would be unknown downstream, while `control.json` and the copyable curl carried a name for the client to resolve *again*, possibly to the other one. Resolving it on the way in removes both.

## Two edges that would otherwise fail silently

- **`--tunnel` + a bind `localhost` does not resolve to** (`192.168.1.5`, or even `127.0.0.2`): cloudflared dials by name, so the tunnel would come up and 502 everything. Refused before anything binds — in `dev` from the flags alone before a worker exists, then again once config is loaded, since `http.host` can carry the address. Exit 2 for a flag combination, exit 1 when the value came from config, and the message names whichever source it was: under `config` there is no flag to drop.
- **A non-wildcard `http.host` travels into the deployed image**, where it makes the container unreachable (loopback) or unbindable (an address that does not exist there): `deploy` warns, `deploy --run` gates, and the docs point local-only binds at `--bind`.

## Discovery and the warning

`control.json` records the address a client should dial, so a consumer never assumes one. The LAN warning goes silent on a loopback bind — there is no LAN reach left to state — and names the actual bind otherwise.

## The startup lines are one message

The bind report and the `try it` curl are two halves of one thing — where it bound, and the command that proves it. They were two call sites that happened to agree, and `--bind` updated only the first, so a LAN bind printed a curl to `localhost`: the very address that bind stops answering. They come from one function now (`readyAddressLines`), so neither can change without the other in view.

## Verification

`npm run lint && npm run typecheck && npm test` — **1251 tests**.

New coverage: `test/bind.test.ts` (parser/classifier agreement across every spelling, the name→address boundary, one label per bind), `serveNode` binding only what it was given, `parseBind`/`assertTunnelBindable` exit codes *and* the message naming its source, `http.host` config validation, the deploy gate (unconditional), the LAN warning's content, and the ready lines agreeing on one dialable address.

End to end:

```
$ fastagent start . --bind localhost
[fastagent] http host on 127.0.0.1:8905
[fastagent] try it: curl -s 127.0.0.1:8905/invoke …          → 200

$ fastagent start ./not-an-agent --bind banana                → exit 2  (as --port)
```

## Review

Three rounds. The `--tunnel`/`http.host` edges and the module split are from the first design; the review added the name→address boundary, `bindLabel`, the one-function ready lines, `--bind` parsing before the placement check (a usage error must not depend on the directory being an agent), and turned four assertions that could not fail into ones that can. Rebased onto `main` after the placement refactor, pi 0.83 and the AgentCore target landed underneath.

### DIFF
--- a/src/bind.ts
+++ b/src/bind.ts
@@ -14,11 +14,25 @@ function normalize(host: string): string {
     .replace(/^::ffff:/, "");
 }
 
-/** A bindable host: an IP literal (v4/v6, brackets optional) or "localhost". No DNS names — a bind
- *  address must be an address of THIS machine, and resolving one at parse time is not our job. */
+/** A bindable host: an IP literal (v4/v6, brackets optional) or "localhost". No other DNS name — a
+ *  bind address must be an address of THIS machine, and resolving one is not this module's job. */
 export function isBindAddress(host: string): boolean {
   const h = normalize(host);
   return h === "localhost" || isIP(h) !== 0;
+}
+
+/**
+ * The ADDRESS form of an accepted bind value: `localhost` becomes `127.0.0.1`, everything else is
+ * already an address. Applied where a value ENTERS (the flag, `http.host`), so nothing downstream ever
+ * holds a name — not deferring the resolution but removing it, which is the module's whole point.
+ *
+ * Two things go wrong otherwise, and both are silent. `server.listen` hands the name to `dns.lookup`,
+ * which picks ONE of 127.0.0.1/::1 by rules we do not control — so what got bound is unknown here. And
+ * `clientHost` would then write that NAME into control.json and the copyable curl, where a consumer
+ * resolves it again, possibly to the other one.
+ */
+export function bindAddress(host: string): string {
+  return normalize(host) === "localhost" ? "127.0.0.1" : host;
 }
 
 /**
@@ -44,6 +58,14 @@ export function answersLocalhost(host: string | undefined): boolean {
   if (classifyBind(host) === "wildcard") return true;
   const h = normalize(host as string);
   return h === "localhost" || h === "127.0.0.1" || h === "::1";
+}
+
+/** How to NAME a bind in a message: the wildcard is every interface, so calling it one address would
+ *  understate it; anything else is dialable as itself. THE renderer — the ready lines, the
+ *  already-in-use refusal and the generic bind failure all read a bind through this one, so a reader
+ *  never sees the same bind described two ways. */
+export function bindLabel(host: string | undefined, port: number): string {
+  return classifyBind(host) === "wildcard" ? `port ${port}` : `${clientHost(host)}:${port}`;
 }
 
 /** The address a local client should dial for a serve bound to `host` (control.json, the ready log). */
--- a/src/bind.ts
+++ b/src/bind.ts
@@ -1,0 +1,54 @@
+/**
+ * The ONE reading of a bind address, shared by everything that parses, binds, warns about, or ships
+ * one (the CLI flag, `http.host` validation, the Node host, the deploy pre-flight). Engine- and
+ * host-neutral on purpose: a bind address is a plain value, and two parsers of it would disagree.
+ */
+import { isIP } from "node:net";
+
+/** Lowercase, unbracketed, IPv4-mapped IPv6 reduced to its IPv4 form — the form the checks below read.
+ *  Brackets come off only as a PAIR: a half-bracketed `[::1` must stay invalid, not become an address. */
+function normalize(host: string): string {
+  return host
+    .toLowerCase()
+    .replace(/^\[(.+)]$/, "$1")
+    .replace(/^::ffff:/, "");
+}
+
+/** A bindable host: an IP literal (v4/v6, brackets optional) or "localhost". No DNS names — a bind
+ *  address must be an address of THIS machine, and resolving one at parse time is not our job. */
+export function isBindAddress(host: string): boolean {
+  const h = normalize(host);
+  return h === "localhost" || isIP(h) !== 0;
+}
+
+/**
+ * How far a bind address reaches: `wildcard` (unset or all-interfaces) reaches every interface and
+ * answers on loopback too; `loopback` is this machine only; `specific` is one interface, reachable
+ * only as itself. Loopback covers the whole reserved range (127/8, ::1) — a `127.0.0.2` bind is no
+ * more LAN-reachable than `127.0.0.1`.
+ */
+export function classifyBind(host: string | undefined): "wildcard" | "loopback" | "specific" {
+  if (host === undefined) return "wildcard";
+  const h = normalize(host);
+  if (h === "0.0.0.0" || h === "::" || h === "::0") return "wildcard";
+  if (h === "localhost" || h === "::1" || /^127\.\d+\.\d+\.\d+$/.test(h)) return "loopback";
+  return "specific";
+}
+
+/**
+ * Does a serve bound to `host` answer a dial of the NAME `localhost`? Only the addresses that name
+ * resolves to (127.0.0.1 / ::1) and a wildcard bind do — `127.0.0.2` is loopback yet unreachable that
+ * way. cloudflared dials by name, so `--tunnel` needs this question, not `classifyBind`'s reach.
+ */
+export function answersLocalhost(host: string | undefined): boolean {
+  if (classifyBind(host) === "wildcard") return true;
+  const h = normalize(host as string);
+  return h === "localhost" || h === "127.0.0.1" || h === "::1";
+}
+
+/** The address a local client should dial for a serve bound to `host` (control.json, the ready log). */
+export function clientHost(host: string | undefined): string {
+  if (classifyBind(host) === "wildcard") return "127.0.0.1";
+  // biome-ignore lint/style/noNonNullAssertion: only a wildcard bind leaves host undefined
+  return host!.includes(":") && !host!.startsWith("[") ? `[${host}]` : host!;
+}
--- a/src/cli/commands/start.ts
+++ b/src/cli/commands/start.ts
@@ -17,6 +17,7 @@ import { setWakeupsSink } from "../../schedule/wakeups.ts";
 import { logAgentLoop } from "../../observe.ts";
 import { installProxyFetch } from "../../proxy.ts";
 import { exists } from "../../paths.ts";
+import { bindAddress } from "../../bind.ts";
 import { failStartup, placementOrExit } from "../fail.ts";
 import {
   assertTunnelBindable,
@@ -45,8 +46,8 @@ export async function runStart(dirArg: string, opts: StartOptions): Promise<void
   // Flag validation first: a bad --port is a USAGE error (exit 2), and reporting it must not depend on
   // the directory being an agent (which is a runtime/environment failure, exit 1).
   const portFlag = parsePort(opts.port, "--port", "flag");
-  const placement = placementOrExit(dir);
   const bindFlag = parseBind(opts.bind);
+  const placement = placementOrExit(dir);
   setLogLevel("info"); // production posture: info+, the debug turn trace (and its end-user content) gated out
   loadDotEnv(placement.agentDir);
   installProxyFetch();
@@ -132,7 +133,10 @@ export async function runStart(dirArg: string, opts: StartOptions): Promise<void
   const routed = await routesFor(agentDir, traced, stateRoot, sessionControl, { builtinInvoke: !agentcore }).catch(
     failStartup,
   );
-  const host = bindFlag ?? config.http?.host;
+  // `http.host` enters here the way the flag enters `parseBind` — through `bindAddress`, so a
+  // configured `localhost` is an ADDRESS by the time anything binds, renders or dials it.
+  const configured = config.http?.host;
+  const host = bindFlag ?? (configured === undefined ? undefined : bindAddress(configured));
   assertTunnelBindable(host, opts.tunnel ?? false, bindFlag ? "flag" : "config");
   const withControl = mountSessionControl(routed.routes, sessionControl, stateRoot, {
     tunnel: opts.tunnel ?? false,
--- a/src/cli/commands/start.ts
+++ b/src/cli/commands/start.ts
@@ -18,11 +18,20 @@ import { logAgentLoop } from "../../observe.ts";
 import { installProxyFetch } from "../../proxy.ts";
 import { exists } from "../../paths.ts";
 import { failStartup, placementOrExit } from "../fail.ts";
-import { maybeTunnel, mountAgentcore, mountSessionControl, routesFor, serve, startSchedules } from "../serve.ts";
-import { parsePort, reportAuth, reportLine, resolveFirstRunModel, reportWorkspaceHint } from "../shared.ts";
+import {
+  assertTunnelBindable,
+  maybeTunnel,
+  mountAgentcore,
+  mountSessionControl,
+  routesFor,
+  serve,
+  startSchedules,
+} from "../serve.ts";
+import { parseBind, parsePort, reportAuth, reportLine, resolveFirstRunModel, reportWorkspaceHint } from "../shared.ts";
 
 export interface StartOptions {
   port?: string;
+  bind?: string;
   model?: string;
   sessionsDir?: string;
   authPath?: string;
@@ -37,6 +46,7 @@ export async function runStart(dirArg: string, opts: StartOptions): Promise<void
   // the directory being an agent (which is a runtime/environment failure, exit 1).
   const portFlag = parsePort(opts.port, "--port", "flag");
   const placement = placementOrExit(dir);
+  const bindFlag = parseBind(opts.bind);
   setLogLevel("info"); // production posture: info+, the debug turn trace (and its end-user content) gated out
   loadDotEnv(placement.agentDir);
   installProxyFetch();
@@ -122,9 +132,12 @@ export async function runStart(dirArg: string, opts: StartOptions): Promise<void
   const routed = await routesFor(agentDir, traced, stateRoot, sessionControl, { builtinInvoke: !agentcore }).catch(
     failStartup,
   );
+  const host = bindFlag ?? config.http?.host;
+  assertTunnelBindable(host, opts.tunnel ?? false, bindFlag ? "flag" : "config");
   const withControl = mountSessionControl(routed.routes, sessionControl, stateRoot, {
     tunnel: opts.tunnel ?? false,
     agent: traced,
+    host,
   });
   // AgentCore + selfSchedule: register the wake-ALARM sink BEFORE the scheduler starts — the boot
   // wake pump may advance a recurring entry (a store save) and that save must already re-arm its
@@ -162,7 +175,7 @@ export async function runStart(dirArg: string, opts: StartOptions): Promise<void
   }
   serve(
     { ...routed, routes },
-    portFlag ?? parsePort(process.env.PORT, "PORT env", "env") ?? config.http?.port ?? 8787,
+    { port: portFlag ?? parsePort(process.env.PORT, "PORT env", "env") ?? config.http?.port ?? 8787, host },
     (p) => {
       withControl.announce(p);
       maybeTunnel(agentDir, routed.routeChannels, p, opts.tunnel ?? false, stateRoot);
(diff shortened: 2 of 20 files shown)


kid7st (commented):
Reviewed the whole delta and ran it. The design is right and the tests are unusually honest — `bind.test.ts` asserting that the parser and the classifier agree on every spelling, and `host-node` asserting the socket behaviour rather than the option, are the two that would have caught a real regression. The two-edges section is the good part: `--tunnel` + a non-localhost bind and `http.host` travelling into an image are exactly the failures this feature would otherwise create silently.

One thing to fix before merge, three smaller, and a rebase.

---

### 1. The "try it" hint tells you to curl an address the bind just closed

`clientHost`'s own docstring says it is for "(control.json, **the ready log**)" — the ready log's bind line uses it, the line two below does not:

```js
`[fastagent] try it: curl -s localhost:${boundPort}/invoke …`
```

Run against this branch, `start . --bind 192.168.1.78 --port 8899`:

```
INFO [fastagent] http host on 192.168.1.78:8899                      ← right
INFO [fastagent] try it: curl -s localhost:8899/invoke …             ← cannot connect

$ curl localhost:8899      → 000   (refused)
$ curl 192.168.1.78:8899   → 400   (answers)
```

So the feature's own success path prints a command that fails, two lines under a correct report of why. `clientHost(host)` is already in scope at that call site.

No test covers the ready log, which is why it slipped — the bind line and the curl line are one message to a reader, and only one of them was updated.

### 2. The generic bind-failure message builds an address by hand

The `EADDRINUSE` branch renders through `classifyBind`/`clientHost`; the fallback four lines below concatenates:

```js
failStartup(new Error(`cannot bind http channel on ${host ?? ""}:${port}: …`))
```

| `host` | concatenated | via the helpers |
|---|---|---|
| unset | `on :8787` | `port 8787` |
| `::1` | `on ::1:8787` | `[::1]:8787` |
| `192.168.1.5` | `on 192.168.1.5:8787` | `192.168.1.5:8787` |

The IPv6 form is unreadable and the wildcard form reads as an explicit bind of nothing. This PR's premise is one reading of a bind address; errors are where a reader needs it most.

### 3. `assertTunnelBindable` takes `source` but writes only the flag story

`source` picks the exit code (2 vs 1) correctly, and then the message says `bind 0.0.0.0 (or 127.0.0.1), or drop --tunnel` either way. Under `source: "config"` there is no `--bind` to change and no flag to drop — the fix is editing `http.host`. `preflightDeploy` gets this right in the same PR (`fastagent.config.ts sets http.host: "…"`), so it is worth matching.

### 4. The LAN probe can pass for the wrong reason, or hang

```js
const lan = Object.values(networkInterfaces()).flat().find((i) => i?.family === "IPv4" && !i.internal)?.address;
if (lan) await expect(fetch(`http://${lan}:${port}/`)).rejects.toThrow();
```

The first non-internal IPv4 is often a VPN/docker/utun address, where a rejection can come from routing rather than from the bind — the assertion passes without proving anything. And on a **filtered** (drop, not reject) interface `fetch` has no timeout, so it hangs until vitest kills the file. This suite just removed its one reliable flake for the same class of reason, so it is worth not adding one: either bound it with `AbortSignal.timeout(…)`, or assert the negative against `127.0.0.2` — a second loopback alias is deterministic, needs no network, and is the sharper claim anyway (loopback by reach, still not the bound address).

### 5. `AGENTS.md` repo map

`src/bind.ts` is a new top-level neutral module and the map enumerates `src/`. It is missing — and "the ONE reading of a bind address" is exactly what someone greps the map for.

---

### Rebase

Eight files conflict with current `main` (`serve.ts`, `preflight.ts`, `dev.ts`, `start.ts`, `program.ts`, `configuration.md`, two tests) — the placement refactor and the pi 0.83 work landed underneath. All mechanical, but the branch cannot merge as it stands and the "1051 tests" figure predates them.

kid7st:
Pushed the review follow-ups onto this branch (`8627abc`), rebased onto `main`.

**The `try it` hint is fixed by removing the failure mode, not by patching the line.** The bind report and the curl are one message to a reader — where it bound, and the command that proves it — so they now come from one function (`readyAddressLines`) instead of two call sites that happened to agree. `serve()` shuts down only on a signal, so it was untestable in-process; the pure function is not, and the test asserts the property that made splitting them a bug (every ready line names the same dialable address, never `localhost`). Mutation-checked: putting `localhost` back fails it.

End to end on a LAN bind, where the printed command used to refuse:

```
[fastagent] http host on 172.20.10.2:8903
[fastagent] try it: curl -s 172.20.10.2:8903/invoke …   → HTTP 200
```

The other four:

- **Generic bind failure** now renders through `classifyBind`/`clientHost` like the `EADDRINUSE` branch above it (`:::8787` → `[::1]:8787`; bare `:8787` → `port 8787`).
- **`assertTunnelBindable`** names its source: under `config` there is no `--bind` to change and no flag to drop, so it points at `http.host` — matching what `preflightDeploy` already said about the same value.
- **The bind probe** asserts against a second loopback alias instead of the first non-internal IPv4. Deterministic, no network, no unbounded `fetch`, and the sharper claim: loopback by reach, still not the address bound.
- **`AGENTS.md`** gains the `src/bind.ts` entry.

**Rebase notes**, since two conflicts were more than mechanical:

- `dev`/`start` now resolve a placement before loading `.env` (#280), so `bindFlag` is parsed alongside it rather than in place of the old `loadDotEnv(dir)`.
- One deploy pre-flight test on this branch asserted the `config.agentDir` kit layout, which #280 deleted. Dropped — main's AGENT-DIR-based test supersedes it.

`npm run lint` + `npm run typecheck` clean, **1249** tests. Design and the two-edges reasoning are unchanged; nothing here touches `bind.ts` itself.

src/cli/commands/start.ts
```diff
@@ -37,6 +46,7 @@ export async function runStart(dirArg: string, opts: StartOptions): Promise<void
   // the directory being an agent (which is a runtime/environment failure, exit 1).
   const portFlag = parsePort(opts.port, "--port", "flag");
   const placement = placementOrExit(dir);
+  const bindFlag = parseBind(opts.bind);
```
kid7st, line 49: `parseBind` runs after `placementOrExit`, contradicting the comment three lines up ("Flag validation first: a bad --port is a USAGE error (exit 2), and reporting it must not depend on the directory being an agent"). `fastagent start /not-an-agent --bind banana` exits 1 with a placement error, while the identical mistake on `--port` exits 2 — same class of error, two exit codes, decided by which flag you mistyped. `dev`'s `serveOnce` gets this right (`parsePort` then `parseBind` then placement). Move `const bindFlag = parseBind(opts.bind)` up next to `parsePort`.

src/cli/serve.ts
```diff
@@ -305,9 +357,20 @@ export function serve(surface: ServingSurface, port: number, onListening?: (boun
...
+        // With a bind address the port is only taken ON THAT interface, so moving the bind is as valid
+        // a fix as moving the port.
+        failStartup(
+          new Error(
+            `${classifyBind(host) === "wildcard" ? `port ${port}` : `${clientHost(host)}:${port}`} is already ` +
+              `in use; choose another with --port${classifyBind(host) === "wildcard" ? "" : " or --bind"}`,
+          ),
+        );
+      }
+      // Through the same renderer as the EADDRINUSE branch above: hand-concatenating gives `:::8787`
```
kid7st: The comment says "Through the same renderer as the EADDRINUSE branch above" — there is no renderer: the ternary `classifyBind(host) === "wildcard" ? \`port ${port}\` : \`${clientHost(host)}:${port}\`` is copy-pasted between the two branches (and `readyAddressLines` renders the same concept a third way, `:${port} (all interfaces)`). The comment asserts the exact property the code doesn't have, which is worse than no comment: the next edit to one branch reads as covered. Extract `bindLabel(host, port)` next to `readyAddressLines` and route all three through it, or drop the claim.

src/bind.ts
```diff
@@ -0,0 +1,54 @@
...
+/** Lowercase, unbracketed, IPv4-mapped IPv6 reduced to its IPv4 form — the form the checks below read.
+ *  Brackets come off only as a PAIR: a half-bracketed `[::1` must stay invalid, not become an address. */
+function normalize(host: string): string {
+  return host
+    .toLowerCase()
+    .replace(/^\[(.+)]$/, "$1")
+    .replace(/^::ffff:/, "");
+}
+
+/** A bindable host: an IP literal (v4/v6, brackets optional) or "localhost". No DNS names — a bind
```
kid7st: `isBindAddress` accepts `"localhost"` while its own doc says "No DNS names — a bind address must be an address of THIS machine, and resolving one at parse time is not our job". `localhost` is precisely a DNS name, and declining to resolve it doesn't remove the resolution — it defers it to `server.listen({host})`, where Node's `dns.lookup` picks **one** address (127.0.0.1 or ::1, resolver-dependent). Downstream, `clientHost("localhost")` then writes a NAME into `control.json` and into the copy-paste curl — the "dial a name the bind may not answer" ambiguity this module exists to remove, reintroduced through the one value the parser lets through. Normalize `localhost` → `127.0.0.1` in `parseBind` and in the `http.host` validator so `bind.ts` really is address-only past the boundary.

test/cli-serve.test.ts
```diff
@@ -87,3 +87,90 @@ describe("mountAgentcore", () => {
...
+    // They are one message: the first says where it bound, the second is the command a reader copies.
+    // Only the first was updated when --bind landed, so `--bind 192.168.1.5` printed a curl to
+    // localhost — the very address that bind stops answering. They come from one function now; this
+    // pins the property that made splitting them a bug.
+    const { readyAddressLines } = await import("../src/cli/serve.ts");
+    for (const host of [undefined, "0.0.0.0", "127.0.0.1", "192.168.1.5", "::1"]) {
+      const [bindLine, tryLine] = readyAddressLines(host, 8899, true);
+      const dial = tryLine!.match(/curl -s (\S+?)\/invoke/)![1]!;
+      expect(dial, String(host)).toContain(":8899");
+      expect(dial, String(host)).not.toContain("localhost"); // never a name the bind may not answer
```
kid7st, line 170: `expect(dial).not.toContain("localhost")` is the property the test says it pins ("never a name the bind may not answer"), but the host list — `undefined, 0.0.0.0, 127.0.0.1, 192.168.1.5, ::1` — excludes the only accepted input that violates it: `--bind localhost` renders `localhost:8899` in both lines. The assertion passes because the one interesting case was left out. Add `"localhost"` to the loop (it will fail until the parser normalizes it, which is the point).

test/deploy-preflight.test.ts
```diff
@@ -57,6 +57,22 @@ describe("deploy/preflight: the host-neutral pre-flight", () => {
...
 
+  it("http.host travels into the image: gates --run, warns in plan mode, wildcard is silent", async () => {
+    const dir = await workspace();
+    const cfg = (host?: string): FastagentConfig => ({ model: "openai/gpt-4o-mini", http: host ? { host } : {} });
+    const gated = await call(dir, cfg("127.0.0.1"), { run: true });
+    expect(gated.ok).toBe(false);
+    if (!gated.ok) expect(gated.gate).toMatch(/nothing outside the container can reach the serve/);
+    // A one-interface address doesn't even exist in the container — different consequence, same gate.
+    const gatedSpecific = await call(dir, cfg("192.168.1.5"), { run: true });
+    if (!gatedSpecific.ok) expect(gatedSpecific.gate).toMatch(/fails to bind at start/);
```
kid7st, line 69: The specific-address gate is asserted only inside `if (!gatedSpecific.ok)`. If the gate stops firing for `192.168.1.5` — the case whose consequence differs (container fails to bind, vs unreachable) — `gatedSpecific.ok` is true, the body never runs, and the test is green. The loopback case above guards with an unconditional `expect(gated.ok).toBe(false)` first; do the same here.

test/control-http.test.ts
```diff
@@ -511,6 +511,32 @@ describe("session control over HTTP (Phase 3)", () => {
...
+      const url = (host: string | undefined, stateRoot: string) => {
+        mountSessionControl({}, control, stateRoot, { host }).announce(9000);
+        return readFile(join(stateRoot, "control.json"), "utf8").then((s) => (JSON.parse(s) as { url: string }).url);
+      };
+      expect(await url("127.0.0.1", join(root, "a"))).toBe("http://127.0.0.1:9000");
+      expect(warn).not.toHaveBeenCalled(); // loopback is not LAN-reachable — nothing to warn about
+      // A specific non-wildcard bind is only reachable as itself: the client must dial that address.
+      expect(await url("192.168.1.5", join(root, "b"))).toBe("http://192.168.1.5:9000");
+      expect(await url(undefined, join(root, "c"))).toBe("http://127.0.0.1:9000"); // wildcard accepts loopback
+      expect(warn).toHaveBeenCalledTimes(2);
```
kid7st, line 537: The test asserts only the warn COUNT, so the behavior the PR advertises — "the LAN warning … names the actual bind otherwise" — is unverified: the new `${options.host} (off this machine)` interpolation could render `undefined` and this stays green. Assert the message content for the `192.168.1.5` case (`expect(warn.mock.calls.at(-1)?.[0]).toMatch(/192\.168\.1\.5/)`).

kid7st (commented):
Thesis: `bind.ts` is the single reading of a bind address (parse / reach / localhost-dialability / URL form), and everything that binds, warns, dials or ships one reads it through that module.

Finding that has no line in the diff:

**`src/channels/control.ts:15-16` — the security note of the control plane still states the pre-PR fact.** It reads "the serve binds ALL interfaces (containers require it), so the port is LAN-reachable and the mount warns accordingly". After this PR that is a *default*, and the mount is deliberately silent on a loopback bind. This is the module-level statement of the control plane's trust story — `config.ts`, `configuration.md` and the warning itself were all updated, this one was missed, so the file that owns the security claim is now the one place asserting the wrong one. Mirror the conditional wording ("binds all interfaces by default; `--bind 127.0.0.1` closes the LAN reach").

kid7st:
Pushed `a87924c`; the six inline threads are resolved.

**Three of the seven were mine** — which is why this pass was worth running on the final state rather than merging on a green CI:

- `start` parsed `--bind` *after* `placementOrExit`, contradicting the comment three lines above it. `start /not-an-agent --bind banana` exited **1** where the same mistake on `--port` exits **2**. That came from my rebase resolution; `dev` had the order right, so nothing in the diff looked wrong side by side.
- I wrote "through the same renderer as the EADDRINUSE branch" — and there was no renderer. The same ternary was pasted twice, with a third spelling inside `readyAddressLines`. `bindLabel` is that renderer now, so one bind is never described two ways.
- My ready-lines test asserted `not.toContain("localhost")` over a host list that **excluded `localhost`**. It passed because the only input that could break it was never tried.

**The sharpest finding is about the original design**, though, and it is a good one: `isBindAddress` accepted `localhost` while its own comment said "no DNS names — resolving one at parse time is not our job". Not resolving is not the same as removing:

- `server.listen` hands the name to `dns.lookup`, which picks **one** of `127.0.0.1`/`::1` by rules this module does not control — so what got bound was unknown to everything downstream;
- `clientHost("localhost")` then wrote a **name** into `control.json` and the copyable curl, for a client to resolve again, possibly to the other one.

`bindAddress` resolves it where the value ENTERS (the flag, `http.host`). That is the module's own premise carried through: not deferring the resolution, removing it. Verified:

```
$ fastagent start . --bind localhost
[fastagent] http host on 127.0.0.1:8905
[fastagent] try it: curl -s 127.0.0.1:8905/invoke …

$ fastagent start ./not-an-agent --bind banana   → exit 2   (--port: exit 2)
```

Two more assertions that could not fail: the specific-address deploy gate was wrapped in `if (!ok)` (green if the gate stopped firing), and the LAN-warning test counted calls without reading them — so the address rendering as `undefined` would have passed, while this PR claims the warning "names the actual bind". Both assert unconditionally now, and `bind.test.ts` pins `bindAddress` and `bindLabel`.

`npm run lint` + `npm run typecheck` clean, **1251** tests.

---

One thing left for whoever merges: **the PR description becomes the squash commit message**, and it still says "1051 tests" and describes the pre-rebase state — no mention of the ready-line fix, `bindAddress`, or `bindLabel`. Worth a pass before the button.

PR #300, opened 2026-08-01
feat(engines/pi): the default coding tools reach the machine through ExecutionEnv

kid7st:
## What

The default coding tools (`read`/`bash`/`edit`/`write`) now come from **pi-agent-core**, which takes the `ExecutionEnv` as the turn's tool **context** (pi 0.83), instead of **pi-coding-agent**, whose identical four are wired to `node:fs` directly.

## Why

`env` was a seam that governed everything **except** the tools that actually touch the machine. The docs had to keep saying so:

> injecting `env` alone does not isolate a directory agent

Now the tools go through it, so a custom env narrows where the agent reads, writes and shells. It is still not a full sandbox — author-written `tools/` are code and can import anything — and the docs now say **that** instead of hedging about the built-ins.

**And that is all it buys.** An earlier draft of this description claimed it also decoupled the serving path from pi's TUI stack; that was false — `definition.ts`, `models.ts` and now `read-image.ts` import pi-coding-agent unconditionally. The env seam is the whole benefit.

## Why it is safe

The swap only holds while the **model-facing surface** is identical, so `test/tools-parity.test.ts` asserts it against both implementations rather than trusting it — names, descriptions, parameter schemas, and results on the paths an agent takes, including **failure** and **image** paths.

Images are where they do NOT match for free, and the review caught it: core's `read` does nothing with an image unless a processor is **injected**. Measured on a 5.6 MB PNG, the unfixed version sent **7.48 MB** of raw base64 where the old tool sent **3.48 MB** of resized JPEG — double the payload to the provider — with no dimension note for the model's coordinate math, and a bmp dropped outright while the description asserted identical above still advertised it.

`processImage` is not exported and not reachable (no deep paths in the package's `exports`), so `src/engines/pi/read-image.ts` rebuilds it from the halves that are — `convertToPng`, `resizeImage`, `formatDimensionNote`. Twenty lines of restated upstream logic, byte-identical on both paths, and covered by the parity test (mutation-checked: removing the processor fails it), which is what keeps the duplication from drifting silently.

It also **pins the two places they do differ**, because saying where parity stops is worth more than a test that avoids looking:

| Case | core | pi-coding-agent |
|---|---|---|
| `read` a missing file | `ENOENT … open …` | `ENOENT … access …` |
| `bash` failing with **no** output | `Command exited with code 3` | `(no output)\n\nCommand exited with code 3` |

With output, the failure results are byte-identical — which is the case that matters. The test fails the day upstream lets anything else drift, which is exactly when this choice needs revisiting.

## Mechanics

- `piDefaultTools()` loses its `cwd` argument — the tools have no root of their own; the env's cwd is their root. `resolveTools` / `resolveAgentTools` follow.
- The harness is context-typed on `ExecutionToolContext` and supplies `toolContext: { env }`. `PiHarnessFactoryOptions.env` returns, this time with a job.
- **`MountedTool`** names what the harness runs — wider than the authored `FastagentTool`, since pi's defaults read a fifth `execute` parameter and fastagent's four-parameter tools are assignable to it unchanged. Every helper that only inspects or reorders tools is typed on it. **`defineTool` is untouched**: authors write exactly what they wrote before.
- `fastagent tool` supplies the context a harness would, rooted at the workspace — without it a direct `execute` crashes on the missing env.
- `chat` is unaffected: it takes these NAMES and lets pi's own runtime rebuild the tools it renders.

## Verify

`npm run lint` + `npm run typecheck` clean, **1237 tests**.

Two `rv` passes. The local one caught L1 (`createPiAgent`) silently dropping `options.env` — the very seam this PR is about. The on-PR one caught the image regression above, plus four claims this branch had made and not finished: ② project context does **not** go through the env (`loadProjectContextFiles` reads node fs directly, a break `definition.ts` already documented), README/ai-start/embedding still asserted the coding tools were cwd/local-fs bound, and `MountedTool` was named in two public option types without being exported.

### DIFF
--- a/docs/design/core.md
+++ b/docs/design/core.md
@@ -282,10 +282,11 @@ test/tools-parity.test.ts, so the choice is about WHERE they touch the machine, 
 Injecting a custom `env` therefore governs the tools that actually do the touching, which is what makes
 it a seam worth having.
 
-It is still not a complete sandbox: `loadProjectContextFiles` and the definition loader read through the
-env, but fastagent's OWN tools (`tools/`) are author code that can import anything, and `deploy`/channel
-machinery runs outside it. A sandbox adapter provides an `ExecutionEnv` AND constrains the process it
-runs in; `env` alone narrows the blast radius rather than closing it.
+It is still not a complete sandbox, and the gaps are specific: `loadProjectContextFiles` reads ② context
+through node fs DIRECTLY rather than the env (definition.ts states this as a known break), fastagent's
+OWN tools (`tools/`) are author code that can import anything, and `deploy`/channel machinery runs
+outside it entirely. A sandbox adapter provides an `ExecutionEnv` AND constrains the process it runs in;
+`env` alone narrows the blast radius rather than closing it.
 
 ## 6. Sessions and concurrency
 
--- a/docs/design/core.md
+++ b/docs/design/core.md
@@ -275,10 +275,17 @@ initial active set at build, and the same builtin loader activates through a ses
 ToolActivation bridge (`sessionToolActivation`) riding the same turn context, so the author debugs
 exactly what serves.
 
-`ExecutionEnv` is a harness assembly seam, not a complete sandbox boundary today. Pi's cwd-bound coding
-tools and `loadProjectContextFiles` still use the local process/filesystem. A future sandbox adapter must
-wire those surfaces as well as provide an `ExecutionEnv`; injecting `env` alone does not isolate a
-directory agent.
+**`ExecutionEnv` is where the default tools reach the machine.** `read`/`bash`/`edit`/`write` come from
+pi-agent-core and take the env as the turn's tool context (pi 0.83), rather than from pi-coding-agent,
+whose identical four are wired to `node:fs` directly — the model-facing surface is asserted equal in
+test/tools-parity.test.ts, so the choice is about WHERE they touch the machine, not what the agent sees.
+Injecting a custom `env` therefore governs the tools that actually do the touching, which is what makes
+it a seam worth having.
+
+It is still not a complete sandbox: `loadProjectContextFiles` and the definition loader read through the
+env, but fastagent's OWN tools (`tools/`) are author code that can import anything, and `deploy`/channel
+machinery runs outside it. A sandbox adapter provides an `ExecutionEnv` AND constrains the process it
+runs in; `env` alone narrows the blast radius rather than closing it.
 
 ## 6. Sessions and concurrency
 
--- a/src/engines/pi/create.ts
+++ b/src/engines/pi/create.ts
@@ -309,10 +309,10 @@ export interface CreatePiAgentOptions {
   authPath?: string;
   /** Session persistence. Defaults to in-memory; inject jsonlSessionStore for restart-surviving continuity. */
   sessions?: PiSessionStore;
-  /** Filesystem/process environment. Defaults to a local NodeExecutionEnv at `process.cwd()`. At THIS
-   *  rung it supplies the agent's cwd and nothing else — pi 0.83 removed the harness's own env, and the
-   *  default coding tools are cwd-bound and local. So injecting it is not a sandbox boundary; a sandbox
-   *  adapter must wire those tools too (pi-agent-core now ships env-backed ones — see create.ts §tools). */
+  /** Filesystem/process environment. Defaults to a local NodeExecutionEnv at `process.cwd()`, and its
+   *  cwd is the agent's. The default coding tools (read/bash/edit/write) take it as the turn's tool
+   *  context, so injecting a constrained one narrows where the agent reads, writes and shells. It does
+   *  NOT constrain author-written `tools/`, which are code and can import anything. */
   env?: ExecutionEnv;
   /** Single-writer lease. Defaults to in-process fail-fast inProcessLease(). */
   lease?: Lease;
@@ -333,6 +333,7 @@ export function createPiAgent(options: CreatePiAgentOptions): Agent {
     tools: options.tools ? withSearchTool(options.tools) : options.tools,
     skills: options.skills,
     sessions: options.sessions,
+    env: options.env,
     lease: options.lease,
     observer: options.observer,
   });
@@ -370,8 +371,9 @@ export interface CreatePiAgentFromDefinitionOptions {
   authPath?: string;
   sessions?: PiSessionStore;
   /** Filesystem/process environment; see {@link CreatePiAgentOptions.env}. At THIS rung it does more
-   *  than carry the cwd: the DEFINITION is read through it (persona.md, skills/, AGENTS.md). The coding
-   *  tools stay local and cwd-bound, so injecting it alone still does not sandbox a directory agent. */
+   *  than root the default tools: the DEFINITION is read through it too (persona.md, skills/,
+   *  AGENTS.md). Author-written `tools/` remain outside it — injecting an env narrows the blast radius
+   *  rather than closing it. */
   env?: ExecutionEnv;
   lease?: Lease;
   /** Observation-plane tap; see {@link CreatePiAgentOptions.observer}. */
--- a/src/engines/pi/create.ts
+++ b/src/engines/pi/create.ts
@@ -11,9 +11,16 @@
  * they come from the definition; the openers own model/tools — from config resolution).
  */
 import { formatSkillsForSystemPrompt } from "@earendil-works/pi-agent-core";
-import type { AgentTool, ExecutionEnv, Skill, ThinkingLevel } from "@earendil-works/pi-agent-core";
+import {
+  type ExecutionEnv,
+  type Skill,
+  type ThinkingLevel,
+  createBashTool,
+  createEditTool,
+  createReadTool,
+  createWriteTool,
+} from "@earendil-works/pi-agent-core";
 import { NodeExecutionEnv } from "@earendil-works/pi-agent-core/node";
-import { createCodingTools } from "@earendil-works/pi-coding-agent";
 import type { Models, Provider } from "@earendil-works/pi-ai";
 import type { Agent } from "../../agent.ts";
 import { type FastagentConfig, defaultAuthPath, resolveModel } from "./config.ts";
@@ -26,11 +33,11 @@ import { type PiSessionStore, inMemorySessionStore } from "./sessions.ts";
 import type { ModuleLoadFailure } from "../../loader.ts";
 import {
   type DefineToolOptions,
-  type FastagentTool,
   type ToolCollision,
   isDeferredTool,
   loadTools,
   mergeDiscoveredTools,
+  type MountedTool,
 } from "./tool.ts";
 import { withSearchTool } from "./search-tools.ts";
 import { type Lease, type SessionObserver, createPiAgentFromHarness, inProcessLease } from "./invoke.ts";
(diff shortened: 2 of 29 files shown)


src/engines/pi/create.ts
```diff
@@ -26,29 +33,42 @@ import { type PiSessionStore, inMemorySessionStore } from "./sessions.ts";
...
+// `chat` is unaffected: it takes these NAMES only and lets pi's own runtime rebuild the tools it
+// renders (see session-builder.ts).
 
-/** pi's core default toolset (read/bash/edit/write), rooted at cwd. */
-export function piDefaultTools(cwd: string): AgentTool[] {
-  return createCodingTools(cwd) as AgentTool[];
+/** pi's core default toolset (read/bash/edit/write). Rooted at the ExecutionEnv's cwd, supplied per
+ *  turn as the harness tool context — hence no argument here. */
+export function piDefaultTools(): MountedTool[] {
+  return [createReadTool(), createBashTool(), createEditTool(), createWriteTool()];
```
kid7st: `createReadTool()` is constructed with no options, so no `imageProcessor` — pi-coding-agent's `createReadTool(cwd)` wires `processImage` (convert→resize→size-limit) by default. Reading an image therefore behaves differently after the swap, on a path agents do take:

- jpeg/png/gif/webp are inlined as raw base64, unresized and with no size cap (pi-coding-agent resizes below the inline limit, or reports `[Image omitted: could not be resized below the inline image size limit.]`). A screenshot-sized file now goes to the provider whole — a token/size blow-up with no diagnostic.
- bmp is dropped entirely with `[Image omitted: configure an imageProcessor to convert BMP images.]`, while the tool description this PR asserts is byte-identical still says "Supports text files and images (jpg, png, gif, webp, bmp)". The parity test pins the description as equal and thereby certifies a description that is false for this build.
- `chat` keeps pi's own tools, so chat and serve now answer `read <image>` differently — the fidelity argument three lines above ("authors vibe in local pi with it, so serving with fewer tools is behavior drift") cuts against this.

This is not a documented divergence: the "two places they DO differ" test lists only the ENOENT syscall name and the empty-output bash message.

Direction: pass an `imageProcessor` to `createReadTool` that delegates to pi-coding-agent's `processImage` — `src/engines/pi/invoke.ts:365-371` already lazy-imports `resizeImage` from that package for the prompt-image path, so the machinery and the precedent are both in-repo — and add the image case to `test/tools-parity.test.ts`.

src/engines/pi/create.ts
```diff
@@ -347,8 +371,9 @@ export interface CreatePiAgentFromDefinitionOptions {
   authPath?: string;
   sessions?: PiSessionStore;
   /** Filesystem/process environment; see {@link CreatePiAgentOptions.env}. At THIS rung it does more
-   *  than carry the cwd: the DEFINITION is read through it (persona.md, skills/, AGENTS.md). The coding
-   *  tools stay local and cwd-bound, so injecting it alone still does not sandbox a directory agent. */
+   *  than root the default tools: the DEFINITION is read through it too (persona.md, skills/,
```
kid7st: "the DEFINITION is read through it too (persona.md, skills/, AGENTS.md)" — AGENTS.md is not. `definition.ts:6-11` states the opposite in its own header: ② project context comes from pi's `loadProjectContextFiles`, which "reads via node fs DIRECTLY (not the injected `env`)" and swallows a permission error in its `existsSync` probe, so under a non-local env the ② files resolve against THIS process's filesystem. Two files in the same engine now describe the same read in contradictory terms, and this PR's whole point is making readers trust the env seam.

Direction: drop AGENTS.md from the list and name it as the exception ("persona.md and skills/ are read through it; ② project context still resolves against the local filesystem — see definition.ts").

src/engines/pi/create.ts
```diff
@@ -26,29 +33,42 @@ import { type PiSessionStore, inMemorySessionStore } from "./sessions.ts";
...
+// fewer tools is behavior drift. Locking down for public exposure = passing a restricted `tools` list
+// (a deployment posture).
+//
+// These are pi-agent-core's tools, which reach the filesystem and the shell through the
+// {@link ExecutionEnv} the harness hands them per turn — NOT pi-coding-agent's, which are the same four
+// tools wired to `node:fs` directly (identical names, descriptions, parameters and results — asserted
+// in test/tools-parity.test.ts). Going through the env is the whole point: it makes
+// {@link CreatePiAgentOptions.env} the one seam a sandbox adapter has to implement, instead of a knob
+// that governed everything except the tools that actually touch the machine. It also keeps the serving
+// path off pi's TUI stack, which the coding-agent tools carry for rendering fastagent never uses (its
```
kid7st: "It also keeps the serving path off pi's TUI stack" is not true of the serving path as a whole: `definition.ts:21` imports `loadProjectContextFiles` and `models.ts:9` imports `ModelRuntime` from `@earendil-works/pi-coding-agent`, both unconditionally, both on the `start`/`dev` path. The package is loaded regardless; only the tools stopped coming from it.

Direction: either delete this sentence (the env argument above carries the change on its own) or scope it honestly — "the tools no longer pull pi's rendering stack, though the serving path still imports the package for `loadProjectContextFiles` and `ModelRuntime`."

docs/design/core.md
```diff
@@ -275,10 +275,17 @@ initial active set at build, and the same builtin loader activates through a ses
...
-wire those surfaces as well as provide an `ExecutionEnv`; injecting `env` alone does not isolate a
-directory agent.
+**`ExecutionEnv` is where the default tools reach the machine.** `read`/`bash`/`edit`/`write` come from
+pi-agent-core and take the env as the turn's tool context (pi 0.83), rather than from pi-coding-agent,
+whose identical four are wired to `node:fs` directly — the model-facing surface is asserted equal in
+test/tools-parity.test.ts, so the choice is about WHERE they touch the machine, not what the agent sees.
+Injecting a custom `env` therefore governs the tools that actually do the touching, which is what makes
+it a seam worth having.
+
+It is still not a complete sandbox: `loadProjectContextFiles` and the definition loader read through the
```
kid7st: "`loadProjectContextFiles` and the definition loader read through the env" — `loadProjectContextFiles` does not; `definition.ts:6-11` records that it reads node fs directly and that this is a known break in the portability contract, deferred with the sandbox work. This paragraph is the section a reader consults to decide what a custom env buys them, so naming the one loader that escapes it as being inside it inverts the answer.

Direction: "the definition loader (persona.md, skills/) reads through the env; `loadProjectContextFiles` (② project context) still does not — see definition.ts — and fastagent's own `tools/` are author code that can import anything."

kid7st (commented):
Thesis: the default coding tools move from pi-coding-agent's cwd-closed four to pi-agent-core's env-backed four, so `ExecutionEnv` becomes the single seam through which the agent touches the machine. That thesis is clear; the findings are about where the swap is not actually equal, and where the companion text now says the opposite of what the code does.

**Findings that do not anchor to a diff line:**

1. `README.md:199`, `docs/ai-start.md:197`, `docs/embedding.md:152-153` still state the old claim — "the pi coding tools and project-context loader are still local" / "the pi coding tools created for a directory ... are cwd/local-fs bound. Injecting `env` alone therefore does not isolate a directory agent". After this PR the tools half is false. `docs/api-reference.md` and `docs/design/core.md` were updated; these three were not, so the repo now asserts both. `docs/ai-start.md` is the worst of the three: it is an instruction sheet for agents ("do not claim that injecting `env` alone isolates a directory agent"), so the stale sentence will be repeated forward. Update all three to the same wording used in api-reference.md.

2. Public option types name a type the package does not export. `CreatePiAgentOptions.tools` and `CreatePiAgentFromDefinitionOptions.tools` are now `MountedTool[]`, but `src/pi.ts` exports `FastagentTool` and not `MountedTool` (nor `ExecutionToolContext`), so a caller who wants to declare the array cannot name its type through the public surface, and the JSDoc `{@link MountedTool}` resolves to nothing for them. Either export `MountedTool` from `src/pi.ts` (and note it in `docs/api-reference.md`, whose `tools` row is still just "Agent tools."), or keep the public options typed `FastagentTool[]` and let `MountedTool` stay the internal harness-facing type.

3. Test coverage gap tied to inline finding #1 below: `test/tools-parity.test.ts` claims "results ... on the paths an agent actually takes" but exercises only text `read`, `bash`, `edit`, `write`. The one path where the two implementations genuinely diverge in RESULT — reading an image — is not exercised, so the test certifies parity precisely where it does not hold. Add an image case (a small png and a bmp) to that test; it will fail today, which is the point.

kid7st:
Measured finding 1 before deciding what to do about it — it is worse than the bmp case, and it lands on the path agents actually take (screenshots).

`read` on a 5.6 MB PNG, both implementations, same file:

| | returns | dimension note |
|---|---|---|
| pi-coding-agent (before) | converts to JPEG, **3.48 MB** base64 | yes — `original 1400x1400, displayed at 1400x1400. Multiply…` |
| pi-agent-core (this PR) | raw PNG, **7.48 MB** base64 | none |

So the swap **doubles the payload sent to the provider** for every image read and drops the scale factor an agent needs for coordinate math. And on `t.bmp`:

```
before: text "Read image file [image/png]\n[Image converted from image/bmp to image/png]" + image
after : text "Read image file [image/bmp]\n[Image omitted: configure an imageProcessor…]"
```

— while the description this PR's parity test pins as *identical* still advertises bmp support. The test certifies a claim that is false for this build, exactly as noted.

**Fixable, but the fix is upstream duplication.** `processImage` is not exported; its two halves are (`convertToPng`, `resizeImage` + `formatDimensionNote`). Rebuilding it faithfully means copying ~45 lines: the supported-mime table, the conversion/dimension hint wording, and both fallback messages — all of which then have to track upstream by hand, with only the parity test to catch drift.

Which changes this PR's ledger enough to re-open the premise, so recording it here rather than grinding through the findings:

- The **second** stated benefit is false as written (finding 3): `definition.ts` and `models.ts` import pi-coding-agent unconditionally on the serving path, and the image fix would add two more imports from it. The swap does not reduce that coupling at all.
- The **first** benefit — `env` reaching read/bash/edit/write — is real but latent: no sandbox adapter exists, and author-written `tools/` stay outside it regardless.

So the choice is: pay ~45 lines of duplicated upstream image logic (plus the five other findings) for a seam nobody consumes yet, or keep pi-coding-agent's tools until pi exports `processImage` — at which point the swap becomes genuinely free.

kid7st:
All six addressed in `d1b3a1c` — the four inline threads are resolved.

**The regression is fixed, and it was the measurement that decided it.** `read-image.ts` rebuilds pi's `processImage` from the halves that *are* public (`convertToPng`, `resizeImage`, `formatDimensionNote`) — 20 lines, not the ~45 I estimated when I thought the mime table and base64 plumbing had to be restated too. Verified byte-identical against pi-coding-agent's real `read` on both paths:

```
huge.png (5.6MB)  IDENTICAL ✓     (3.48MB JPEG + the dimension note)
t.bmp             IDENTICAL ✓     (converted to PNG + the conversion hint)
```

The parity test now covers images, and I mutation-checked it: dropping `imageProcessor` from `createReadTool` fails it on the bmp case. That is what turns the restated upstream logic from silent drift into a failing build.

**Which settles the question this PR was left hanging on.** The ledger comment above weighed "45 lines of duplication for a latent seam" and leaned toward closing. At 20 lines, guarded by a test, the balance goes the other way — so this stays open.

But the second benefit really was false, and the comment now says so rather than being quietly deleted: `definition.ts`, `models.ts` and `read-image.ts` all import pi-coding-agent, so **the swap buys the env seam and nothing else.** No decoupling.

Also corrected, since they were claims this branch made and did not finish:

- ② project context does **not** go through the env — `loadProjectContextFiles` reads node fs directly, which `definition.ts` already documented as a known break. The option doc and core.md said the opposite.
- README.md, docs/ai-start.md and docs/embedding.md still asserted the coding tools are cwd/local-fs bound. `ai-start.md` is instructions *for agents*, so that one was actively propagating.
- `MountedTool` is exported from `src/pi.ts`, and api-reference's `tools` row explains the widening instead of saying "Agent tools."

PR #301, opened 2026-08-01
fix(deps): force brace-expansion past the DoS advisory (GHSA-mh99-v99m-4gvg)

kid7st:
Closes the open Dependabot alert on `main` ([#10](https://github.com/fastagent-sh/fastagent/security/dependabot/10), HIGH).

## The advisory

`brace-expansion` **5.0.7** — [GHSA-mh99-v99m-4gvg](https://github.com/advisories/GHSA-mh99-v99m-4gvg): unbounded expansion length can OOM the process. It reaches us transitively and is a **runtime** dependency of the published package, so every consumer's scanner reports it:

```
@fastagent-sh/fastagent
└─ @earendil-works/pi-coding-agent@0.83.0
   └─ minimatch@10.2.5
      └─ brace-expansion@5.0.7        ← advisory; patched in 5.0.8
```

## Why `overrides`, and not `npm audit fix`

minimatch asks for `^5.0.5`, and 5.0.9 satisfies it — so the obvious fixes *should* work. None of them do; I checked each:

| Attempt | Result |
|---|---|
| `npm audit fix` | reports the fix as available, changes nothing |
| `npm update brace-expansion` | no move |
| **delete the lockfile and re-resolve from scratch** | **still lands on the nested 5.0.7** |

Only `overrides` moves it. So the three lines are load-bearing, not belt-and-braces — worth stating, since "why not just `npm update`" is the first thing to ask.

Side effect, welcome: the regenerated lockfile is **smaller** (`-790/+65`), because the nested copies get hoisted and de-duplicated. The only other movement is `@emnapi/core` and `@emnapi/runtime` 1.11.1 → 1.11.2, both dev-only transitives.

## Reachability, for the record

Not a live hole here, and the assessment belongs with the fix rather than in someone's head:

`minimatch` is called in pi-coding-agent from two places — `model-resolver` (globs from `--models` / `enabledModels` scoping) and `package-manager` (workspace patterns out of a `package.json`). **fastagent supplies neither**: it never uses pi's model scoping (no reference anywhere in `src/`), and runtime/package-manager detection is its own (`src/runtime.ts`).

So this is supply-chain hygiene, not an incident response. Worth doing anyway: we do not control pi's pin, and three lines are cheaper than tracking the advisory.

## When to remove it

Once pi ships a minimatch whose tree resolves 5.0.8+ on its own, the override becomes a no-op floor and should go. Until then it is what keeps a fresh `npm install` off the vulnerable copy — noted in the commit so it does not rot.

## Verify

`npm audit` → **found 0 vulnerabilities**. `npm run lint` + `npm run typecheck` clean, **1234 tests** passing.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -41,8 +41,6 @@
     },
     "node_modules/@anthropic-ai/sdk": {
       "version": "0.91.1",
-      "resolved": "https://registry.npmjs.org/@anthropic-ai/sdk/-/sdk-0.91.1.tgz",
-      "integrity": "sha512-LAmu761tSN9r66ixvmciswUj/ZC+1Q4iAfpedTfSVLeswRwnY3n2Nb6Tsk+cLPP28aLOPWeMgIuTuCcMC6W/iw==",
       "license": "MIT",
       "dependencies": {
         "json-schema-to-ts": "^3.1.1"
@@ -61,8 +59,6 @@
     },
     "node_modules/@aws-crypto/sha256-browser": {
       "version": "5.2.0",
-      "resolved": "https://registry.npmjs.org/@aws-crypto/sha256-browser/-/sha256-browser-5.2.0.tgz",
-      "integrity": "sha512-AXfN/lGotSQwu6HNcEsIASo7kWXZ5HYWvfOmSNKDsEqC4OashTp8alTmaz+F7TC2L083SFv5RdB+qU3Vs1kZqw==",
       "license": "Apache-2.0",
       "dependencies": {
         "@aws-crypto/sha256-js": "^5.2.0",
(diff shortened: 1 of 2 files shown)


kid7st:
Closing — verified that this does not do what the description claims.

**`overrides` does not reach consumers.** npm honors the field only for the ROOT project. I packed this branch and installed it into a clean project: the field ships in our published `package.json`, is ignored, and the consumer still resolves `brace-expansion@5.0.7`, with `npm audit` reporting the same 1 high. So this would only have silenced the alert on *this* repository while looking like it protected users — worse than not fixing it.

**The fix is upstream, and already exists.** `minimatch@10.2.6` requires `brace-expansion: ^5.0.8`, the patched line. It cannot reach anyone because `@earendil-works/pi-coding-agent` pins its dependency **exactly**:

```json
"minimatch": "10.2.5"
```

No caret — so neither a consumer's fresh install nor our lockfile can move off it. Relaxing that pin to `^10.2.5` (or bumping to `10.2.6`) fixes it for every dependent at once; nothing in our own `package.json` can.

**Not reachable here in the meantime.** `minimatch` is called from pi-coding-agent in exactly two places — `model-resolver` (globs from `--models` / `enabledModels` scoping) and `package-manager` (workspace patterns out of a `package.json`) — and fastagent supplies neither.

Tracking upstream instead. The advisory stays visible in the security tab rather than being papered over by a field that does nothing for the people it claims to protect.

Issue #284, opened 2026-07-26
SessionControl: add a `navigate` command so a client can move the session leaf


kid7st:
## What

Add a `navigate` command to `SessionControl`, so a client can move a session's leaf to another point
in its tree — and derive `SessionCapabilities` from the dispatch map rather than writing it by hand.

## Why

The session control plane already publishes `SessionEntry.parentId` and `SessionEntries.leafEntryId`,
and `entries()` tells clients to reconstruct the active path by walking parent chains from the leaf.
**The contract has already committed to a session being a tree.** What it has no verb for is moving
the leaf, so a client can read the branch structure and never act on it — which means a session
driven only through this plane is effectively linear, since navigation is also how sibling branches
come to exist.

This is the last thing keeping a GUI client from replacing `pi --mode rpc`. pi's own `/tree` is
unreachable over its RPC surface (`navigateTree` is exposed there only to extensions running inside
pi, not as a command a client may call), and going through fastagent removes that ceiling for every
other operation — this one included, once the verb exists.

## Why it is small

`navigate` is already a primitive on `SessionStorage`, the `pi-agent-core` port in use today:

```ts
getLeafId(): Promise<string | null>;
/** Persist a leaf entry that records the active session-tree leaf. */
setLeafId(leafId: string | null): Promise<void>;
```

`setLeafId` **is** the navigation. pi-coding-agent's `navigateTree` adds one thing on top: an
optional LLM summarisation of the branch being left. That belongs behind a capability flag, not in
the core of the command.

## Why it does not cost engine neutrality

`navigate` is not a pi concept leaking in. It is the verb for a noun the contract already has —
entries with parents and a leaf. An engine whose sessions are linear reports `navigate: false` and
nothing changes for it. Only `summarize` is engine-flavoured, and it is optional.

## Proposed shape

```ts
type SessionCommand =
  | …existing…
  | { type: "navigate"; targetId: string; summarize?: boolean; instructions?: string };
```

A boundary mutation: takes the same lease runs take, refused mid-run like `compact` already is, and
answered with the existing codes — `unsupported_capability` when not offered, `session_busy` while a
run holds the lease, `invalid_command` when `targetId` names no entry.

### Capabilities, derived rather than declared

A client decides whether a tree view exists *at all* from `capabilities()`, so a capability that is
declared and unimplemented produces an affordance that appears and then fails. That is avoidable by
construction:

```ts
interface SessionCapabilities {
  …existing…
[...]

PR #302, opened 2026-08-03
refactor(session-control)!: one resolution for the session's settings pair

### LINKED ISSUE #288: capabilities() answers from a constant where the model has the answer (thinking levels lie; model metadata discarded)
## Two symptoms, one cause

`capabilities()` computes `modelSelection` and `thinkingLevel` without consulting the model, even
though the model object carrying both answers is in hand at that moment.

### 1. Thinking levels are advertised for models that cannot do them

```ts
// src/engines/pi/session-control.ts:262
thinkingLevel: b ? { allowedLevels: [...THINKING_LEVELS] as string[] } : false,
```

A static constant. But pi-ai's `Model` is explicit that this is per-model:

```ts
reasoning: boolean;
/** Maps pi thinking levels to provider/model-specific values.
 *  Missing keys use provider defaults. null marks a level as unsupported. */
thinkingLevelMap?: ThinkingLevelMap;
```

So on a `reasoning: false` model, `capabilities()` still lists every level, and `set_thinking`
validates only against that same static set:

```ts
// :459
if (!(THINKING_LEVELS as ReadonlySet<string>).has(command.level)) { … INVALID_COMMAND_CODE … }
```

The dispatch returns `ok: true` and the level does nothing. A client that gates its UI on
`capabilities()` — which the contract instructs it to do — renders a control that reports success
and has no effect. That is the one failure mode `SessionCapabilities` exists to prevent, and it is
worse than an unsupported command, because `UNSUPPORTED_CAPABILITY_CODE` at least tells the truth.

The same applies after a `set_model`: capabilities do not change, so a client cannot learn that the
levels it is offering stopped being real.

### 2. `allowedModels: string[]` discards metadata already in the loop

```ts
// src/engines/pi/config.ts:285
for (const model of provider.getModels()) specs.push(`${provider.id}/${model.id}`);
```

`model` here is the full pi-ai `Model` — `name`, `contextWindow`, `maxTokens`, `cost`, `reasoning`,
`input`, `thinkingLevelMap` — projected to one string. This is not a missing data source that would
have to be built; it is a projection that could keep a few more fields.

An interactive client shows a model picker. With strings it can show ids. `contextWindow` and
`reasoning` are what make the choice informed, and `state.usage.contextWindow` only covers the model
already selected — the picker is about the ones that are not.

## Suggested shape

```ts
modelSelection: false | {
  allowedModels: Array<{
    id: string;              // "provider/model", today's string — unchanged as the identifier
    name?: string;
    contextWindow?: number;
    maxTokens?: number;
    reasoning?: boolean;
  }>;
};

// per-model, and re-read after set_model
thinkingLevel: false | { allowedLevels: string[] };
```

Everything past `id` is optional, so an engine with no catalog returns bare entries and the contract
does not force metadata nobody has. The breaking part is `string[]` → object array; if that is
unwelcome, a parallel optional map keyed by spec would also work, at the cost of two things to keep
in step.

The thinking-level half is fixable without any contract change at all — derive `allowedLevels` from
the session's current model instead of the constant, and have `set_thinking` validate against the
same source. That is the half I would prioritise: one is a missing nicety, the other is a capability
that lies.

## Relationship to #284

Same principle from the other end. #284 argues capabilities should be *derived* rather than
hand-written so an unimplemented one cannot be advertised; this is a case where a hand-written one
already is. Derivation there would not have caught it by itself — `set_thinking` does have a handler
— so the rule needs the second half: **derived from the handler, and evaluated against the session's
current state**, not fixed at construction.

## Context

duang.ai, same client as #284–#287. Its model picker renders `contextWindow` and `reasoning` today
via pi's RPC `get_available_models`; the thinking-level issue is one it would inherit rather than
one it hit, since it has not switched yet.

Happy to send the PR — probably as two, since the thinking-level fix is contained and the metadata
one touches the contract.

kid7st:
Closes #288 (the thinking-level half).

## What was wrong

`capabilities()` advertised pi's whole thinking scale and `set_thinking` validated against the same constant, so on a non-reasoning model a client rendered every level, the dispatch answered `ok`, a durable override landed, and the run ignored it — the exact failure `SessionCapabilities` exists to prevent.

The first fix answered from the model instead. That moved the lie rather than removing it, and the next three rounds each moved it again — to `state()`, then to the execution resolve, then to a `set_model` special case. That pattern is the signal, so this is a re-derivation rather than a fourth patch.

## The tension

> **"How hard to think" is only meaningful relative to "which model", and the two are set independently, at different times, by different actors** — a control client, the config, pi's own TUI.

So any surface that reports or validates one without the other can be asked a question it cannot answer correctly. Solving that needs *one* resolution and one place that owns it.

What the branch had instead: a sessionless capability list computed from the configured model, a pre-lease dispatch check documented as *advisory*, an execution clamp documented as a *backstop*, a `state()` reporting recorded-not-executed with a paragraph explaining the split, a `set_model` re-record to close one of two paths into it, and prose in five places about which of those was authoritative. **That gap was most of the diff.**

The missing abstraction is **"the session's current settings"** — nobody owned it, so three neighbours each re-derived it and then documented their own partiality.

## What this does

`src/engines/pi/session-settings.ts` owns the resolution. `state()` reports it, the `set_thinking` gate validates against it, and the fresh-harness resolve executes it.

**Contract (breaking, pre-1.0):**

| | before | after |
|---|---|---|
| `capabilities().thinkingLevel` | `false \| { allowedLevels }` | `boolean` |
| `state().availableThinkingLevels` | — | the levels this session's model supports |
| `state().model` / `.thinkingLevel` | what was RECORDED | what will RUN (resolved + clamped) |

`allowedModels` **stays** on `capabilities()`: the registry is a deployment fact, identical for every session, so a static list there is correct. Only thinking levels are session-dependent — which narrowed this from the first sketch.

Clients (#284/#286) are not built yet, so this is the cheap moment; after them it is not.

## A behaviour improvement, not just less code

Deleting the `set_model` re-record made something better. Resolving *per read* means a level the current model cannot do is reported and executed as the clamped one **while the record keeps the user's preference**:

```
set_thinking high            → recorded: high
set_model <non-reasoning>    → state: off,  runs: off,  available: [off]
set_model <reasoning again>  → state: high              ← preference restored
```

The re-record overwrote `high` with `off` permanently. Pinned by a test.

## What survives, and why it is irreducible

The execution resolve still clamps and warns — because pi appends `model_change`/`thinking_level_change` itself (its TUI's `/model`), and a deployment's configured model can change between restarts, so a recorded pair can stop being executable with **no control-plane command involved**. Its comment says that now instead of "backstop", and states the clamp's real direction: pi takes the lowest supported level **at or above** the recorded one, so a gap in a model's scale resolves **upward** — it costs more, not less.

## Measurement

Line count is roughly flat (1040 → 1062 across the three files). What changed is **the number of places that can answer the question: three → one**, and the number of surfaces that had to document their own partiality: three → zero.

## Verify

`npm run lint && npm run typecheck && npm test` — **1258 tests**, rebased onto `main`.

New coverage: one resolution answering all three surfaces (state / gate / execution) including the preference-restored case, a non-reasoning model offering only `off` with `set_thinking` rejected *before* any entry is recorded, `state()` reporting the pair with no overrides at all, and the clamp's upward direction pinned against pi (the faux registry cannot express a gapped scale, and the pre-existing clamp test used a non-reasoning model where the direction is unobservable).


src/engines/pi/harness.ts
```diff
@@ -216,24 +235,34 @@ export function resolveHarnessOverrides(
...
         `[fastagent] session ${sessionId}: recorded thinking level "${recorded.thinkingLevel}" is unknown — using the configured default`,
       );
+    } else {
+      // Against the model resolved ABOVE: `set_thinking` is admitted against the session's model of
+      // the moment, and a later `set_model` can strip that level's support out from under it — the
+      // one order the control-plane check cannot see. pi's own clamp (the one its TUI applies), not
+      // a rule of ours — and only for a level fastagent ACCEPTED; with no override, the configured
+      // default reaches pi's clamp at the model call as before.
+      const level = recorded.thinkingLevel as ThinkingLevel;
+      thinkingLevel = clampThinkingLevel(model, level) as ThinkingLevel;
```
kid7st: **"nearest supported level" describes a clamp that biases UPWARD.** `clampThinkingLevel` scans from the requested index *up* the scale first, and only falls back downward if nothing above exists. A recorded `low` on a model with `{minimal: null, low: null}` executes as `medium`/`high`/`max` — a silent cost increase, not a conservative degrade. The comment here and `docs/api-reference.md:487` both say "nearest".

Direction: state the real rule ("the lowest supported level at or above the recorded one, else the highest below") and add the upward case to the tests — the only clamp test exercises a non-reasoning model, where the answer is always the trivial `off`.

src/engines/pi/harness.ts
```diff
@@ -216,24 +235,34 @@ export function resolveHarnessOverrides(
...
+      if (thinkingLevel !== level) {
+        warnOnce(
+          `${sessionId}\u0000thinking\u0000${model.provider}/${model.id}\u0000${level}`,
+          `[fastagent] session ${sessionId}: recorded thinking level "${level}" is not supported by ${model.provider}/${model.id} — running at "${thinkingLevel}"`,
+        );
+      }
     }
```
kid7st: **The clamp warns the operator, not the party responsible for the choice.** The `set_thinking` → `set_model` order is settled with a server-side `log.warn`, while `state()` keeps reporting the recorded level and no `state_changed` is emitted. For a remote control-plane client — the consumer this plane exists for — the degrade is invisible: it set `high`, `state()` says `high`, runs execute at `off`. That is the same "reports success and does nothing" failure this PR opens by naming, moved from the dispatch to the state surface; "warned not silent" only holds for whoever reads the server log.

Direction: emit `state_changed{thinkingLevel}` when the clamp changes the executed level, or have `set_model` reject / append the clamped level at the boundary so the recorded pair is always executable.

src/engines/pi/session-control.ts
```diff
@@ -177,6 +179,9 @@ export interface PiBoundaryWiring {
   lease: Lease;
   models: Models;
   harnessFactory: PiHarnessFactory;
+  /** The assembly's configured model — the base every session's thinking-level answer starts from
+   *  (a session without a `set_model` override runs on exactly this). */
+  defaultModel: AnyModel;
```
kid7st: **`defaultModel` duplicates the harness factory's configured model with nothing keeping them equal.** `PiBoundaryWiring` now carries a copy of the value `piHarnessFactory` was constructed with. `create.ts` hands both out of the same local, but the interface admits any pair — a caller that wires them differently makes `capabilities()` and `set_thinking` validate against a model no run uses, the exact class of lie this PR removes.

Direction: expose the factory's own model (a property on `PiHarnessFactory`) rather than adding a fourth independent field, so the disagreement is unrepresentable.

kid7st (commented):
Thesis: thinking levels stop being a static scale and become a per-model derived set, with `capabilities()` answering for the configured model, `set_thinking` for the session's, and `resolveHarnessOverrides` clamping at execution.

**`src/session.ts:25-33` + `docs/design/session-control.md:174` — the contract still calls capabilities a hard gate; this PR made `thinkingLevel.allowedLevels` advisory.**

The `SessionCapabilities` docstring says: "COMMAND GATES … clients MUST gate dispatch on them; an unsupported command is rejected … with `unsupported_capability`". After this change `allowedLevels` is the CONFIGURED model's set, so for a session that ran `set_model` it is wrong in both directions (a listed level rejects; an unlisted level may be valid), and the rejection code is `invalid_command`, not `unsupported_capability`. Only `docs/api-reference.md` learned this — the normative text in `src/session.ts` and design §5.3 did not.

Direction: say in both places that `allowedLevels` is a hint whose authority is the dispatch, or put the session's real set on `state()`. Today no surface exposes it, so a client discovers it only by being rejected against a list it was never shown (the dispatch error prints `allowed: …` from the session's model; `capabilities()` printed the default's).

**`src/engines/pi/harness.ts:143-150` — the `THINKING_LEVELS` docstring is now false.**

"THE single source for fastagent (… session-control's dispatch validation and capabilities derive from this)": capabilities now derives from `getSupportedThinkingLevels`, and dispatch uses this set only to produce the off-scale message. It is the vocabulary, not the source of truth — the comment claims the role this PR just took away from it. Rewrite it as the scale/exhaustiveness anchor it now is.

kid7st:
Read the whole delta and ran it against the registry. The core move is right and the tests are sharp — `not.toContain("thinking_level_change")` (rejected *before* acceptance, not recorded as a no-op) and `state.thinkingLevel` staying `"high"` while the resolve returns `"off"` are both the hard assertions, not the easy ones. Two claims in the description that I checked and that hold:

- pi-ai really does clamp at the model call (`api/openai-completions.js:469`), so deferring the config-level case to it is accurate, not a hand-wave.
- `getSupportedThinkingLevels` behaves as described: `openai/gpt-4o-mini` → `["off"]`, `openai-codex/gpt-5.5` → six levels, and `xhigh`/`max` need an explicit mapping.

Also: **this rebases onto `main` cleanly** (I tried it), despite #294 and #300 landing under it.

One thing to add to the review above — **its findings 1 and 2 compound into a reachable user-visible bug**, which changes them from wording to fix-before-merge.

### The clamp biases upward, and the state surface hides it

`clampThinkingLevel` scans **up** the scale from the requested level, and only falls back downward when nothing above exists (`pi-ai/dist/models.js:411-420`). Confirmed on a model whose support set has a hole:

```
supported: ["off", "high"]
  clamp(minimal) → high
  clamp(low)     → high
  clamp(medium)  → high
```

So `"nearest supported level"` in the comment (and `docs/api-reference.md:487`) is not what happens: it is *the lowest supported level at or above the request, else the highest below*. The direction matters, because now put it together with the state split:

```
set_thinking low   (model A supports low)   → ok
set_model    B     (B supports off, high)   → ok
state()                                     → thinkingLevel: "low"
every run executes at                       → "high"
```

The client asked for the **cheaper** level, was told ok, is still told `low`, and every turn bills at `high`. The only signal is a `log.warn` on the server — which is exactly the shape this PR's own opening paragraph condemns ("the dispatch answered ok, a durable override landed, and the run ignored it"), moved from the dispatch to `state()`.

In fairness: the "state reports the recorded value, the resolve owns the fallback" split is **inherited**, not invented here — the same comment already covers `model`. What is new is that a *valid, accepted* level can now be silently executed as a *different and more expensive* one. Before this change a recorded level was either executed or unknown; there was no third case.

That is why I would not leave it at a comment fix. Either:

- emit `state_changed{thinkingLevel}` when the clamp changes the executed level (the control plane already emits one for `set_model`), or
- have `set_model` reject / re-record at the boundary, so the recorded pair is always executable.

Either way the description of the clamp should say which direction it goes, and the tests should cover a gapped model — the only clamp test today uses a non-reasoning model, where the answer is the trivial `off` and the direction is unobservable.

### On the third finding (`defaultModel`)

Agreed that the pair can drift, though the practical risk is small — `create.ts` hands both out of one local, and `PiBoundaryWiring` is only ever constructed through `onAssembly`. If exposing the factory's model is awkward, saying *in the type* that `defaultModel` must be the one `harnessFactory` was built with is most of the value for none of the surgery.

kid7st:
Pushed `15331e7`, rebased onto `main`; the three inline threads are resolved.

**The two findings compound, so I fixed the cause rather than the wording.** pi's clamp scans *up* the scale and only falls back downward when nothing above exists, which makes a gap resolve to a **more expensive** level — and `state()` reports the recorded value, so:

```
set_thinking low   (model A supports low)   → ok
set_model    B     (B supports off, high)   → ok
state()                                     → "low"
every run executes at                       → "high"
```

The client asked for the cheaper level, was told ok, is still told `low`, and every turn bills at `high` — the only signal a server-side warn. That is this PR's own opening complaint, one surface over.

So `set_model` **re-records**: when the new model cannot do the session's recorded level, the clamped level is appended in the same mutation and the single `state_changed` carries both halves. Whatever is recorded is executable, and `state()` agrees with the run. It reads and writes under the **lease** — unlike the `set_thinking` check, which only reads and is advisory by construction — so a concurrent `set_thinking` cannot be clamped against a model it never saw.

The resolve keeps clamping as the **backstop** for the case the boundary cannot see: a session with no model override whose *configured* model changed between restarts. Its comment (and `docs/api-reference.md`) now state the direction instead of saying "nearest".

On the other three:

- **`src/session.ts` + design §5.3.** Both now separate the halves: each gate's **boolean** is authoritative, its **list** is a hint. `allowedLevels` answers for the configured model, so after a `set_model` it can be wrong in both directions, and the dispatch — which reads the session's model — is the authority. Both also say that moving the per-session set onto `state()` is a contract change, not a patch, so the next reader does not "fix" it in passing.
- **`THINKING_LEVELS`.** Rewritten as the vocabulary it now is (which strings are levels at all — what the pre-session scale check needs), not the source of truth it stopped being.
- **`defaultModel`.** Took the cheap half of your suggestion: the type now says it must be the model `harnessFactory` was built with. Exposing the factory's model would make it unrepresentable, but `PiBoundaryWiring` is only ever constructed through `onAssembly` out of one local, so the surgery buys little over saying so.

Tests: the re-record (mutation-checked — dropping the append fails it), the no-op case (a model that *can* do the level records nothing extra), and the clamp's upward direction pinned against pi directly, since `FauxModelDefinition` has no `thinkingLevelMap` and the existing clamp test uses a non-reasoning model where the direction is unobservable.

`npm run lint` + `npm run typecheck` clean, **1257** tests.

PR #309, opened 2026-08-03
feat(session-control): navigate — a client can move the session leaf

### LINKED ISSUE #284: SessionControl: add a `navigate` command so a client can move the session leaf

kid7st:
## What

`navigate` — a `SessionCommand` that moves a session's active leaf, so a client can act on the tree
the contract already publishes (`SessionEntry.parentId` + `SessionEntries.leafEntryId`). Closes #284.

```ts
| { type: "navigate"; targetId: string };   // capabilities().navigate gates it
```

A boundary mutation like `set_model`: same lease as a run (`session_busy` mid-run), `no_such_session`
on an unknown session, `invalid_command` for a `targetId` that is not an id `entries()` published, and
the move rides out as `state_changed{leafEntryId, model, thinkingLevel}`. Implemented over pi's
`Session.moveTo` (the `SessionStorage.setLeafId` primitive) — no new engine surface.

## The part that is not the verb

A movable leaf falsifies the premise the last-wins reads were built on: *"serving sessions never
branch, so a flat `getEntries()` read is correct"*. After a move, the journal still holds the
abandoned branch, so the fresh-harness override/activation walk, `state()` and the `set_thinking`
gate would all apply a setting the session left behind. They now read the ACTIVE PATH
(`activePathEntries`, in `sessions.ts` beside the session handles). Same reason, `attach`'s replay
renders the active branch instead of interleaving the abandoned one.

Not `Session.getBranch()`: that walk is bounded by the last compaction's retained window — the right
bound for model context, the wrong one for settings.

## Decisions worth naming

| | |
|---|---|
| no branch summarization | `navigateTree`'s optional LLM summary is a model call and engine-flavoured; moving a leaf is not |
| `targetId: string`, no root | "start from nothing" is a new session, not an emptied one |
| a no-op move | `ok: true`, writes nothing (pi journals a move as a `leaf` record), still emits the position |
| `leaf` records | withheld from `entries()` and refused as targets through ONE predicate, so "everything published is navigable" holds by construction |
| capabilities derived from a dispatch map | deliberately NOT done (#284's suggestion): `navigate: boolean` is one line, and the capabilities shape moved in #302 already |
| `state()` may reject | an unreadable chain has no answer for "what would a turn run with"; `entries()` keeps serving the journal. This is the one control-plane failure with no `error.code` channel — giving reads a code channel is a contract change, not this PR |

## Verification

`npm run lint && npm run typecheck && npm test` — 1276 passing. New coverage: the branch move (the
model sees the branch it moved to, not the one it left), settings/activations from an abandoned
branch not following the session back, lease contention, the dangling tool-result repair after a
move, the no-op, the wire round trip, and the corruption dispositions.


src/engines/pi/sessions.ts
```diff
@@ -112,6 +117,41 @@ async function reconcileInterruptedToolCalls(session: Session): Promise<void> {
...
+ * The entries on the session's ACTIVE path, root→leaf — what every last-wins read must walk.
+ * `getEntries()` is the whole TREE: once `navigate` can move the leaf, the journal still carries
+ * the abandoned branch, and reading it flat would run the session on a setting it moved away from.
+ * Deliberately NOT `Session.getBranch()`: that walk is bounded by the last compaction's retained
+ * window, which is the right bound for MODEL CONTEXT and the wrong one for settings — an override
+ * recorded before a compaction is a preference, and it still governs the session after one.
+ */
+export async function activePathEntries(session: Session): Promise<SessionTreeEntry[]> {
+  const entries = await session.getEntries();
+  const leafId = await session.getLeafId();
```
kid7st, line 133: **The walk reads the journal and the leaf non-atomically, in the order that makes a live run look like corruption.**

`getEntries()` returns a `.slice()` copy (both pi storages), and `getLeafId()` is read *after* it. Any append landing between the two — i.e. any turn in progress — yields a leaf newer than the snapshot, and the walk throws `session leaf "…" is missing from the journal`. Two lock-free callers hit this: `state()` (attach calls it every round while a run streams) and the **pre-lease** `set_thinking` gate, which then answers `boundary_command_failed` for what is actually `session_busy`. The throw is designed to mean "this session is corrupt"; here it means "someone was talking".

Read `getLeafId()` **first**, then `getEntries()`: the snapshot is then guaranteed to be a superset of the leaf's chain, and the throw regains its intended meaning.

src/engines/pi/session-control.ts
```diff
@@ -300,7 +320,10 @@ export function createPiSessionControl(options: CreatePiSessionControlOptions):
     async entries(session, opts): Promise<SessionEntries> {
       const opened = await sessions.openIfExists(session);
       if (!opened) return { entries: [] };
-      const all = (await opened.getEntries()).map(toSessionEntry);
+      const all = (await opened.getEntries()).filter(isNavigable).map(toSessionEntry);
+      // A leaf the journal does not hold is not reported: a client walking parentId from a dangling
+      // head would have no signal at all, and "the head is one of these entries" is the whole point
+      // of publishing it.
```
kid7st: **The comment describes a filter the code does not have, and trusts the port the rest of the PR refuses to trust.**

"A leaf the journal does not hold is not reported" — `leafEntryId` is assigned straight from `getLeafId()` with no membership check against `all`. Under pi's storages that read throws instead, so `entries()` fails wholesale, contradicting the neighbouring promise that `entries()` keeps serving the journal; under a swappable `PiSessionReader` — the exact reason `activePathEntries` re-validates the chain itself (sessions.ts:137-140) — it publishes a dangling head.

Pick one trust level: either drop `leafEntryId` when `all` does not contain it (making the comment true), or delete the comment and let the storage own it, consistently with the walk.

src/cli/commands/attach.ts
```diff
@@ -456,6 +456,26 @@ export interface AttachIo {
...
+ * compaction, leaves sibling branches behind), and printing the slice raw renders the abandoned
+ * branch interleaved with the live one as if it were one conversation. Reachability is computed
+ * backwards from the leaf; an entry whose parent lies BEFORE the slice is path-connected by
+ * construction (its ancestors were rendered in an earlier round), so only what the walk cannot
+ * reach is dropped. With no leaf reported, nothing is known to be off-path and the slice stands.
+ */
+export function activePathSlice(entries: SessionEntry[], leafEntryId: string | undefined): SessionEntry[] {
+  if (leafEntryId === undefined) return entries;
+  const byId = new Map(entries.map((e) => [e.id, e]));
+  if (!byId.has(leafEntryId)) return entries; // the leaf predates the slice: it is all one tail
```
kid7st: **The fallback's stated reason is false in the case the function was written for.**

`the leaf predates the slice: it is all one tail` holds when nothing branched. After a `navigate` backwards with no new turn yet, the leaf is *behind* the cursor and every entry in the slice is the abandoned branch — so the fallback renders exactly the interleaved abandoned branch `activePathSlice` exists to suppress, and the test (`activePathSlice(slice, "older")`) pins that as intent without naming the case.

Since an append always moves the leaf, "leaf not in slice" means nothing in the slice is on the active path: render nothing (or keep only what the slice cannot resolve, and say why).

kid7st (commented):
Thesis: `navigate` is the write verb for the tree `entries()` already publishes, and its real cost is that a movable leaf falsifies the flat last-wins reads — so the PR is as much about `activePathEntries` as about the command.

Findings below; one of them (the attach diagnosis) spans a line outside the diff, so it lives here.

**`src/session.ts:15-21` + `src/cli/commands/attach.ts:567` — a permanent, session-scoped failure enters the only in-repo client through the transport-error path**
`state()` now rejects on an unreadable chain, with no code channel; over HTTP that is a 500 from `attachRound`'s post-backfill sync, which `decideRound` classifies as a generic non-auth error → retry until the grace budget → exit with `the remote endpoint has been unreachable for ~Ns — check the serve`. The diagnosis names the transport for a condition that lives in the session record and will never recover on retry. Either surface the server's message on the first non-transient sync failure (a distinguishable outcome kind in `decideRound`), or have `attachRound` print the state error and continue instead of failing the round — the session is still readable through `entries()`.

kid7st:
All four findings addressed in 12441ed.

**1. Non-atomic read order (`sessions.ts`)** — real bug, and the sharpest of the four: `getEntries()` is a snapshot, so reading it before `getLeafId()` let any concurrent append produce a leaf the snapshot cannot contain, and the integrity throw fired on *a live turn* instead of a corrupt session. `state()` (attach polls it every round) and the pre-lease `set_thinking` gate were both exposed. Fixed by reading the leaf first — the snapshot is then a superset of the leaf's chain — in the walk and in `entries()`.

**2. `entries()` comment vs code** — the comment survived the deletion of the check it described. It is now true by construction thanks to (1), and says so.

**3. `activePathSlice` fallback** — correct: an append always moves the leaf, so a leaf *behind* the slice means every entry in it was appended and then abandoned. Renders nothing now; the test names the case instead of pinning the old intent. The `leafEntryId === undefined` case still returns the slice whole — that is an engine reporting no leaf, where "unknown" must not read as "off-path".

**4. `state()` rejecting → the transport diagnosis** — taken at the root rather than at `attach`. A read with no error-code channel was the anomaly; `state()` is TOTAL again and leaves the settings pair ABSENT on an unreadable chain (the shape a control-less deployment already answers with). Nothing is swallowed: the fault surfaces where codes exist — the next invoke's `failed` event (the harness build walks the same chain) and a boundary dispatch's `boundary_command_failed`. `attachRound`'s strict "sync failure closes the round" discipline is left untouched, and the test that pinned it stays green.

Issue #286, opened 2026-07-26
Session lifecycle: create / fork / clone / name / list have no plane


kid7st:
## What

A place to create, fork, clone, name and list sessions. Today a session comes into being only as a
side effect of `invoke`, and the rest have no home at all.

Companion to #284 (`navigate`), which deliberately left this out to keep that issue to one verb.

## Why

The control plane is explicit that this is not its job, and the reasoning is sound:

> Six commands; deliberately NO `prompt` — starting work is the data plane's definition.

That fits serving: a webhook or a chat message starts the work, and the control plane modulates a
run already in flight. Sessions are created by the data plane, and
`NO_SUCH_SESSION_CODE` exists precisely so a boundary mutation cannot mint a ghost record.

But that leaves operations which are neither starting work nor modulating a run:

| | today |
|---|---|
| list this agent's sessions | — |
| create an empty session (before any turn) | only via `invoke` |
| fork a session at an entry | — |
| clone a session | — |
| name a session | — |

None of these are "start a turn", so the data plane's boundary is not the reason they are missing —
they simply have no plane. For a serving deployment that is fine; nothing lists conversations. For
an interactive client it is most of the session-management surface, and it is the other half of what
`navigate` is for: navigation and forking are the two verbs that make a session tree usable, and one
without the other is a tree you can walk but not grow.

## Why it is small

Same story as #284 — these are primitives on `SessionStorage`, the `pi-agent-core` port already in
use: `getSessionName` / session naming, entry creation with an explicit parent (fork), and session
enumeration at the repository level.

## Shape

Two of them are per-session and fit `SessionCommand` beside `navigate`:

```ts
| { type: "fork"; fromEntryId: string }   // → new session id
| { type: "set_name"; name: string }
```

The rest are *about* sessions rather than *within* one, so they do not fit an interface keyed by
session id. A sibling repository reads more honestly than overloading the control plane:

```
GET  /control/sessions             → list
POST /control/sessions             → create (empty; no turn yet)
POST /control/sessions/:id/clone   → copy
```

Whether `create` here conflicts with "sessions are created by the data plane" is the one real design
question in this issue. Two readings, and I do not have a stake in which wins:

- it does conflict — an empty session is a ghost record, and a client should call `invoke` and let
  the first turn create it;
- it does not — the invariant is that a *boundary mutation* must not create a session implicitly,
  and an explicit `POST /sessions` is not implicit.

If the first reading wins, drop `create` and the rest still stands.

## Context

[...]

PR #310, opened 2026-08-03
feat(session-control): commands() — the names a composer completes

### LINKED ISSUE #287: Expose the agent's command list (`GET /control/commands`)
## What

A read that answers "what commands does this agent offer" — the definition's prompt commands, its
skills, and (once #285 lands) its extensions' commands.

```
GET /control/commands → Array<{ name: string; description: string; argumentHint?: string; source: string }>
```

## Why

An interactive client needs the list *before* the user types anything: a `/` in a composer opens a
completion list, and that list is either populated from the agent or it is empty. There is no way to
discover the set by trying — an unknown command is just text that reaches the model.

The information already exists on fastagent's side. `AgentDefinition` carries `skills`, and skills
are commands users invoke by name; the assembly step is the only place that knows the full set after
overrides and collisions have been resolved (`SkillCollision` is already tracked there, first-wins —
a client cannot reconstruct that from the directory).

## Why it is engine-neutral

"Which named invocations does this agent expose" is not a pi concept. Every engine with skills,
prompt templates or extensions has an answer; an engine with none returns `[]`, which is a complete
answer rather than a missing one. `source` is a free-form string precisely so engines can label their
own kinds without the contract enumerating them.

This is a read, so it needs no capability gate beyond being routed — though if reads become
discoverable (mentioned in #284), this is one of them.

## Relationship to #285

Extension-provided commands only exist if extensions load, so today this read would return the
definition's skills and nothing else. That is still worth having — and the two are independent: this
issue is about exposing what is assembled, #285 is about what gets assembled.

## Context

duang.ai, same client as #284 / #285 / #286. It calls pi's `get_commands` today and renders the
result as composer completions; that is the one call with no path forward at all, since unlike
lifecycle or navigation it cannot be reconstructed from entries or storage primitives.

Happy to send the PR.

kid7st:
## What

`SessionControl.commands()` — the names an agent exposes, so a client can populate a `/` completion
list. Closes #287.

```ts
interface AgentCommand { name: string; description?: string; source: string }
commands(): Promise<AgentCommand[]>;   // GET /control/commands
```

Sessionless like `capabilities()`, but **async**, and that is the whole design: the directory is the
agent *live* (fastagent re-reads it per turn), so a boot snapshot would advertise names the running
agent has already left behind. It reads `skills/` only — the full definition load's ② context walk
(every `AGENTS.md` from cwd to root) buys nothing for a name list.

## A listing, not a dispatch surface

The issue frames skills as "commands users invoke by name". In fastagent they are not: nothing in the
data plane expands `/name`, and a prompt is text. Rather than imply an invocation path that does not
exist, the contract says what is true — this LISTS, and what typing one means is the client's choice.

The one in-repo consumer now makes that choice: `fastagent attach` gained `/commands` (list), answers
a typed `/triage` with what that name actually is, and tells the user to name it without the slash.

## Notable decisions

| | |
|---|---|
| `source: string`, free-form | which kinds exist is an engine's business; pi's closed union would make `AgentCommand` a pi type. Field names still follow pi's `get_commands` so a client maps directly |
| `[]` is complete | an L1 agent (`createPiAgent({ model, instructions, tools })`) genuinely has no names. `commands` is therefore an optional hub option — but the directory constructor always wires it, where `[]` would be a lie |
| may reject | a definition the server cannot READ has no truthful degraded value. It carries no code (the read-side error-channel gap tracked on #286); a missing route is mapped to a distinguishable "predates the route" so skew never reads as a broken definition |
| one findings memo | a broken `SKILL.md` vanishes from the list, so this read must warn — through the SAME memoized reporter the turn path uses (keyed by the resolved definition dir), so a finding is announced once, by whichever reader sees it first. The "record without printing" door was removed: it could swallow findings for a caller that never printed |

## Verification

`npm run lint && npm run typecheck && npm test` — 1282 passing. New coverage: live re-read in both
directions (added / removed) plus first-wins collision resolution, a failed skill warned exactly once,
the wire read per call rather than cached at connect, the auth sweep derived from the mounted routes
(so a new route that forgets `guard` fails it), the rejection shape, and route skew.

### DIFF
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -482,6 +482,15 @@ accepted `abort` can still settle `completed`, and an accepted `steer`/`follow_u
 without its prompt being consumed, when the run finishes inside the window — acceptance is not
 outcome; the settlement is the truth.
 
+`commands()` lists what a `/` composer completes: `{ name, description?, source }` per named thing
+the definition exposes (`source: "skill"` today). It is a LISTING, not a dispatch surface — the data
+plane takes prompts as text and nothing expands `/name`, so what typing one means is the client's
+choice. It is read live and uncached (one full definition load per call), so a skill added while
+serving appears at once; `[]` means the agent exposes none. It is also the one read that can REJECT:
+a definition the server cannot read at all is a deployment fault with no truthful degraded value, and
+the rejection carries no stable code (remotely: an uncoded non-2xx → `ControlRequestError`). Wrap the
+call, and expect no `error.code` to branch on.
+
 Boundary mutations run between runs, under the SAME lease (`session_busy` while a run is active,
 retryable at idle):
 
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -140,12 +140,18 @@ Common options:
 function createPiAgentFromDefinition(
   dir: string,
   options: CreatePiAgentFromDefinitionOptions,
-): Promise<{ agent: Agent; definition: LoadedDefinition }>;
+): Promise<{
+  agent: Agent;
+  definition: LoadedDefinition;
+  readDefinition: () => Promise<LoadedDefinition>;
+}>;
 ```
 
 Load `persona.md`/`skills/` from `dir` (the agent dir) and assemble the pi prompt. `②` project context is sourced via pi's `loadProjectContextFiles({ cwd, agentDir: dir })` — the dir's own `AGENTS.md` plus every `AGENTS.md` walking `cwd` (option; default `dir`) up to root. Pass `cwd` to decouple the workspace (where tools operate, whose repo `AGENTS.md` is context) from the agent dir — `createPiAgentFromDir` always passes the resolved workspace, which is the directory fastagent was pointed at (the agent dir's parent when the agent was found one level inside it, the agent dir itself when you aimed straight at it).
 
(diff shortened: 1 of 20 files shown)


kid7st:
Both taken in 55462bf.

**Conceptual integrity — the §11 pi-surface table contradicted §5.1.2.** Correct, and it is exactly the failure mode of adding a contract section without re-reading the document that decides what the contract excludes. The row now reads `get_commands → Adapt to commands()`, and states the split so the next maintainer does not have to infer it: the definition-derived LISTING is included (it is the one thing a client cannot reconstruct), while pi's slash execution and presentation stay out — fastagent's data plane takes prompts as text and does not expand `/name`.

**Coverage illusion — attach's slash branch.** Also correct: ~35 lines of user-visible behavior sat inside an `rl.on("line")` closure where no test could reach it, and every claim in its comment block was unverified. Extracted as `answerSlashInput(trimmed, control, println)` and pinned for all five paths you named:

- a mistyped token → the reserved-prefix line lands FIRST (asserted before awaiting, so the "certain half does not wait on the remote read" property is real), then `[/aboort names nothing this agent defines]` — not a dump of every skill;
- `/triage summarize my inbox` → first-word token, so a name with arguments resolves;
- `/commands` → descriptions rendered, and no preamble (its whole answer is the read);
- empty list → `[this agent defines no names]`;
- a rejected read → `[command list unavailable: …]` rather than an unhandled rejection leaving the user with only the preamble.

The extraction is also what makes the async failure path reachable at all: it now returns the promise instead of firing two `.then` arms into the void.

Issue #289, opened 2026-07-26
Extensions need a way to ask the user: an `ask` event and an `answer` command


kid7st:
## What

A way for an extension-authored command to ask the user something and get an answer back:
an `ask` event on the session stream, and an `answer` command that resolves it.

Companion to #285. That issue is about extensions loading at all; this one is about what happens
the first time one of them speaks.

## Why it is not optional once #285 lands

pi gives extensions four ways to reach the user:

```ts
interface ExtensionUIContext {
  select(title: string, options: string[], opts?): Promise<string | undefined>;
  confirm(title: string, message: string, opts?): Promise<boolean>;
  input(title: string, placeholder?: string, opts?): Promise<string | undefined>;
  notify(message: string, type?: "info" | "warning" | "error"): void;
}
```

The first three **return a promise the extension awaits**. With no channel to carry them, an
extension that calls one does not degrade — it stops, mid-command, with the user seeing nothing.

That makes the two issues a pair in a specific order: loading extensions without this is worse than
not loading them, because "the command did nothing and never came back" is harder to diagnose than
"the command isn't there". If only one can land, this one should not be the one left out.

These are not exotic. A representative real extension — an auth-backup command — uses three of the
four in one flow:

```ts
await ctx.ui.input("New auth backup name:", "auth-backup-name")
await ctx.ui.confirm("Overwrite backup", …)
ctx.ui.notify(`Restored auth backup: ${name}`, "info")
```

## What is deliberately *not* being asked for

`ExtensionUIContext` has a second half — widgets (`placement: "aboveEditor" | "belowEditor"`), raw
terminal input handlers, a working-indicator config, an editor factory. Those are a TUI's layout
model, not "what the extension needs to express", and porting them would push terminal geometry
into a contract that has no terminal.

The dialog half is different: `select`/`confirm`/`input` are questions with answers, which is a
shape any client can render — a modal, a sheet, an inline card. Only the four dialog methods are in
scope here. A widget-shaped extension would want a different mechanism (data out, client decides
presentation), and that can wait for a case.

## Shape

An event and a command, mirroring the existing pairs:

```ts
type AskEvent = SessionEvent<"ask", {
  id: string;
  kind: "select" | "confirm" | "input";
  title: string;
  message?: string;      // confirm
  options?: string[];    // select
  placeholder?: string;  // input
  timeoutMs?: number;    // pi's dialogs support a timeout with a live countdown
}>;

type SessionCommand =
  | …
[...]

Issue #285, opened 2026-07-26
Definition should carry its own extensions, the way it carries skills


kid7st:
## What

Load `<agentDir>/extensions/` as part of the agent definition, the way `<agentDir>/skills/` is
already loaded, and pass the result through to the engine instead of `noExtensions: true`.

## Why

`src/engines/pi/session-builder.ts` currently disables pi's extension discovery outright:

```ts
noExtensions: true,
noPromptTemplates: true,
noContextFiles: true,
```

The comment gives the reason, and it is a good one — *"suppress pi's machine-global discovery … so
this runtime runs the same agent that gets served, not the authoring machine's pi setup on top."*
Reproducibility is worth protecting.

But the three flags are not equivalent, because two of them have a replacement and one does not:

| flag | replacement |
|---|---|
| `noContextFiles` | `loadAgentDefinition` reads the agentDir's AGENTS.md and its ancestors into `contextFiles`, injected as segment ② |
| `skillsOverride` | `definition.skills`, loaded from `<agentDir>/skills/` |
| `noExtensions` | **none** — extensions are dropped and nothing takes their place |
| `noPromptTemplates` | **none** |

So an agent directory carrying an `extensions/` folder is served with those extensions silently
absent. Machine-global discovery is correctly suppressed; the definition's own extensions are
collateral.

## The shape already exists

Skills are the precedent: not discovered from the machine, **declared by the definition**, loaded
from a known subdirectory, and handed to the engine as data. Extensions want exactly that treatment:

```ts
export interface AgentDefinition {
  …
  skills: Skill[];
  extensions: Extension[];   // from <agentDir>/extensions/, same rules
}
```

That keeps the property the flag was protecting. `<agentDir>/extensions/` travels with the
definition, so the same agent runs the same way locally and in the cloud — no host-kind branch, no
"local mode" that re-enables machine discovery and quietly makes the two diverge. The reproducibility
boundary stays where it is; it just stops cutting through the definition.

## Why this cannot be worked around downstream

`noExtensions` is set inside `session-builder.ts`. pi's loader does keep CLI-supplied extension
paths when the flag is on:

```js
const extensionPaths = this.noExtensions
  ? cliEnabledExtensions
  : this.mergePaths(cliEnabledExtensions, enabledExtensions);
```

— but that channel is fastagent's to use. A client spawning `fastagent serve` has no seam.

## Context

Building a desktop client (duang.ai) that is moving from spawning `pi --mode rpc` to attaching to
[...]

PR #313, opened 2026-08-04
docs(design): conformance levels — state locality and engine class are two deployment axes

kid7st:
## What

A design note establishing that **where a session's state lives, and which engine class serves it,
are two independent deployment choices** — plus the companion-text corrections that follow.

No code changes.

## Why now

Four issues in a row (#284 navigate, #287 commands, #286 fork/clone/name, #285+#289 extensions) are
the same request wearing different hats: *give me the surface pi's `AgentSession` has*. Two of them
shipped as adaptations onto the harness path, each with its own contract extension and its own
divergence from the native semantics — `commands()` is a listing here precisely because nothing in
this engine class expands `/name`, while on the other class the same read is a dispatch surface.

That is a reasonable price when the verb has to work on the portable path anyway. It is the wrong
answer when the request is "use the other engine class", and by the third verb it was worth stopping
to re-derive rather than continuing to patch.

## What the note decides

SPEC §6 already settled half of it, and I had been reading it as an axiom rather than a level:

> A resident stateful Agent can still conform to Agent Handler, but it does not satisfy portable
> conformance.

Permitted, cost named, level optional — and `test/spec-conformance.ts` already encodes it (`pair?()`
is an OPTIONAL subject capability). So the question "should `invoke` be built on `AgentSession`?"
conflates two axes:

| Axis | Values | Decides |
|---|---|---|
| State locality | `per-invoke` \| `resident` | SPEC MUST 6 — whether a turn may require the previous turn's process |
| Engine class | `harness` \| `session` | Whether `extensions/` does anything, whether `/name` dispatches, whether branch navigation can summarize |

## The measured surprise

The cell I had assumed impossible is the cheap one. `per-invoke` + `session`:

- shared assembly (resource loader + extension modules) **once**: ~3 ms
- a fresh `AgentSession` over the same jsonl, **per turn**: ~1 ms
- extension hooks fire every turn; one durable record accumulates across turns

Both engine classes read and write the same versioned session format (`{"type":"session","version":3,…}`),
verified in both directions with the leaf and parent chain intact — so this is a swap at **L0 alone**,
with `Agent`, `SessionControl`, the assembly ladder, tools and channels untouched.

Portability is therefore not the price of extensions. Residency is a separate choice with a separate
bill (~25 KB per idle session, transcript retention ≈ the record on disk, ~0.8 GB for 1000×100-turn
sessions), which only a resident posture pays.

## Companion text

- `AGENTS.md` — the "Stateless invoke" rule is re-scoped: it is the default level's invariant (what
  MUST 6, AgentCore and scaled channel hosts require of *that path*), not an axiom that overrides the
  contract. As written it would have vetoed something the SPEC explicitly permits — it vetoed it in my
  own reasoning for three rounds.
- `docs/design/core.md` §3 — names the two axes at the L0 rung where they are decided.
- `docs/deploy.md` — AgentCore's "no resident process" is stated as a constraint on the **agent**
  (MUST 6), not only on the host.

## Known gap recorded, not hidden

`SessionManager` buffers a NEW session's entries until the first assistant message, so a crash between
"the user asked" and "the model answered" loses the question; pi-agent-core's storage persists it
immediately. Any `session`-class implementation must close or state that before it serves a channel.

### DIFF
--- a/docs/design/conformance-levels.md
+++ b/docs/design/conformance-levels.md
@@ -75,8 +75,11 @@ The seam is **L0 only** ([core.md](core.md) §3: `createPiAgentFromHarness` is t
 
 ## 5. What each level owes
 
-**`per-invoke` owes MUST 6**: state reconstructible from the store, no process affinity, and — because
-fastagent's directory is the agent, live — a fresh definition read per turn.
+**`per-invoke` owes MUST 6**: state reconstructible from the store, no process affinity. Nothing more —
+in particular, "the directory is the agent, LIVE" is a fastagent product property, not part of this
+level: the harness path buys it by re-reading the definition every turn, while a `session`-class
+factory that shares one `ResourceLoader` across turns (what makes its per-turn cost ~1ms) trades that
+freshness away. Which one is in force is the factory's choice and must be stated by it.
 
 **`resident` owes an explicit lifecycle**: when a session is created, evicted, and rebuilt, plus a
 rebind path when the definition changes (an `AgentSession` snapshots its assembly at build). It also
--- a/docs/design/conformance-levels.md
+++ b/docs/design/conformance-levels.md
@@ -1,0 +1,111 @@
+---
+title: Conformance levels
+description: "Where a session's state lives is a deployment choice the SPEC already defines — not an implementation detail. Two axes (state locality × engine class), the postures that pin them, and what each owes."
+status: current
+---
+
+# Conformance levels
+
+## 1. Decision
+
+**Where a session's state lives is an explicit deployment choice, and the engine class that serves it
+is a separate one.** Neither is a property of "fastagent"; both are properties of a deployment.
+
+The protocol says so already. [SPEC](../SPEC.md) §6, under *Portable conformance (optional; required
+for Agents claiming serverless portability)*:
+
+> **No location dependence**: the Agent MUST NOT require multiple invocations with the same `session`
+> to land in the same process or instance. Session state must be reconstructible from external state.
+> **A resident stateful Agent can still conform to Agent Handler, but it does not satisfy portable
+> conformance.**
+
+A resident Agent is permitted, the cost is named, and the level is optional. `test/spec-conformance.ts`
+encodes the same shape: `pair?()` is an OPTIONAL subject capability, because MUST 6 is optional.
+
+So the recurring question — "should `invoke` be built on pi's `AgentSession` instead of its
+`AgentHarness`?" — is malformed. It conflates two independent axes.
+
+## 2. Two axes, not one
+
+| Axis | Values | What it decides |
+|---|---|---|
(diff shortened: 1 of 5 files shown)


docs/design/conformance-levels.md
```diff
@@ -0,0 +1,111 @@
...
+  Every method maps onto `AgentSession` natively, most of them more directly than onto the harness
+  (`dispatch`'s verbs are `steer`/`followUp`/`abort`/`compact`/`setModel`/`setThinkingLevel`/
+  `navigateTree`; `commands()` is the extension+prompt+skill registry). Nothing in the contract bends.
+- **The assembly ladder, the definition, tools, channels, schedules** — level-agnostic. They produce
+  inputs; the level decides who consumes them.
+
+## 5. What each level owes
+
+**`per-invoke` owes MUST 6**: state reconstructible from the store, no process affinity, and — because
+fastagent's directory is the agent, live — a fresh definition read per turn.
```
kid7st: §5 与 §2 相互矛盾, 且这个矛盾在后续 PR (#315) 的实现里已经被点名.

§5 把 “fresh definition read per turn” 绑给了 `per-invoke` 这个 level (“per-invoke owes MUST 6: …, and — because fastagent's directory is the agent, live — a fresh definition read per turn.”). 但 §2 的 per-invoke × session 格子的可行性恰恰建立在 “shared assembly once (~3 ms), then ~1 ms per turn” —— 即 ResourceLoader 跨 turn 共享、definition 是快照, 这正是它便宜的原因. 换言之: “live definition read” 只是 per-invoke × harness 的产品性质, 不是 per-invoke 这条 axis 的义务.

#315 的 session-invoke.ts 头注释已经把这条讲得很明白: “a shared loader is a SNAPSHOT, so 'the directory is the agent, live'… is not free here. A factory that wants it must rebuild or reload the loader; one that shares it trades definition freshness for the ~1 ms turn cost.” —— 与本 §5 的归属直接冲突.

后果: 读者会带着 “只要 per-invoke 就自动有 live definition” 的理解去看 #285 (extensions 加载), 而 session class L0 默认不给这个保证 —— 恰好是需要在装配层显式表态的地方. 这份 design doc 应该正确划分归属, 才能给 #285 的接续实现留一个可引用的锚.

建议方向: 把 “fresh definition read per turn” 从 “per-invoke owes” 这句里拿出去, 单独作为 fastagent 产品层的、harness class 现有行为的陈述; 再明确说 per-invoke × session 的 factory MUST state whether it preserves definition freshness (rebuild loader) 还是 trades it away (share loader).

kid7st (commented):
Thesis: 本 PR 立一个概念 — State locality 与 Engine class 是两个独立的 deployment axis, 且 per-invoke × session 这个格是可行且便宜的 (~3 ms 装配 + ~1 ms/turn), SPEC §6 早已允许. 概念清晰.

以下 finding 能锚到 diff 行, 已作为 inline comment 发出.

PR #315, opened 2026-08-04
feat(engines/pi): an L0 for the session engine class, at per-invoke locality

kid7st:
> Stacked on #313 (the design note it cites). Review that one first; this branch contains it.

## What

An L0 for the **`session` engine class** — `createPiAgentFromSession` — that builds a fresh
pi-coding-agent `AgentSession` over the session's jsonl record **per invoke** and discards it.

Same state locality as the reference harness path (SPEC MUST 6 holds), different engine surface. It
is the substrate #285 (extensions) and #289 (extension dialogs) need; neither is implemented here.

**No existing path changes.** Nothing in `src/` constructs this agent yet — the L1 assembly is
deliberately not in scope, so the class can be judged before it is wired anywhere.

## Proven, not asserted

The conformance suite runs against it with `pair()` provided, so MUST 6 is *asserted*: two
independently assembled instances share only the record on disk, and instance B's turn sees instance
A's. Beyond the shared suite:

| Test | Pins |
|---|---|
| `an extension command runs instead of reaching the model` | THE point of the class — `/ping` dispatches, the model never sees it, the turn still settles |
| `an image prompt reaches the model` | images go through the same conversion the harness path uses, rather than being silently dropped |
| `a fastagent tool sees this session and this turn's activation` | the turn context is bound (missing, it degrades silently: `wake` writes nowhere, `search_tools` no-ops) |
| `nothing resident survives a turn` | a *different* session object per turn (identity, not call count) and one record carrying both |
| `a factory that throws becomes a failed EVENT` / `session_busy` | this module's own two failure branches |
| `an auto-retried failure … settles on the recovered turn, once` | `agent_end{willRetry}` is not the settle |

## The one thing that looked broken was my probe

The pull-driven cancellation test failed for two days' worth of suspicion before yielding a plain
answer: `tool_started` is emitted **before** pi calls the tool's `execute`, so a consumer that walks
away on that event cancels in the window between the two — and `execute` then receives a signal that
is **already aborted**, which never fires its `abort` event again. The gate tool in the test only
registered a listener, so it hung forever, `AgentSession.abort()` could never reach idle, and the
whole thing presented as a broken cancellation path in this L0.

With the tool checking `signal.aborted` first, cancellation is correct end to end: `return()` settles
in ~1 ms and the in-flight tool stops. Two things came out of that:

- a `doorFired` teardown workaround was deleted — it only ever compensated for the broken probe, so
  teardown is back to the harness L0's plain awaited abort;
- the trap is real for anyone writing a fastagent tool, so it is now documented on
  `ToolContext.signal`: check `signal.aborted` first, then listen.

## Shared, not duplicated

- `EventQueue` and `toPiPromptOptions` are exported from `invoke.ts` — both engine classes have pi's
  two-port shape (a promise for the turn + a subscription for the stream) and the same image path.
- The session-class tool bridges moved out of `session-builder.ts` into `session-bridge.ts`, now used
  by both of that class's consumers (the resident TUI runtime and this per-invoke L0) — one source of
  truth for how a fastagent tool sees the session and the activation chain, whichever class drives the turn.

## Verification

`npm run lint && npm run typecheck && npm test` — 1308 passing, 1 todo.

## Scope, stated rather than implied

This lands the engine seam and its conformance — **not** a wired serving path. Tracked as #321: the
assembly (models/auth/tools/extensions → a session factory), the control-plane wiring
(`SessionObserver` + run controls, which `/control/*` consumes), and the export decision, which
belongs with the assembly rather than before it. Both gaps are stated in the module header and in
`conformance-levels.md` §4/§5; the issue is what keeps them from becoming the default.

### DIFF
--- a/src/engines/pi/session-invoke.ts
+++ b/src/engines/pi/session-invoke.ts
@@ -60,16 +60,12 @@ import {
  * {@link CreatePiAgentFromSessionOptions.onExtensionError}, which is forwarded every failure; binding
  * pi's slot directly would instead be dropped every turn with no signal.
  *
- * PASS `sessionStartEvent: { type: "session_start", reason: "resume" }`: the record always exists
- * before the session is built (the binder ensures it), and binding emits this event on EVERY turn —
- * pi's default `reason: "startup"` would tell an extension that turn 500 of a conversation is its
- * first, so anything that greets, seeds or logs on startup would do it every turn.
- *
- * BIND TO AN EXISTING RECORD — use `sessionRecordBinder` (session-record.ts), do not let
- * `SessionManager` open a file of its own. Not a style preference: a SessionManager that starts a
- * fresh file BUFFERS everything until the first assistant message, so a crash between "the user
- * asked" and "the model answered" loses the question and the file is never created. The binder is
- * where that rule is implemented; this contract is where it is required.
+ * BIND TO AN EXISTING RECORD — spread `sessionRecordBinder`'s result (session-record.ts) into
+ * `createAgentSession`, do not let `SessionManager` open a file of its own. Not a style preference:
+ * a SessionManager that starts a fresh file BUFFERS everything until the first assistant message, so
+ * a crash between "the user asked" and "the model answered" loses the question and the file is never
+ * created. The binder carries the matching `sessionStartEvent` too, so the two rules cannot come
+ * apart; both are implemented there, and this contract is where they are required.
  */
 export interface CreatePiAgentFromSessionOptions {
   sessionFactory: (sessionId: string) => Promise<AgentSession>;
@@ -276,6 +272,10 @@ export function createPiAgentFromSession(options: CreatePiAgentFromSessionOption
       // one, release the object. Written once because the three exits differ only in how far setup
       // got — three hand-written subsets drift the moment a line moves between them.
       let unsub: (() => void) | undefined;
+      /** Whether THIS turn bound the extensions — i.e. whether a `session_start` was emitted for it.
+       *  `hasHandlers` cannot answer that: the runner and its handlers exist from construction, so it
+       *  would send a shutdown for a lifecycle that never began (the cancel-during-build path). */
+      let bound = false;
       const teardown = async (): Promise<void> => {
         try {
           await session.abort();
@@ -290,10 +290,9 @@ export function createPiAgentFromSession(options: CreatePiAgentFromSessionOption
         try {
           // Extensions saw a `session_start` when this turn bound them; without its counterpart they
           // would see a birth every turn and never a death, so anything acquired there (a handle, a
-          // timer, a subscription) would live until the process exits.
-          // pi's own helper is not on the package's public surface, and its whole body is this
-          // guard plus the emit (runner.js: `hasHandlers` → `emit`) — both public.
-          if (session.extensionRunner.hasHandlers("session_shutdown")) {
+          // timer, a subscription) would live until the process exits. Only when this turn actually
+          // bound them: a death for a birth that never happened is the same asymmetry mirrored.
+          if (bound && session.extensionRunner.hasHandlers("session_shutdown")) {
             await session.extensionRunner.emit({ type: "session_shutdown", reason: "quit" });
           }
         } catch (error) {
@@ -345,6 +344,7 @@ export function createPiAgentFromSession(options: CreatePiAgentFromSessionOption
             }
           },
         });
+        bound = true;
       } catch (error) {
         // Binding is this engine class's own setup step, and its failures obey the same rule as the
         // factory's: an EVENT, never a thrown iteration — and the session still gets torn down.
--- a/src/engines/pi/session-invoke.ts
+++ b/src/engines/pi/session-invoke.ts
@@ -325,6 +325,11 @@ export function createPiAgentFromSession(options: CreatePiAgentFromSessionOption
 
       const queue = new EventQueue<AgentEvent>();
       let settled: AssistantMessage | undefined;
+      // SUBSCRIBED BEFORE BINDING (below): pi emits `session_start` inside `bindExtensions`, and a
+      // handler's `sendMessage` lands as a `message_end` right there — a subscription armed
+      // afterwards would drop it, which is the silence this projection exists to prevent, in the one
+      // case `reason: "startup"` exists to enable (an extension that greets or seeds on a
+      // conversation's first turn).
 
       // EVERY extension-dispatch failure, not just a command handler's: pi reports them all here,
       // and one that vanished would be a silent failure inside a turn we are about to call
@@ -336,6 +341,26 @@ export function createPiAgentFromSession(options: CreatePiAgentFromSessionOption
       // the assembly's stay as they were — but `onError` is a single slot, not a merge, so this L0
       // CLAIMS it (stated in the factory contract above).
       const consumingErrors: string[] = [];
+      unsub = session.subscribe((event) => {
+        if (event.type === "agent_end") {
+          // `willRetry` means pi will run again for this same prompt — an auto-retry, not a settle.
+          // Taking the message from a retried end would classify the turn on an error pi is about
+          // to recover from.
+          if (!event.willRetry) settled = lastAssistant(event);
+          return;
+        }
+        const projected = projectSessionEvent(event);
+        if (!projected) return;
+        if (projected.type === "retrying") {
+          // Same reason the harness L0 logs it: the event only reaches an ATTACHED consumer, while
+          // an operator tailing logs must also see a backoff that would otherwise read as a hang.
+          log.warn(
+            `[fastagent] retry ${projected.attempt}/${projected.maxAttempts} in ${projected.delayMs}ms (session ${scope.session}): ${projected.reason}`,
+          );
+        }
+        queue.push(projected);
+      });
+
       try {
         bound = true;
         await session.bindExtensions({
@@ -362,25 +387,6 @@ export function createPiAgentFromSession(options: CreatePiAgentFromSessionOption
         yield { type: "failed", details: failureDetails(error), retryable: false };
         return;
       }
-      unsub = session.subscribe((event) => {
-        if (event.type === "agent_end") {
-          // `willRetry` means pi will run again for this same prompt — an auto-retry, not a settle.
-          // Taking the message from a retried end would classify the turn on an error pi is about
-          // to recover from.
-          if (!event.willRetry) settled = lastAssistant(event);
-          return;
-        }
-        const projected = projectSessionEvent(event);
-        if (!projected) return;
-        if (projected.type === "retrying") {
-          // Same reason the harness L0 logs it: the event only reaches an ATTACHED consumer, while
-          // an operator tailing logs must also see a backoff that would otherwise read as a hang.
-          log.warn(
-            `[fastagent] retry ${projected.attempt}/${projected.maxAttempts} in ${projected.delayMs}ms (session ${scope.session}): ${projected.reason}`,
-          );
-        }
-        queue.push(projected);
-      });
       try {
         // prompt() resolves when the turn is accepted-and-run; waitForIdle() closes the window in
         // which pi is still draining its own queues (a follow-up, a retry), so the terminal below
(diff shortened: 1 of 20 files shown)


kid7st:
On the recurring `ToolContext.signal` note (flagged as a drive-by): **intentional, and I would rather it ride here than go separately.**

It is not a doc improvement someone thought of — it is the finding this branch produced. The pull-driven cancellation test failed for a long stretch and looked like a broken `AgentSession.abort()`; the actual cause was that `tool_started` is emitted *before* pi calls `execute`, so a consumer cancelling on that event hands the tool a signal that is **already aborted**, which never fires `abort` again. My gate tool only listened, so it hung forever.

That trap belongs to every fastagent tool author on both engine classes, and the two places a tool author reads are `ToolContext.signal`'s docstring and the api-reference paragraph — so both say it now. Splitting it out would separate the warning from the reproduction that justifies it.

Same PR, same origin: the `sendMessage`-is-on-the-API-not-the-command-context fix, and `AgentToolResult` having no `isError` field (a tool reports failure by throwing). All three were fixtures I wrote from assumption instead of from the type, and each cost real debugging time — which is itself the argument for writing them down where the next person looks.

src/engines/pi/session-invoke.ts
```diff
@@ -0,0 +1,489 @@
...
+/**
+ * pi's label for the dispatch that can END A TURN with no model call — the one whose failure is
+ * therefore the turn's outcome rather than a warning beside it.
+ *
+ * `input` is deliberately NOT here, despite also being able to consume a turn (`action: "handled"`):
+ * a handler that THROWS does not consume it. pi catches, records `event: "input"`, and continues the
+ * loop, so the turn reaches the model and settles on its answer — which makes that failure exactly
+ * the kind that must warn beside a delivered result rather than overturn it.
+ */
+const TURN_CONSUMING_DISPATCH = new Set(["command"]);
```
kid7st: `TURN_CONSUMING_DISPATCH` 破了本文件其他地方强制的 drift-detection 纪律.

本文件对 pi 类型加成员的漂移有一整套防线: `DELTA_CHANNELS` 和 `SESSION_EVENTS` 用 `satisfies Record<..., "project"|"ignore">` 保证 pi 的联合类型缩窄/扩宽任一都会挂 typecheck; `projectProjectable` 的 default 写了 `const unhandled: never = event`; 头注释反复强调 "a bare `default:` waves this drift through".

但 `TURN_CONSUMING_DISPATCH` —— 决定一次 extension dispatch 失败究竟是**这次 turn 的判决**还是**陪跑的告警**这个语义分类器 —— 是一句裸的 `new Set(["command"])`, 键的静态类型是 `error.event: string`. pi 若加一个新的、能结束 turn 的 dispatch verb (或把 `"command"` 改名), 本文件会静默地归为 non-consuming, 于是一次 extension 代码抛出并 consume 了 turn 的失败会给 caller 报 `completed` —— 正是上面两张 map 存在就是为了拦截的那种漂移.

这不是假想: `TURN_CONSUMING_DISPATCH` 上方的头注释已经手工排除了 `"input"`, 说明 pi 侧的 dispatch taxonomy 是真实存在的多成员集合, pi 是这些成员的宿主.

建议方向:
- 如果 pi 导出了 dispatch label 的枚举 / 联合类型, 照 `SESSION_EVENTS` 的形状写 `Record<DispatchLabel, "consuming"|"non-consuming"> satisfies …`, 让新成员触发 typecheck 失败.
- 如果 pi 没导出, 走 `turn-plumbing.ts` §2 已经开的先例 (那里已有一条 "Upstream ask: a first-class retryable/kind on pi's terminal error would retire the prose path entirely"), 在本处也留一条 upstream ask + 明写 "pi 扩此集合的沉默失败模式是什么", 至少让下一个读者知道该盯什么.

test/session-engine-conformance.test.ts
```diff
@@ -0,0 +1,1155 @@
...
+    const agent = createPiAgentFromSession({ sessionFactory });
+    for (const text of ["first", "second"]) {
+      for await (const e of agent.invoke({ session: "s" }, { text })) void e;
+    }
+    // Every turn resumes a durable record — including the first, which resumes an empty one — and
+    // the death pairs the birth: the record is picked up again, the process is not ending.
+    expect(seen).toEqual(["start:resume", "shutdown:resume", "start:resume", "shutdown:resume"]);
+  });
+
+  it("a binding that fails AFTER the birth still reports the death", async () => {
```
kid7st, line 471: `bound = true` 的悲观预设有代码注释, 但只测了 AFTER 一半.

本测试 (AFTER) 用 monkey-patch 让 bindExtensions 先跑真的 (session_start 已 emit), 再抛错, 断言 seen === ["start", "shutdown"].

`session-invoke.ts` 里 `bound = true; await session.bindExtensions(...)` 的语义是: **即使 bindExtensions 在 emit session_start 之前就抛错, 也会 emit shutdown** —— 代码注释明写这是刻意悲观 ("a shutdown for a birth that only partly landed is cheaper than a handler holding what it acquired until the process exits").

问题: 没有测试 pin 这个 trade. 现在的 AFTER 测试 与 "把 `bound = true` 挪到 bindExtensions 成功之后" 这个错误重构 (即取消悲观, 只有 emit 成功才 fire shutdown) 都能过 —— 于是 "早期抛错也 fire shutdown, 换取不泄漏 handler 持有的资源" 这个决定, 在测试面上是不可见的.

建议方向: 加一个测试, 让 monkey-patched bindExtensions **不调用真的 bindExtensions, 直接抛** (相当于 emit session_start 之前就失败), 断言 seen 里包含 shutdown (即 ["shutdown"] —— start 从未进过 handler, shutdown 是悲观触发的). 那条 pessimism 就 pin 住了.

kid7st (commented):
Thesis: 本 PR 加了 `createPiAgentFromSession` —— session 引擎类的 L0. 它跟 harness L0 同 locality (per-invoke, SPEC MUST 6), 换的是引擎表面 (extensions / 命令派发). 配套动作: `sessionRecordBinder` 强制 "绑定既有 record" 关掉 pre-first-assistant 崩溃丢用户提问的窗口; `PiRecordLocator` port 把 layout+path 查询在 store 上封住; 共享 turn plumbing 从 `invoke.ts` 抽到 `turn-plumbing.ts`, session-class 的两个 tool bridge 从 `session-builder.ts` 抽到 `session-bridge.ts` —— 两个 L0 谁也不 import 谁. Not wired into a serving path (assembly + control-plane 归 #321). 概念清晰.

---

**PR body §"Shared, not duplicated" 与代码不一致.** PR 描述里写:

> The activation chain is keyed by the `AgentSession` object, so two bridges over one session serialize against each other instead of each holding a private (and therefore useless) chain.

但代码里 `sessionToolActivation` (`src/engines/pi/session-bridge.ts:44`) 是 `let chain: Promise<string[]> = Promise.resolve([]);` 在函数体内, 每次调用都返回一条新链 —— **per-bridge, 不是 per-session**. 代码里紧跟的注释 ("One chain per BRIDGE, which is sufficient because each consumer builds exactly one bridge per session object … nothing constructs [a second bridge over the same session]") 才是正确描述, 与 PR body 直接冲突.

读者若照 PR body 理解, 会以为对同一 session 构造两个 bridge 会自动 serialize —— 实际不会 (代码也不需要处理这个 case, 因为没人这么用). 不是代码 bug, 是 PR 叙事与代码的分歧. 建议方向: 改 PR body, 把这半句换成代码注释的措辞, 或者直接删掉 —— 该 bullet 想表达的 "shared not duplicated" 本来说的是 "两个 helper 的源码共享", 不是运行时的 chain.

src/engines/pi/session-invoke.ts
```diff
@@ -0,0 +1,508 @@
...
+ * L0 for the `session` ENGINE CLASS: an Agent whose every turn builds a fresh pi-coding-agent
+ * `AgentSession` over the session's durable record and discards it when the turn ends.
+ *
+ * Same state locality as the harness L0 (`invoke.ts`) — per-invoke, so SPEC MUST 6 holds and the
+ * record is the only continuity — but a different engine surface: extensions, `/name` dispatch,
+ * branch operations. The two axes are independent; see design/conformance-levels.md.
+ *
+ * Why per-invoke is affordable HERE, where a resident `AgentSession` was assumed: the expensive half
+ * of that class is the `ResourceLoader` (extension modules, skills), which is built ONCE with the
+ * assembly and shared across turns. What is per-turn is binding a session object to a record.
```
kid7st, line 11: 问题：这个 L0 把共享 `ResourceLoader` 定义为“每轮只重建 session、record 是唯一连续性”的前提，但 `bindExtensions()` 会调用 `ResourceLoader.extendResources()`，而该方法把 `resources_discover` 返回的 skill/prompt/theme 路径累加进共享 loader，且不会移除上一轮或另一 session 的路径。于是 session A 动态发现的资源会留在进程内并进入后续 session B；并发的不同 session 还会共同修改同一 loader。实际状态局部性不再是 per-invoke，结果取决于该进程先前服务过哪些 session，和 MUST 6 / “record is the only continuity” 的核心语义冲突。

建议方向：不要在不同 session/turn 间共享会被 `bindExtensions()` 修改的 loader；为每次 factory 调用提供从只读 assembly 快照派生的独立 loader，或先在 pi 上提供可 clone/reset 的 extension-resource 层。补一个按 session 返回不同 `resources_discover` 路径的双-session 测试，断言资源不会串线。

src/engines/pi/sessions.ts
```diff
@@ -201,7 +239,10 @@ export function jsonlSessionStore(options: { dir: string; cwd?: string }): PiSes
   };
 }
 
-/** Injective filename-safe encoding: [A-Za-z0-9._-] verbatim, the rest %-escaped. */
+/** Injective filename-safe encoding: [A-Za-z0-9._-] verbatim, the rest %-escaped. THE encoding for
```
kid7st, line 242: 问题：注释把 session id 编码定义为 injective，但实现对 UTF-16 code unit 使用可变宽十六进制；不同 opaque id 会碰撞，例如 `"\u0100"` 和 `"\u0010" + "0"` 都编码成 `%100`。两个调用方 session id 因而可能落到同一 jsonl，直接破坏本 PR 所依赖的“一 id 一 durable record”语义；新增的 channel-shaped ASCII 测试覆盖了路径字符，却没有验证 opaque/Unicode id 的唯一性。

建议方向：改用无歧义的字节级编码（例如 UTF-8 后每字节固定两位十六进制/百分号编码），并增加 Unicode、控制字符与构造碰撞对的测试。

kid7st (commented):
本次改动新增 per-invoke `AgentSession` L0：每次 invoke 从 durable record 构造并销毁 session，同时复用 assembly 资源，以在满足可移植状态局部性的前提下提供 extensions / commands / branch 能力。

kid7st:
Closing without merge. Not because the code is wrong, but because the shape of the deliverable is.

Two rounds of review here found different, real, load-bearing invariants each time (round 1: dispatch classifier drift + BEFORE-birth pessimism test; round 2: `ResourceLoader` cross-session contamination + `encodeSessionId` not injective). Each round's fixes were correct. The pattern isn't: it's a contract without a consumer, whose review surface is open-ended by construction. Every imagined consumer discovers another "the factory MUST …" clause the L0 needs.

The header currently carries ~4 paragraphs of "the factory MUST …" prose enforced by nothing typechecker-visible. The conformance suite claims MUST 6 is "proven, not asserted" but its fixture only exercises the sub-space where the two round-2 pathologies don't fire. That's structural, not fixable with another patch commit — see the retrospective in the private review thread for the full derivation.

**Where the code goes:**

- New branch `feature/session-engine` (branched from this branch's HEAD, all 42 commits carried forward).
- Will be re-opened as ONE PR combining #315's L0 + #321's factory:
  - `session-invoke.ts` / `session-record.ts` / `session-bridge.ts` / `turn-plumbing.ts` mostly unchanged (they earn their place; the L0 header prose gets slashed because a real factory removes the need to write requirements at an imaginary one)
  - a real factory built on top of `createPiAgentFromDefinition`, choosing ONE loader strategy in code (not prose)
  - `encodeSessionId` fixed to be actually injective (UTF-8 bytes, fixed-width %XX)
  - conformance suite binds against the real factory, not a fixture that dodges the pathologies

**What this unblocks:** #321 closes with the same PR. #285 (extensions loading) + #289 (ask/answer) become the immediate next batch because they can land on the real factory instead of waiting for wiring.

**Docs already on main from #313** (`docs/design/conformance-levels.md`) stay — the concept is right, and the round-1 review that corrected §5 landed there.

PR #327, opened 2026-08-07
feat(feishu): resolve a quoted message's reply chain to its root

Sukitly:
Part 1 of 3 in the thread-context work (#327 reply chain → #328 session inheritance → #330 room-buffer fold). This PR stands alone; the other two build on it.

## Problem (field-observed on Lark)

A user posts an image, the agent answers with a card, and a topic opened on that card asks about the image. The referent path read exactly ONE link — the card — so the model saw `[image]`'s placeholder two hops away and reported it could not see any picture.

Quoting a reply points at one link of an exchange. Reading only that link resolves the pointer halfway.

## The walk

The referent path now resolves the reply chain ABOVE the replied-to message: each ancestor fetched, decoded (cards included, via the #324 branch), attributed (`from you, the agent` / `from app …` / `from user …`), and rendered oldest-first — a transcript of the exchange the user is pointing into.

```text
[replied-to message (msg om_card, from you, the agent): 需要先确认两项…]
[reply chain above it, oldest first:
(msg om_ask, from user ou_alice): 看看这个 [image]]
```

**The bound is structure, not a picked number.** Every Feishu reply threads back to a root; the walk ends where the platform says the chain ends. The guards each answer one failure mode:

| Guard | Behavior |
|---|---|
| 8-ancestor cap | IO guard (one serial `getMessage` per hop; field chains are 1–3). Trips **visibly**: `(…the chain continues above this point)` |
| Cycle | Ends the walk silently — corrupt platform data, and ending is the whole fix |
| Unreadable ancestor | Keeps the part already fetched; a warn, not a marker — nobody pointed at it |
| Any author | The chain is the platform's structure, not a conversation the agent took part in — it walks through other bots' messages and attributes them honestly |

## Ancestor attachments: buffered, never primary

A chain ancestor's images/files join the **buffered** tier, deduplicated against the context buffer's refs by message-scoped identity. Nobody pointed at them, so an unloadable one costs the standard missing-attachment note — never the turn. The referent's own resources stay primary (fail-fast), unchanged.

This is the half that fixes the field case: the image behind the card now reaches the model as a background vision image, and its loss degrades instead of erroring.

## Relation to the hop removed in dc6f7da

A one-hop version of this walk was built and removed as a memory substitute. The questions that killed it — how many levels? deduplicated against the session how? an image two levels up as prompt text how? — are answered here by a **narrower job**, not a better hop: pointer resolution bounded by the chain's own root, attachments on the degradable tier, and history left to session inheritance (`participant-model.md` §8 — #328 and #330). §8's rung-2 entry now records both halves of that story.

## Tests (1336 green: lint / typecheck / test)

Own-card topic resolves the chain to the ask; another bot's card is walked but not self-attributed; cap at 8 with the visible marker and oldest-first order pinned; cycle terminates with the ancestor's image riding the buffered tier (manifest attribution included); an unloadable chain attachment yields the note while the turn runs; an unreadable ancestor keeps the fetched part. Mutation-verified: disabling the walk fails 6 tests.

### DIFF
--- a/docs/design/core.md
+++ b/docs/design/core.md
@@ -509,8 +509,11 @@ a stable hand-authored surface. What is platform-different:
   place's next answered turn. The Telegram consume invariant carries over: peek at dequeue, commit only
   on `completed`, and retain failures plus messages arriving in-flight. Non-`user` senders are dropped.
   Summon matches the `mentions` array by the bot's open_id (fail-closed until resolved). A reply summon
-  carries only `parent_id` — the referent's content and attachments are fetched as primary input;
-  buffered attachments are background input and degrade per resource.
+  carries only `parent_id` — the referent's content and attachments are fetched as primary input, and
+  the chain ABOVE the referent is resolved as context (oldest-first; one shared text budget;
+  ancestors' attachments join the buffered tier under its cap; a walk ending short of the root leaves
+  a visible truncation line — participant-model.md §8); buffered attachments are background input and
+  degrade per resource.
 - **Ingress is an onboarding-time app choice.** `add feishu|lark` asks for WebSocket or webhook and
   writes the corresponding transport-specific factory into the channel module. WebSocket needs only App ID/Secret, skips token capture,
   tunnel, Request URL registration, and platform crypto; the official SDK authenticates the outbound
--- a/docs/design/participant-model.md
+++ b/docs/design/participant-model.md
@@ -238,7 +238,7 @@ A thread must start from something. Four rungs, increasing in cost:
 | Rung | Mechanism | Gives the thread |
 |---|---|---|
 | 1 | referent anchor, truncated | the followed-up message, cut at some display-sized bound |
-| **2** | **referent anchor, bounded by the platform** | **the followed-up message in full (`REFERENT_MAX_CODE_POINTS`), plus its reply chain up to the root** |
+| **2** | **referent anchor, bounded by the platform** | **the followed-up message in full (`REFERENT_MAX_CODE_POINTS`)** |
 | 3 | seed injection | the room's last N exchanges, in the thread's first prompt |
 | 4 | session fork | the room's entire history, reasoning and tool results included |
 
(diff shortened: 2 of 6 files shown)


Sukitly:
**Review response** — triage and what changed in 2f4b999.

| Item | Decision | Resolution |
|---|---|---|
| Chain closure = history, re-injected per turn | **accepted the cost, kept the walk** (product call: an unconditional, stateless guarantee that what a quote points into is visible) | The cost is now *bounded and documented*: ancestor text shares ONE further `REFERENT_MAX_CODE_POINTS` across the whole chain — a walk costs at most one more referent per turn. §8 replaces the "questions dissolved" claim with the true accounting: text cannot be deduplicated from a layer that cannot read the session, and that price is carried deliberately. |
| Truncation indistinguishable from the root | **accept** | Every short-of-root ending (cap, budget, unreadable, cycle) leaves the same neutral truncation line; cycles also warn the operator (temporally impossible ⇒ corrupt data). The test that pinned "no marker" now pins the marker. |
| `BUFFER_ATTACH_MAX` bypassed | **accept** | Chain refs and buffer refs share the tier's one budget per kind; chain refs take slots first (direct upstream of the pointed-at message outranks ambient discussion); drops are counted into the missing-attachments note. |
| Docs promise the chain for all channels | **accept** | Rung 2's ladder row is channel-neutral again; the chain is scoped to Feishu/Lark (Telegram embeds exactly one `reply_to_message`; Slack has no chain); core.md + feishu.md aligned: direct referent primary, ancestors degradable context. |
| *(self-raised)* ancestor text used the fidelity bound | **accept** | Folded into the shared-budget fix above — the fidelity bound stays reserved for the message the user pointed at. |

Mutation-verified: marker −4 tests, budget −1, cap −1, chain-first priority −1.

kid7st (commented):
Thesis: extends the Feishu channel's reply-referent resolution from a single hop to the full reply chain up to its root. Ancestor attachments ride the buffered tier; primary semantics (the referent itself + the user's direct attachments) are unchanged. Four boundaries (8-ancestor IO cap / referent-sized shared text budget / cycle / unreadable ancestor) all render the same short-of-root marker, so the model cannot mistake a cut chain for a root-terminated one. Concept / naming / layering / responsibility-of-errors reviewed dimension by dimension; docstrings match the code; 8 scenario tests cover the failure paths, and the PR body cites mutation verification (disabling the walk fails 6 tests).\n\nNo findings.

kid7st (commented):
Second pass at the same HEAD (branch did not move since the first review). Took a fresh angle to try to break through 'no findings': re-examined the referent-direct-attachment fail-fast vs. referent-message soft-fail asymmetry (pre-existing, PR body explicitly names the choice, defensible reasoning: pointer resolvable vs. unresolvable is a legitimate distinction), the mentions cast at the parse-walk seam (defensiveness is parseContent's job, not this diff's), the cycle-guard interaction with the current summoning message (impossible under temporal acyclicity), and capMerge accounting (correct — dropped counts only cap overflow, not primary-dedup or chain/buffer twin merges). Nothing new surfaces.\n\nNo findings.

PR #328, opened 2026-08-07
feat(pi+feishu): a new thread session starts from what the room knew

Sukitly:
Part 2 of 3 in the thread-context work (#327 reply chain → **#328 session inheritance** → #330 room-buffer fold). Stacked on #327; the diff shows only this PR's changes.

## Problem

`participant-model.md` §5 has said from the start: *"A thread starts from what the room knew and keeps its own history … separate, not amnesiac."* The implementation answered with amnesia — a topic opened on a room answer got an **empty session**, and no amount of referent-anchoring (rung 2) can carry images, tool results, or the exchanges nobody quoted.

## The mechanism: fork-on-first-open

Channel-neutral by construction. A channel states the two facts only it knows; the engine owns every boundary:

```text
契约   Scope.parentSession + Scope.branchHints     (SPEC §8's sanctioned extension mechanism)
引擎   PiSessionStore.openOrCreate(id, inherit?)   fork at the branch point + window mark
渠道   feishu: two facts, ~10 lines               telegram/slack: one wiring change away
```

**One-time by construction.** The store's create path forks the parent once; an existing session ignores the fields entirely. The session existing IS the record that the decision was taken — no marker to persist, no decision to retry per turn.

**Fork at the branch point.** `branchHints` carry the referent's id and its reply chain's ids, nearest first (the chain from #327 doubles as the locator). The most recent parent entry carrying a hint wins, extended to the end of its exchange — a mid-exchange fork inherits a question without its answer. The order matters for the own-card case: a thread rooted on the agent's answer has an *unfindable* root id (the card was sent after its turn), but the chain above it is exactly the ids that did pass through room turns. A miss falls back to the room's present, with a warn.

**Windowed by a mechanical mark, not a model call.** One `appendCompaction` with a plain string bounds what the model sees: the newest **50 exchanges** within a **~50K-token estimate**, images priced **flat** at ~1.6K (their base64 length measures storage, not context — priced by chars, one photo would evict the whole text window). Disk keeps the full copy: storage and context are different budgets, and the elided half stays inspectable.

## Fail-open at every edge, visibly

| Edge | Behavior |
|---|---|
| Parent missing / unreadable | empty session + warn — context is not the ask |
| Parent journal > 32 MiB | skipped (the fork reads everything into memory) + warn |
| Torn tail line (fork racing the parent's own append) | empty session + warn |
| Dangling tool call (parent forked mid-turn) | repaired by the existing crash-recovery path — the child's first turn never hands the provider an invalid transcript |

## Wire

The HTTP invoke body and `session-remote` carry the two fields end-to-end; the Scope drift guard is updated so the **next** field forces the same decision. Malformed values are a 400, never a silent drop.

## Docs — §8 corrects its own record

Rungs 3 and 4 were one mechanism with two parameters. The old "not planned" verdict rested on two assumptions, both now corrected in the text: a fork does not "pay on every turn" (one-time by construction), and "copying the whole room" conflated storage with context. The seed rung 3 wanted is exactly the window mark — generated from real session entries, so images and tool results come along for free.

**Flagged for the maintainer:** SPEC §8's extension table gains a `Session lineage` row. §8 is the extension mechanism itself ("extensions attach through additional scope fields"), so this documents a use of it rather than changing locked semantics — but SPEC edits deserve an explicit nod.

## Tests (1352 green: lint / typecheck / test)

Store-level: fork-once + bidirectional independence, existing-session short-circuit, missing parent, hint hit (exchange-complete) and miss, the 50-exchange window with its visible count, the token budget cutting deeper on heavy exchanges, flat image pricing (a 400K-char photo does not evict the window), dangling-tool-call repair, torn journal, oversize skip, in-memory parity, and end-to-end `scope.parentSession → invoke → store`. Channel-level: a thread turn's scope names its room and the hint order; a main-timeline turn carries no lineage.

Mutation-verified: dropping the window mark −2, flat pricing −1, post-fork repair −1, hint matching −1, the channel's parentSession −1.

### DIFF
--- a/docs/SPEC.md
+++ b/docs/SPEC.md
@@ -146,6 +146,7 @@ The core stays small. The following extensions attach through additional `scope`
 |---|---|
 | Identity / multi-tenancy | `scope.principal`, e.g. `{ type, issuer, subject }` |
 | Source / trace | `scope.source` and other identifying fields |
+| Session lineage | `scope.parentSession` (+ optional `scope.branchHints`) — read once, when the named session is first CREATED, to seed it from the session it branched off; ignored for existing sessions and by engines without the capability |
 | Execution constraints such as deadline / budget | Middleware; for example, `budget_exceeded` becomes a `failed` event produced by Middleware |
 | Mid-turn steering | **Extension, not part of the v0.1 core signature** — provided by the [session control plane](design/session-control.md): `dispatch(session, { type: "steer" \| "follow_up" \| "abort", … })` modulates the run an invoke is driving, beside the invoke rather than through its signature. (An earlier sketch — an optional third `input?: AsyncIterable<Prompt>` parameter — was rejected in favor of the control plane.) If the desired behavior is “discard the current turn and go another way”, cancel + a new `invoke` with the same `session` still works without any extension. |
 | Thinking / citations / artifact streaming | Add new non-terminal `AgentEvent` types |
--- a/docs/design/participant-model.md
+++ b/docs/design/participant-model.md
@@ -287,7 +287,9 @@ to the end of its exchange — a mid-exchange fork inherits a question without i
 falls back to the room's present, with a warn). What the model sees is then bounded by one
 mechanical compaction mark — the newest 50 exchanges within a ~50K-token estimate, images priced
 flat — which IS rung 3's seed, generated from real session entries instead of re-serialized prompt
-text, so images and tool results come along for free. Every edge (missing parent, oversize journal,
+text, so images and tool results come along for free. Both limits govern how far the window extends
+into older history; the newest exchange is a floor, kept whole even when it alone exceeds the
+budget — an inheritance that drops the exchange the thread branched off would be no inheritance. Every edge (missing parent, oversize journal,
 torn tail line, a dangling tool call from forking a mid-turn room) fails toward an empty session
 with a warn: context is not the ask, and a thread must not lose its first turn to it.
 
--- a/docs/design/participant-model.md
+++ b/docs/design/participant-model.md
@@ -235,12 +235,12 @@ commit on `completed`.
 
 A thread must start from something. Four rungs, increasing in cost:
 
-| Rung | Mechanism | Gives the thread |
-|---|---|---|
-| 1 | referent anchor, truncated | the followed-up message, cut at some display-sized bound |
-| **2** | **referent anchor, bounded by the platform** | **the followed-up message in full (`REFERENT_MAX_CODE_POINTS`)** |
-| 3 | seed injection | the room's last N exchanges, in the thread's first prompt |
-| 4 | session fork | the room's entire history, reasoning and tool results included |
+| Rung | Mechanism | Gives the thread | Status |
+|---|---|---|---|
+| 1 | referent anchor, truncated | the followed-up message, cut at some display-sized bound | rejected |
+| **2** | **referent anchor, bounded by the platform** | **the followed-up message in full (`REFERENT_MAX_CODE_POINTS`)** | implemented |
+| 3 | seed injection | the room's last N exchanges, in the thread's first prompt | subsumed by rung 4 |
+| **4** | **session inheritance (fork at the branch point)** | **the room's history up to where the thread branched — text, images, tool results — windowed** | **implemented** |
 
 Rung 1 fails the model's own main path: following up on the agent's answer, where the answer is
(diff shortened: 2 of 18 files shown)


Sukitly:
**Review response** — triage and what changed in a1432b0.

| Item | Decision | Resolution |
|---|---|---|
| [P1] Branch hints can't hit in the canonical flow | **accept** | The envelope never carried the message's own id — the reviewer is right that both hints missed and the tests papered over it with hand-written ids. The envelope meta now embeds `msg <id>` (documented as load-bearing), and a new test builds the parent text with the **real** `feishuEnvelope()` so channel format and engine search cannot drift apart silently. Old sessions without ids degrade to leaf-fork + warn, as before. |
| [P1] JSONL fork is not atomic | **accept** | Write-then-publish: drafts fork into `.inherit-tmp/<cwd>/` — invisible to `list({cwd})` and the cwd-less scan alike — get repaired and window-marked there, and enter the real directory in **one atomic rename**. Torn half-children, complete-but-unmarked children, and the catch-path double-file are all structurally impossible now: every failure leaves nothing or one complete file. Tests pin draft invisibility, single-file-after-failure, and single-file-after-success. |
| [P1] Opaque routed session mis-parsed as place key | **accept** — the sharpest catch | `parentPlaceKey` deleted (built on a false premise: the key isn't always `placeKey`'s output). Lineage is now derived at RECORD time from the message, where routed-ness is known (`routed === undefined && m.thread_id`), stored on the turn record, and only read back at dequeue. Test: a routed `tenant:user:alice` thread message produces no lineage. |
| [P2] Unbounded hint scan | **accept** | Entries pre-serialized once; hints capped at 16 × 128 chars with a warn — the guard lives where the cost does, covering every caller including the wire. |
| [P2] 50K window not a hard cap | **partially accept** | (b) accepted: a parent compaction's summary + retained tail now charge the budget as base cost (test: a 40K summary squeezes the window to one exchange). (a) is deliberate and now documented as such, in code and §8: the newest exchange is a **floor** — the limits govern extension; an inheritance that drops the branched-off exchange would be no inheritance. |

Mutation-verified: envelope id −3 tests, routed-check −1, hint length cap −1, compaction base −1 — the last two added this round after the first mutation pass exposed the original tests as non-biting.

kid7st (commented):
Thesis: introduces the mechanism behind participant-model.md §5 ("a thread starts from what the room knew") as three coordinated pieces — a `Scope` extension (`parentSession` + `branchHints`, SPEC §8), a store `openOrCreate(id, inherit?)` that forks the parent's active path up to a hint-located branch point (or its present) and marks a windowed compaction, and a channel wire that records `parentSession` at ingress only for default-derived sessions (routed opaque ids exempt, defending against `tenant:user:alice`-shaped ids being re-parsed as place keys). The load-bearing simplification: "session exists" IS the record that the decision was taken — one-time by construction, no marker to persist, no per-turn re-decision. Drafts publish via one atomic rename, so crash-mid-fork leaves only invisible garbage the next attempt discards. Fail-open at every edge (missing parent / oversize journal / torn tail / no hint match / dangling tool call) with operator warns.

Reviewed dimension by dimension: semantics (one-time enforced structurally by the check-existing gate; fork-at-branch-point extends forward to end-of-exchange; window keeps newest exchange as floor), concurrency (invoke lease acquired before the store, so concurrent inherit at the same id is impossible), reconciliation (post-fork `reconcileInterruptedToolCalls` handles parent-forked-mid-turn dangling tool_use), boundary validation (`http.ts` rejects malformed fields with 400, session-remote drift guard updated), routed-exemption tested with the exact `tenant:user:alice` injection vector, substring hint match shape-agnostic on purpose. Test coverage is exhaustive (19 store tests + 3 channel tests + real-feishu-envelope integration test), PR body cites mutation verification.

Context (not a blocking finding): `encodeSessionId` (`src/engines/pi/sessions.ts:490`, pre-existing on main, not touched by this PR) uses `charCodeAt(0).toString(16).padStart(2, "0")` on UTF-16 code units — variable-width, not injective (`"\u0100"` and `"\u0010" + "0"` both encode to `"%100"`). This PR now passes user-controlled `parentSession` through the same encoder, amplifying the blast radius from "two sessions collide" to "wrong-parent inheritance". Not blocking THIS PR — the encoder was broken before it landed here, and the fix belongs in the session-engine factory work planned on `feature/session-engine`. Surfacing so it is not lost.

No findings.

PR #330, opened 2026-08-07
feat(feishu): a thread's first turn reads the room's un-answered discussion

Sukitly:
Part 3 of 3 in the thread-context work (#327 reply chain → #328 session inheritance → **#330 room-buffer fold**), and the last piece of participant-model.md §5's "a thread starts from what the room knew".

## Problem

"What the room knew" has two halves. The fork (#328) carries the half that is in the room's **session** — but a room's session only advances when the agent is summoned, so discussion since its last answered turn sits in the context buffer, absorbed by nothing. Open a topic right after a burst of chatter and the inherited memory ends *before* the chatter began.

## Mechanism

The thread's first answered turn folds the room's bucket into its prompt: its own labelled block above the thread's usual fold, its attachments riding the buffered tier behind the thread's own. **Peeked, never committed** — the room's own next answered turn still folds and commits the same messages, so each place delivers the discussion to its own memory exactly once.

## Why the first turn only

A prompt is not an independent request: it lands in the session, so the thread's second turn already carries the first turn's fold in its context. Re-folding would put a second identical copy of the block — **and of its images** — inside one context window (measured: three turns, three copies of each, three re-fetches of the same resource). Several copies of one screenshot read as several postings. The gate is therefore part of the feature, not an optimisation.

## Why an evictable cache may gate it

The first-turn fact is `agentSpokeIn`, a new reader on the shared participants store, asked at ingress before this turn records its own participation. That store is a cache by contract — bounded at 1000, LRU-evicted, warn-only persistence, documented as deletable — and that is sound **here** for a reason worth stating as a rule, because it does not generalise:

| Where the copy lives | What an evicted record costs |
|---|---|
| A prompt (this change) | One extra fold of a block whose label — *discussion in the room this thread branched from, not yet answered there* — is true whenever it is written. The record is rewritten as soon as that turn answers. |
| A session (rejected) | A permanent claim that the room's **current** chatter is *what this thread branched from*, for a thread that branched long before. |

Repetition degrades an answer; a stale durable label misstates the past. So this rung is prompt-bound, and rung 4 — which is session-bound — is gated on the session's own existence instead.

Two designs were built and dropped on the way here: the same fold gated on the same cache but written into the newborn session's birth mark (the right-hand column above), and a `Scope.seedContext` extension that made the gate unforgeable by moving it into the engine — correct, but it spread one best-effort context feature across eleven files, a SPEC extension row, the HTTP boundary and the remote client, and could only carry attachments as decoded marker lines.

## Rejected: making the room absorb eagerly

Writing each un-summoned message into the room's session at ingress would delete this rung entirely — the fork would see everything. It fails on three counts, measured rather than assumed:

- durable session writes take the run lease ([session-control.md](docs/design/session-control.md) §9), so ingress would block behind a multi-minute turn;
- that blocking sits on the pre-ACK path, where the webhook must answer before the platform re-pushes;
- consecutive ambient user entries reach the provider unmerged — the direct Anthropic API now combines same-role turns, but Bedrock still rejects them (`roles must alternate`), and `deploy agentcore` is a shipped host.

A room's session advancing only when summoned is therefore a constraint, not an oversight. Recorded in §8 so the question does not get re-opened from scratch.

## Size

18 lines of code and 13 of comment, across the feishu channel and one reader on the participants store. No contract change: nothing added to `Scope`, the wire, or the HTTP boundary.

## Tests

Lifecycle: the first thread turn folds the chatter, the second does not, and the room's own next answered turn still folds **and** commits the same messages — proof nothing was stolen. Attachments: the room's image crosses as a real image on the first turn, and a later turn neither re-attaches nor re-fetches it.

Mutation-verified: removing the first-turn gate fails two tests; committing the room bucket from the thread fails one; dropping the room attachments fails one.

Independent of #335, which carries the session-store work this branch once accumulated.

### DIFF
--- a/docs/design/participant-model.md
+++ b/docs/design/participant-model.md
@@ -302,18 +302,28 @@ results is fork's exclusive capability — is the reason rung 3 never shipped as
 
 **Rungs 3 and 4 are both implemented, and they carry different halves.** The fork carries what the
 room's SESSION knew; a room's session only advances when the agent is summoned, so discussion since
-its last answered turn sits in the context buffer, absorbed by nothing. A thread turn folds that
-bucket into its prompt, read-only — the room's own next answered turn still commits it, so each place
-sees the discussion exactly once in its own memory, and the fold's attachments cross as real
-attachments on the buffered tier.
+its last answered turn sits in the context buffer, absorbed by nothing. The thread's FIRST turn folds
+that bucket into its prompt, read-only — the room's own next answered turn still commits it, so each
+place sees the discussion exactly once in its own memory, and the fold's attachments cross as real
+attachments on the buffered tier rather than as marker lines.
 
-Folded on EVERY thread turn, not just the first. "First" is the tighter fold, and it was built and
-reverted: every way for a channel to know it is a memory that can be lost (the participants store is
-a cache by contract — bounded, evictable, deletable), and a forgotten entry re-labels an old thread's
-turn as newborn, injecting the room's CURRENT chatter under "what this thread branched from".
-Unconditional folding needs no such memory: it repeats the block while the room stays un-answered,
-and stops by itself the moment the room absorbs its bucket. The cost is a repeated block in a few
-prompts, bounded by the buffer's own char budget.
+**Once, not per turn**, and this is not an optimisation — a prompt is not an independent request. It
+lands in the session, so the thread's second turn already has the first turn's fold in its context;
+re-folding puts a second identical copy of the block, and of its images, in ONE context window
+(measured: three turns, three copies of each). Several copies of one screenshot read as several
+postings. After the first turn the content is already present, so only newly arrived chatter would
+carry any information, and it reaches the thread the ordinary ways — someone says it there, or the
+room answers and the thread's own buffer carries on.
+
+The first-turn fact comes from the participants store, which is a cache by contract (bounded,
+evictable, deletable) — and that is acceptable HERE precisely because the fold is prompt-bound. An
+evicted record costs one extra fold of a block whose label ("discussion in the room this thread
+branched from, not yet answered there") is true whenever it is written, and the record is rewritten
+as soon as that turn answers. The same cache may not gate anything whose label would become a LIE
+after eviction: an earlier revision wrote this content into the newborn session's birth mark under
+"when this thread started", where a forgotten record would permanently attribute the room's CURRENT
+chatter to a thread that branched long before. That is why the room's discussion is folded into a
+prompt and not into a session.
 
 *Rejected: making the room absorb eagerly* — writing each un-summoned message into the room's session
 at ingress. It would delete this rung entirely (the fork would see everything), and it fails on three
--- a/docs/design/participant-model.md
+++ b/docs/design/participant-model.md
@@ -319,11 +319,13 @@ The first-turn fact comes from the participants store, which is a cache by contr
 evictable, deletable) — and that is acceptable HERE precisely because the fold is prompt-bound. An
 evicted record costs one extra fold of a block whose label ("discussion in the room this thread
 branched from, not yet answered there") is true whenever it is written, and the record is rewritten
-as soon as that turn answers. The same cache may not gate anything whose label would become a LIE
-after eviction: an earlier revision wrote this content into the newborn session's birth mark under
-"when this thread started", where a forgotten record would permanently attribute the room's CURRENT
-chatter to a thread that branched long before. That is why the room's discussion is folded into a
-prompt and not into a session.
+as soon as that turn answers.
+
+The converse is the rule worth keeping: **this cache may not gate a claim that outlives the turn.**
+Write the same content into the newborn session instead — as a birth mark reading "when this thread
+started" — and a forgotten record permanently attributes the room's CURRENT chatter to a thread that
+branched long before. Repetition degrades an answer; a stale durable label misstates the past. That
+is the whole reason this rung is prompt-bound and rung 4 is not.
 
 *Rejected: making the room absorb eagerly* — writing each un-summoned message into the room's session
 at ingress. It would delete this rung entirely (the fork would see everything), and it fails on three
--- a/docs/design/participant-model.md
+++ b/docs/design/participant-model.md
@@ -311,9 +311,10 @@ attachments on the buffered tier rather than as marker lines.
 lands in the session, so the thread's second turn already has the first turn's fold in its context;
 re-folding puts a second identical copy of the block, and of its images, in ONE context window
 (measured: three turns, three copies of each). Several copies of one screenshot read as several
-postings. After the first turn the content is already present, so only newly arrived chatter would
-carry any information, and it reaches the thread the ordinary ways — someone says it there, or the
-room answers and the thread's own buffer carries on.
+postings. After the first turn the content is already present, so only chatter arriving LATER could
+carry anything new — and that does not reach the thread at all. It is the same asymmetry stated at
+the end of this section: what a room says after a thread branches stays in the room, until someone
+repeats it there or an asker quotes the message and rung 2 loads it.
 
 The first-turn fact comes from the participants store, which is a cache by contract (bounded,
 evictable, deletable) — and that is acceptable HERE precisely because the fold is prompt-bound. An
--- a/docs/design/participant-model.md
+++ b/docs/design/participant-model.md
@@ -239,7 +239,7 @@ A thread must start from something. Four rungs, increasing in cost:
 |---|---|---|---|
 | 1 | referent anchor, truncated | the followed-up message, cut at some display-sized bound | rejected |
 | **2** | **referent anchor, bounded by the platform** | **the followed-up message in full (`REFERENT_MAX_CODE_POINTS`)** | implemented |
-| 3 | seed injection | the room's last N exchanges, in the thread's first prompt | subsumed by rung 4 |
+| **3** | **room-buffer fold, in the prompt** | **what the room heard but no session absorbed — text and attachments** | **implemented** |
 | **4** | **session inheritance (fork at the branch point)** | **the room's history up to where the thread branched — text, images, tool results — windowed** | **implemented** |
 
 Rung 1 fails the model's own main path: following up on the agent's answer, where the answer is
@@ -300,9 +300,34 @@ conflated storage with context (disk keeps the full inspectable copy; the mark g
 the model — different budgets). The one prediction it got right — preserving reasoning and tool
 results is fork's exclusive capability — is the reason rung 3 never shipped as prompt text.
 
+**Rungs 3 and 4 are both implemented, and they carry different halves.** The fork carries what the
+room's SESSION knew; a room's session only advances when the agent is summoned, so discussion since
+its last answered turn sits in the context buffer, absorbed by nothing. A thread turn folds that
+bucket into its prompt, read-only — the room's own next answered turn still commits it, so each place
+sees the discussion exactly once in its own memory, and the fold's attachments cross as real
+attachments on the buffered tier.
+
+Folded on EVERY thread turn, not just the first. "First" is the tighter fold, and it was built and
+reverted: every way for a channel to know it is a memory that can be lost (the participants store is
+a cache by contract — bounded, evictable, deletable), and a forgotten entry re-labels an old thread's
+turn as newborn, injecting the room's CURRENT chatter under "what this thread branched from".
+Unconditional folding needs no such memory: it repeats the block while the room stays un-answered,
+and stops by itself the moment the room absorbs its bucket. The cost is a repeated block in a few
+prompts, bounded by the buffer's own char budget.
+
+*Rejected: making the room absorb eagerly* — writing each un-summoned message into the room's session
+at ingress. It would delete this rung entirely (the fork would see everything), and it fails on three
+counts, all measured rather than assumed: durable session writes take the run lease
+([session-control.md](session-control.md) §9), so ingress would block behind a multi-minute turn; that
+blocking sits on the PRE-ACK path, where the webhook must answer before the platform re-pushes; and
+consecutive ambient user entries reach the provider unmerged — the direct Anthropic API now combines
+same-role turns, but Bedrock still rejects them (`roles must alternate`), and `deploy agentcore` is a
+shipped host. The room's session advancing only when summoned is therefore a constraint, not an
+oversight.
+
 Adoption is per channel, because only a channel knows its places' lineage: Feishu/Lark set the
-scope fields (threads inherit); Telegram and Slack do not yet (their threads still start empty —
-the fields are one wiring change away).
+scope fields and fold the room bucket (threads inherit); Telegram and Slack do not yet (their threads
+still start empty — the fields are one wiring change away).
 
 *Known asymmetry, accepted:* room → thread transfers once, when the thread starts, and nothing flows
 back (§7). A thread neither learns about later room activity nor reports its own.
--- a/src/channels/feishu/feishu.ts
+++ b/src/channels/feishu/feishu.ts
@@ -98,10 +98,12 @@ interface StoredFeishuTurn {
   /** The place this thread branched from (§5 lineage) — recorded at ingress, where the session's
    *  routed-ness is known; absent for main places, routed (opaque) sessions, and pre-upgrade records. */
   parentSession?: string;
-  /** The ROOM's buffer bucket, folded READ-ONLY into this thread turn (§5's other half: the room's
+  /** The ROOM's buffer bucket, folded READ-ONLY into this turn (§5's other half: the room's
    *  not-yet-answered discussion is part of "what the room knew", and the session fork cannot carry
-   *  what no session has absorbed). Recorded at ingress because only there is the SOURCE chat known
-   *  — a route may point the answer at a different one. Never committed from here. */
+   *  what no session has absorbed). Set only on a thread's FIRST turn, and recorded at ingress
+   *  because only there are both facts known — whether the agent has answered here before, and which
+   *  chat the message came FROM (a route may point the answer at a different one). Never committed
+   *  from here. */
   roomBufferKey?: string;
   images: { msg: string; key: string }[];
   files: { msg: string; key: string; name?: string }[];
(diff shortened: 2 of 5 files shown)


Sukitly:
**Review response** — the [P1] is accepted in full; fixed in ae2fb11.

The diagnosis was exact: `thread-participants.json` is a cache **by contract** — bounded at 1000 with LRU eviction, warn-only persistence, documented as deletable to re-bootstrap. Forgetting costs the summon rule one extra @mention (self-healing); it costs an exactly-once decision its truth — a forgotten entry would re-classify an old thread as newborn and inject the room's *current* chatter under a false "branched from" label.

**The fix binds the decision to the fact that cannot be forgotten** — whether this invoke creates the session, which only the engine knows:

```text
channel  peeks the room bucket on EVERY default-route thread turn (pure snapshot)
         → scope.seedContext
engine   session exists  → ignored (same as parentSession)
         session created → written into the newborn's birth mark (merged into the
                           inheritance window's compaction entry; seed-only mark when
                           nothing is elided; applied on every fallback path too — a
                           missing parent does not un-say what the room was discussing)
```

The field self-expires at the source: the room's next answered turn commits its bucket and the channel stops sending. The wrong binding's scaffolding is reverted whole (`agentSpokeIn`, `roomBufferKey`, the prompt-side fold block).

**One honest loss, stated in the docs:** seed text is session-bound, so room-chatter attachments arrive as decoded marker lines, not pixels — the price of a gate that cannot lie. An image the user actually points at still travels the reply chain (#327) and the fork (#328).

Tests: the lifecycle test now pins the drift-free shape end to end — seed present on turn 1 *and* turn 2 (the channel does not pretend to know which was first), room's own fold+commit untouched, field empty on turn 3; engine-side, the SECOND-SEED test pins that an existing session never absorbs one. 1363 green; mutations: fallback-seed −1, window-mark-seed −2.

src/channels/feishu/feishu.ts
```diff
@@ -466,6 +466,16 @@ function createFeishuRuntimeFactory(
...
         });
+        // The ROOM's un-answered discussion — context no session has absorbed yet. The channel cannot
+        // know whether THIS turn creates the thread's session (that fact lives with the engine), so
+        // the seed travels on every thread turn and the engine applies it exactly once, at creation —
+        // bound to the session's existence, never to a channel-side cache that may forget
+        // (thread-participants is a cache by contract: bounded, evictable, deletable). PEEK only: the
+        // room's own next answered turn commits its bucket, which empties this field by itself.
+        const seedContext =
+          rec.parentSession !== undefined
+            ? buffer.peek(feishuBufferPlaceKey({ chatId: rec.chatId })).text || undefined
```
kid7st: `seedContext` is re-read from the shared room buffer at dequeue instead of being frozen with the turn intent that was already persisted. A thread turn and a room turn are different sessions and run concurrently: if the room turn completes and commits the bucket first, and the thread's session is created after that, the fork's branch point still sits before the room turn — the un-answered discussion then disappears from both the seed and the inherited history. The reverse interleaving is the other half of the same defect: chatter that arrived after the thread was created gets labelled "when this thread started". The lifecycle test only exercises strictly sequential turns, so neither ordering is covered.

Direction: snapshot the room buffer text at ingress for every default-route thread turn and store it on the durable turn record; dequeue then only forwards that snapshot, and the engine still decides whether to apply it from whether this invoke creates the session. Add tests that interleave the room and thread queues, and one that replays a stored turn across a restart.

src/engines/pi/sessions.ts
```diff
@@ -350,25 +387,32 @@ async function createInheriting<M>(
   parentId: string,
   maxBytes: number,
 ): Promise<Session> {
+  // Every fallback still carries the seed: the pending discussion is real context whether or not
+  // the fork succeeded, and this is the one turn that can deliver it.
+  const createSeeded = async () => {
+    const child = await repo.create(id);
+    await appendSeedOnly(child, inherit.seedContext);
```
kid7st: The seeded fallback publishes a discoverable child via `repo.create(id)` and appends the birth mark afterwards. A failure in the append — or a crash between the two writes — leaves a session that exists but carries no seed, and the retry then takes the existing-session short-circuit and ignores the seed forever. That contradicts this PR's own "applied exactly once" claim, and it reintroduces the discoverable half-written child that #328 eliminated with the draft + atomic rename.

Direction: route the seeded fallback through the same write-then-publish path — fork/create into an undiscoverable draft, write the birth mark there, publish with one atomic rename. Add a fault-injection test proving that a failed append leaves no discoverable child and that the retry still applies the seed.

kid7st (commented):
This change adds `seedContext`: the channel supplies discussion that has not entered the parent session yet, and the pi session store writes it into the newborn session's birth mark on the create path only — completing the pending-hearing half of thread inheritance.

src/channels/feishu/feishu.ts
kid7st: Accepted, fixed in 920ce80. The snapshot moved to ingress, onto the durable turn record beside `parentSession`, and it is keyed by the SOURCE chat rather than the (route-redirectable) answer target, so the seed and the lineage always name the same room. Correction to the diagnosis: only one of the two orderings needed concurrency — a room turn answering before a thread is opened on an older message leaves the discussion below the fork AND already committed out of the bucket, which the ingress snapshot does not close. That half is now recorded as a known gap in §8 rather than fixed: closing it needs the channel to read the parent session, which is the layering the lineage fields exist to avoid. New test pins the queued-ask case (mutation-verified against the old dequeue read).

src/engines/pi/sessions.ts
kid7st: Accepted, fixed in 920ce80 — and the defect was worse than described. Besides the crash window, the empty-parent case returned its creation from INSIDE the try, so a throw after it landed in the catch, which created the same id a second time: two files under one session id, the catch-path double-file #328 removed. Now: one `createDraft` + atomic `publish` for the fallback too, every edge falling through to a single call site, and a failed birth mark degrading to the empty session that edge was already heading for instead of costing the turn. The fault-injection test you asked for is not there — making `appendCompaction` fail needs a seam in `InheritRepo` that exists only for the test; what is pinned instead is the empty-parent seed edge publishing exactly one discoverable session with nothing stranded in the draft realm.

src/engines/pi/sessions.ts
```diff
@@ -444,16 +523,20 @@ export function jsonlSessionStore(options: {
...
       const draftPath = ((await draft.getMetadata()) as unknown as { path: string }).path;
-      // The parent lives in the REAL cwd directory — its metadata is the authoritative way to name
-      // it (no re-deriving pi's cwd encoding). Same filesystem, so the rename is atomic.
-      const target = join(dirname(parentMeta.path), basename(draftPath));
+      // `<dir>/.inherit-tmp/<encodedCwd>/<file>` → `<dir>/<encodedCwd>/<file>`: the draft's own parent
+      // directory NAME is pi's cwd encoding, already computed — read it back rather than re-deriving
+      // it, and rather than borrowing the parent session's path (a parentless draft has none). Same
+      // filesystem, so the rename is atomic; the real directory may not exist yet when this store has
+      // never created a session for this cwd.
+      const target = join(options.dir, basename(dirname(draftPath)), basename(draftPath));
```
kid7st: Deriving the publish target from `options.dir` turns "absolute path" from a derived fact into an undeclared precondition. The previous `join(dirname(parentMeta.path), …)` consumed a path pi had already resolved, so the shape of `options.dir` never mattered. It does now, and the two resolution bases in this store disagree: `NodeExecutionEnv({ cwd })` resolves a relative path against `cwd`, while `mkdir`/`rename` resolve against `process.cwd()`. When they differ, `mkdir(recursive)` CREATES the wrong directory and the session is renamed into it — `list({ cwd })` then cannot find it, so "the session existing IS the record that the decision was taken" silently stops holding, with no error anywhere. In-tree callers all pass absolute paths today, so this is latent rather than live — but `jsonlSessionStore` is a published injection port and its documented signature (`docs/api-reference.md:409`) asks nothing of `dir`.

Direction: derive the target entirely from `draftPath`, which is already absolute — the draft root is `<dir>/.inherit-tmp/<encodedCwd>/<file>`, so three `dirname` hops give `<dir>` without touching `options.dir`. Alternatively normalize `dir` once at the `jsonlSessionStore` entry and state the requirement in its options doc.

src/engines/pi/sessions.ts
```diff
@@ -271,13 +285,38 @@ function locateBranchPoint(path: SessionTreeEntry[], hints: string[]): string |
...
 
+/** The seed as it may be WRITTEN: bounded, with empty normalized to "no seed" so every downstream
+ *  `if (seed)` reads one thing. Truncation warns — a silently halved discussion misreports the room. */
+function usableSeed(seed: string | undefined): string | undefined {
+  if (!seed) return undefined;
+  if (seed.length <= MAX_SEED_CONTEXT_CHARS) return seed;
+  log.warn(
+    `[fastagent] seedContext is ${seed.length} chars (limit ${MAX_SEED_CONTEXT_CHARS}) — truncating; it is a digest of pending discussion, not a payload`,
+  );
+  return seed.slice(0, MAX_SEED_CONTEXT_CHARS);
```
kid7st: `seed.slice(0, MAX_SEED_CONTEXT_CHARS)` cuts UTF-16 code units, so it can split a surrogate pair and leave a lone surrogate. This text goes into the birth mark, which means it is in EVERY later context of that session — a provider that rejects the resulting invalid UTF-8 poisons the whole session, not one turn. The repo already owns the safe primitive for exactly this (`truncateCodePointPrefix` in `src/channels/text.ts`, whose header calls it Unicode-safe code-point slicing), but it sits in the channel layer and importing it from the engine would invert the dependency. Note who can reach this: a channel seed is bounded by the buffer's 4000-char budget and never trips the cap, so the only callers that can are the wire and direct embedders — the least trusted side.

Direction: slice by code point (`Array.from`) — either lift the shared helper to a layer both can consume, or keep a small engine-local one. Keeping the helper's ellipsis marker would also make the truncation visible in the text rather than silent.

test/feishu.test.ts
```diff
@@ -1768,6 +1768,176 @@ describe("turn flow", () => {
...
+        messageEvent({
+          id: "om_room_ask",
+          chatType: "group",
+          content: JSON.stringify({ text: "@_bot 总结一下" }),
+          mentions: mention,
+        }),
+      ),
+    );
+    await flush();
+    release();
```
kid7st: The scenario this test claims to pin requires a precondition it never asserts: the room turn must have completed AND committed its bucket before the gated thread turn is dequeued. That is currently achieved only by `await flush()` polling the fetch mock until quiet. If the room turn ever gains a non-fetch wait, `flush()` returns early and the test passes under the OLD dequeue-time read too — it degrades into a tautology that is green forever instead of failing.

Direction: assert the precondition before `release()` — that the room ask is present in `seenTurns`, and that the room's bucket in `buffers.json` has been emptied — so a broken precondition fails the test instead of silently removing its protection.

kid7st (commented):
Thesis: `seedContext` carries the parent place's not-yet-absorbed discussion across the lineage seam — the channel snapshots it at ingress (where "when this thread started" is true), the wire forwards it, and the pi store applies it exactly once, into the newborn session's birth mark. This round moved the snapshot to ingress, routed the seeded fallback through the fork's draft-then-publish path, collapsed every inheritance edge onto one creation site, and bounded the seed where its cost is.

The three findings below are all on the latest commit.

src/engines/pi/sessions.ts
kid7st: Accepted, fixed in 407312d. `dir` is now resolved once at the store entry by pi's own rule (`resolve(cwd, dir)` — what `NodeExecutionEnv.absolutePath` does), and both repos plus the publish target read that single base; the raw option no longer reaches any path join. New test builds a store with a relative dir and an explicit cwd and asserts the inherited child is discoverable and lands under the workspace — it fails when the publish target goes back to the raw option. Incidentally confirmed while mutation-testing: the broken version wrote a stray session tree into the repo root, exactly as described.

src/engines/pi/sessions.ts
kid7st: Fixed in 407312d, with the severity corrected: a halved pair does NOT reach a provider as invalid UTF-8 — JSON round-trips a lone surrogate and UTF-8 encoding substitutes U+FFFD — so the claim about a poisoned session in the original comment was wrong. What stands is that the cut disagreed with how this repo cuts user text everywhere else (channels/text.ts), and that the debris would sit in the birth mark rather than in one turn. Now sliced by code point, with the constant renamed to what it measures; the engine keeps its own four lines rather than importing the channel kit backwards. Worth noting for the test: the first version asserted through `JSON.stringify`, which ESCAPES a lone surrogate — it passed under the broken implementation. It now reads the mark's raw summary and bites.

test/feishu.test.ts
kid7st: Accepted, fixed in 407312d. The precondition is now waited on and asserted — the room turn has run and its bucket no longer holds the chatter — plus an assertion that the gated turn is still queued at that point. `flush()` is now an accelerator rather than the thing the scenario depends on.

src/channels/feishu/feishu.ts
```diff
@@ -461,8 +467,23 @@ function createFeishuRuntimeFactory(
...
+        // discussion exactly once in its own memory.
+        const room = rec.roomBufferKey !== undefined ? buffer.peek(rec.roomBufferKey) : undefined;
+        const roomBlock = room?.text
+          ? `[recent discussion in the room this thread branched from — not yet answered there:\n${room.text}\n]\n\n`
+          : "";
+        const threadBlock = recent ? `[recent group discussion:\n${recent}\n]\n\n` : "";
+        const prompt = `${roomBlock}${threadBlock}${rec.baseText}`;
+        // Room entries FIRST: the collector keeps the tail under its cap, so the thread's own
+        // discussion outranks the parent room's when both compete for attachment slots.
+        const buffered = collectFeishuBufferedAttachments([...(room?.consumed ?? []), ...consumed], {
```
kid7st, line 477: The "repeat every turn" trade was argued for TEXT, and it does not hold for attachments — but attachments ride the same unconditional path.

Measured, with three thread turns while the room stays un-summoned: `turns=3 downloads=3 promptsWithImage=3`. The same room image is re-fetched from the platform three times and re-sent as vision content three times.

For text the cost is genuinely bounded: the buffer's own 4000-char budget caps what one fold can carry, so repetition is cheap and self-limiting. Attachments are outside that budget, they are the expensive kind of context, and the multiplier — how many turns the thread runs before the room is next summoned — is not something the thread controls or bounds. `participant-model.md:315` and the commit message both describe the cost as "a repeated block in a few prompts, bounded by the buffer's own char budget", which is true of the text half only; the code and its own description disagree.

Two directions, and this needs a deliberate choice rather than a default:

- **Fold room TEXT only, let room attachments stay behind.** Capability then equals the rejected `seedContext` design (marker lines, not pixels) while keeping the whole 33-line, no-protocol win. A user who wants the picture can still quote it — rung 2 (#327) loads a referent's images.
- **Keep attachments and state the cost accurately.** Document that a room attachment re-crosses on every thread turn until the room is summoned, and why that is acceptable (at most 3 slots per turn, and the thread's own buffered attachments take them first).

src/channels/feishu/feishu.ts
```diff
@@ -660,6 +681,13 @@ function createFeishuRuntimeFactory(
       // time, where routed-ness is still known; the dequeue path only reads it back.
       const parentSession =
         routed === undefined && m.thread_id !== undefined ? placeKey(kind, { chat_id: m.chat_id }) : undefined;
+      // The room's un-summoned chatter is the half of "what the room knew" that no session holds yet.
+      // Recorded here because the SOURCE chat is only known here (a route may answer elsewhere), and
+      // derived FROM the lineage rather than beside it: a place with no parent has no room to read,
+      // and an opaque routed session must get neither. (Untested on purpose — the two cannot diverge,
+      // and a routed deployment never fills a room bucket anyway: route-null messages are dropped,
+      // not buffered.)
+      const roomBufferKey = parentSession !== undefined ? feishuBufferPlaceKey({ chatId: m.chat_id }) : undefined;
```
kid7st: The choice of `m.chat_id` (the room the message came FROM) over `chatId` (`r.chatId ?? m.chat_id`, where the answer goes) is deliberate and explained in the comment, but nothing pins it: changing it to `chatId` leaves all 75 tests green, because every existing case has the two equal.

This is exactly the kind of decision that needs a test rather than a comment — non-obvious, invisible in the common case, and wrong in a specific direction: when a route redirects the answer to another chat, reading the target's bucket folds a DIFFERENT room's discussion into the thread, under a label that says it came from the room this thread branched from.

Direction: one case with `route: () => ({ chatId: "oc_other" })`, un-answered chatter in the source room and none in the target, asserting the fold carries the source room's text.

kid7st (commented):
Thesis: rung 3 of the inheritance ladder, as a channel-local read — a thread turn folds the room's un-absorbed context buffer into its prompt, peeked and never committed, so the room's own next answer still delivers the same messages to its own memory. The load-bearing decision is the absence of a first-turn gate: unconditional folding needs no channel-side memory to be correct, and self-expires when the room absorbs its bucket.

Two findings, both on the consequences of that decision rather than on the decision itself.

src/channels/feishu/feishu.ts
kid7st: Accepted, and the measurement went further than the finding. A prompt is not an independent request — it lands in the session — so re-folding does not merely repeat a cost: it puts a second identical copy of the block AND of its images inside ONE context window. Measured at three turns: three copies of each, three re-fetches. Several copies of one screenshot read as several postings.

That makes the first-turn gate part of the feature rather than an optimisation, so it is restored (agentSpokeIn). The cache-forgets objection that killed this gate earlier does not transfer once the copy is prompt-bound: eviction costs ONE extra fold of a block whose label is true whenever it is written, whereas the same cache gating a session-bound birth mark would permanently attribute the room current chatter to a thread that branched long before — a lie rather than a repetition.

Fixed in 709db83. The attachment test now pins that a later turn neither re-attaches nor re-fetches, and removing the gate fails two tests.

src/channels/feishu/feishu.ts
kid7st: Still open, deliberately. With the first-turn gate restored the source-chat choice sits behind two conditions rather than one, but a route redirecting the answer elsewhere is still the only case where source and target diverge, and it is still green under a mutation to the target chat.

Left as is because the fold is prompt-bound and best-effort: reading the wrong bucket costs one wrong block on one turn, not a durable claim. Say so and I will add the route: () => ({ chatId }) case.

kid7st: Closing this as won't-fix rather than leaving it open: the decision is made, so the thread should not read as pending work.

The fold is prompt-bound and best-effort, so reading the target bucket instead of the source one costs a single wrong block on a single turn — no durable claim, and it self-corrects on the next turn. The choice is also visible in one expression next to the lineage it derives from. A test would pin it, but the case it guards (a route redirecting the answer to another chat, in a room that has un-answered chatter) is worth less than the test it would cost.

kid7st (commented):
Thesis: rung 3 as a channel-local, prompt-bound read — the thread's first answered turn folds the room's un-absorbed context buffer into its prompt, peeked and never committed, so each place delivers the discussion to its own memory exactly once. The first-turn fact comes from the participants cache, which is legitimate here precisely because the copy is prompt-bound: eviction costs one extra fold, never a stale durable claim.

One finding, in the design doc rather than the code.

Two things I checked and am NOT filing, recorded so the next reviewer does not re-derive them: the first-turn gate never closes under a redirect route (participation is recorded only when the answer lands in the same chat), and it never closes for p2p threads (participation is not recorded there at all). Both are inert — a custom route disables buffering entirely (`route === undefined && isHumanGroup` guards the only `buffer.push`), and a p2p main place never buffers, so in both cases the room bucket the gate protects is always empty.

docs/design/participant-model.md
```diff
@@ -300,9 +300,46 @@ conflated storage with context (disk keeps the full inspectable copy; the mark g
...
+**Rungs 3 and 4 are both implemented, and they carry different halves.** The fork carries what the
+room's SESSION knew; a room's session only advances when the agent is summoned, so discussion since
+its last answered turn sits in the context buffer, absorbed by nothing. The thread's FIRST turn folds
+that bucket into its prompt, read-only — the room's own next answered turn still commits it, so each
+place sees the discussion exactly once in its own memory, and the fold's attachments cross as real
+attachments on the buffered tier rather than as marker lines.
+
+**Once, not per turn**, and this is not an optimisation — a prompt is not an independent request. It
+lands in the session, so the thread's second turn already has the first turn's fold in its context;
+re-folding puts a second identical copy of the block, and of its images, in ONE context window
```
kid7st, line 312: This sentence offers two routes for later room chatter to reach the thread, and the second one does not exist:

> …it reaches the thread the ordinary ways — someone says it there, **or the room answers and the thread's own buffer carries on**.

The room answering commits the ROOM's bucket; the thread's own buffer holds un-summoned messages posted *in the thread*, which has nothing to do with room chatter. After the first fold, nothing carries later room discussion into the thread — which is exactly what this same section states a few lines further down: *"Known asymmetry, accepted: room → thread transfers once, when the thread starts, and nothing flows back. A thread neither learns about later room activity nor reports its own."*

So one document asserts a mechanism and then denies it within the same section. A reader who trusts the first version ships expecting the thread to stay current with its room.

Direction: drop the second clause and state the negative outright — after the first turn, later room discussion does not reach the thread; someone repeats it there, or the asker quotes the message and rung 2 loads it. Then point at the Known asymmetry paragraph instead of implying a path around it.

kid7st: Anchor correction: the clause is on lines 315-316, not 312 — same paragraph, four lines down. The quoted text in the finding is exact.

kid7st: Accepted, fixed. The clause is gone; the paragraph now states the negative and points at the Known asymmetry it was contradicting. Root cause worth noting: the sentence was true-ish while the fold ran on every turn, and was not revised when the fold became first-turn-only — code changed, its companion text did not.

PR #331, opened 2026-08-12
feat(config): make coding tools configurable

Sukitly:
Add `codingTools` so a directory agent can select which of pi's built-in `read`, `bash`, `edit` and
`write` it mounts: all-on (`true`/unset), all-off (`false`/`[]`), or a least-privilege subset such as
`["read"]`. One resolved posture applies across `dev`, `start`, `invoke`, `chat`, `tool` and `info`.

Original work by @Shuqian; rebased onto the single-engine `main` and continued here.

## Why a subset, not a switch

Capabilities are not interchangeable. Skills and downloaded non-image channel attachments are
file-backed and need `read`; they do not need a shell or mutation. Without the subset, a
public-facing agent has to choose between a full coding surface and none — and losing `read` quietly
breaks its own skills.

So the failures are made visible rather than left to surprise: model-visible skills that have no
reader are reported at assembly, and Telegram, Slack and Feishu refuse to hand the model a local
attachment path it cannot read instead of passing one that will fail.

## It is a boundary, not a default

`noTools: "builtin"` keeps pi's own copies out of the *active* set, but they remained in the session
registry — so anything that activates by name (a `chat` command, the control plane, the
deferred-tool loader) could put one back. A public-facing agent on `codingTools: false` was one
`/tools` away from a shell.

Disabled built-ins are now passed as pi's `excludeTools`, which filters inside
`_refreshToolRegistry`, so they are never mountable. The test asks for `bash` explicitly and does not
get it.

Two consequences that needed their own handling:

- **Disabling a name frees it.** The denylist matches by name, so it would have deleted an *authored*
  tool that reused one of the four — the opposite of what the docs promise. The exclusion now
  subtracts whatever the mounted set already provides.
- **An explicit `tools` list states its own capabilities.** L2 callers passing `piDefaultTools()`
  were treated as having no coding tools, which gave them a capability-neutral identity and a false
  "skills have no reader" warning. Capabilities are read from the built-in names present in the list.

## What the rebase simplified

The original narrowed chat a second time, by name, through pi's `tools` allowlist:

```ts
const defaultNames = codingToolNames;
const customTools = tools.filter((t) => !defaultNames.includes(t.name));
```

That allowlist is gone (#345 removed it — it froze the tool set against extensions registering from
a handler), and `resolveTools()` already applies `codingTools`. Both assemblies now take an
already-narrowed list from `resolveAgentAssembly`, so the posture lands in one place instead of two
that could disagree.

## Validation

- [x] `npm run lint`, `npm run typecheck`, `npm test` (108 files, 1379 tests)
- [x] Rebased onto the single-engine `main`; `invoke.ts`/`sessions.ts`/`harness.ts` no longer exist
- [x] `rv.sh local` — **No findings.** Each behavioural fix falsified by reverting it and confirming
      the test goes red

## Checklist

- [x] Docs were updated (`docs/configuration.md`, `docs/api-reference.md`, `docs/cli.md`, `docs/design/core.md`)

### DIFF
--- a/src/channels/feishu/invoke-turn.ts
+++ b/src/channels/feishu/invoke-turn.ts
@@ -12,7 +12,7 @@
  * ancestors, and degrade per attachment: one expired background file must not block the current ask
  * or hide its still-readable siblings.
  */
-import type { Agent, AgentEvent, ImageRef, Scope } from "../../agent.ts";
+import { type Agent, type AgentEvent, ATTACHMENT_UNSUPPORTED_CODE, type ImageRef, type Scope } from "../../agent.ts";
 import { log } from "../../log.ts";
 import {
   type BusyRetry,
@@ -375,7 +375,7 @@ export async function* invokeFeishuTurn(
       type: "failed",
       details: unavailable ? e.message : `could not load attachment: ${String(e)}`,
       retryable: !unavailable,
-      ...(unavailable ? { code: "attachment_unsupported" } : {}),
+      ...(unavailable ? { code: ATTACHMENT_UNSUPPORTED_CODE } : {}),
     };
     return;
   }
--- a/src/channels/feishu/invoke-turn.ts
+++ b/src/channels/feishu/invoke-turn.ts
@@ -227,6 +227,11 @@ async function resolveTurnInputs(t: FeishuTurnTransport, attachments: FeishuTurn
   const { canReadLocalFiles } = t;
   const images = [...attachments.primary.images];
   const files = [...attachments.primary.files];
+  // The ASK's own files, before a referent's join them. Fail-fast applies to these only: the user
+  // pointed at them, so silently dropping them would answer a question they did not ask.
+  if (!canReadLocalFiles && files.length > 0) throw new LocalFileAccessUnavailable(t.label);
+  // A referent's unreadable files degrade instead (see below); counted for the note.
+  let unreadableReferentFiles = 0;
   let referentBlock = "";
   let chain: ReplyChain = { block: "", images: [], files: [], ids: [] };
   if (attachments.primary.parentId !== undefined) {
@@ -257,8 +262,14 @@ async function resolveTurnInputs(t: FeishuTurnTransport, attachments: FeishuTurn
         mentions: parent.mentions as FeishuMention[] | undefined,
       });
       // The referent's own resources join the turn as primary inputs, carried by the PARENT message id.
+      // Images always: vision needs no reader. Files only if there IS one — a quoted file must not
+      // cost the user their turn, which is what this function's own policy says two paragraphs up.
       for (const key of parsed.imageKeys) images.push({ msg: parentId, key });
-      for (const ref of parsed.fileRefs) files.push({ msg: parentId, key: ref.key, name: ref.name });
+      if (canReadLocalFiles) {
+        for (const ref of parsed.fileRefs) files.push({ msg: parentId, key: ref.key, name: ref.name });
+      } else {
+        unreadableReferentFiles += parsed.fileRefs.length;
+      }
       const from = fetchedSenderLabel(parent.sender, t.appId);
       referentBlock = `\n\n[replied-to message (msg ${parentId}${from ? `, from ${from}` : ""}): ${truncateCodePointPrefix(parsed.text, REFERENT_MAX_CODE_POINTS) || "(empty)"}]`;
       // The referent's own parent starts the chain walk; the referent id seeds the cycle guard.
@@ -267,8 +278,6 @@ async function resolveTurnInputs(t: FeishuTurnTransport, attachments: FeishuTurn
     }
   }
 
-  // Primary first and fail-fast: these are resources the current user explicitly pointed at.
-  if (!canReadLocalFiles && files.length > 0) throw new LocalFileAccessUnavailable();
   const imageRefs: ImageRef[] = [];
   for (const ref of images) imageRefs.push(await t.api.fetchImage(ref.msg, ref.key));
   const downloaded: DownloadedFile[] = [];
@@ -313,7 +322,9 @@ async function resolveTurnInputs(t: FeishuTurnTransport, attachments: FeishuTurn
       log.warn(`${t.label} could not load an earlier (buffered) image: ${String(result.reason)}`);
     }
   }
-  const unreadable = canReadLocalFiles ? 0 : bufferedFiles.length;
+  // Everything skipped for want of a reader: this turn's buffered files, plus any the referent
+  // carried. One number, one note — the model is told what it did not get.
+  const unreadable = canReadLocalFiles ? 0 : bufferedFiles.length + unreadableReferentFiles;
   const fileResults = canReadLocalFiles
     ? await Promise.allSettled(
         bufferedFiles.map(async (ref) => ({
--- a/src/channels/feishu/invoke-turn.ts
+++ b/src/channels/feishu/invoke-turn.ts
@@ -56,8 +56,8 @@ export interface FeishuTurnTransport {
    *  (participant-model.md §5). The engine reads it once, at session creation; every later turn
    *  carries it inertly. */
   parentSession?: string;
-  /** Undefined preserves the legacy embedder posture; standard serving supplies an explicit fact. */
-  canReadLocalFiles?: boolean;
+  /** Whether the agent has a reader for absolute local paths; see ChannelContext. */
+  canReadLocalFiles: boolean;
 }
 
 /** An attachment reference: the resource key inside its CARRYING message (the resource API addresses
@@ -224,7 +224,7 @@ async function walkReplyChain(
  * degrade independently.
  */
 async function resolveTurnInputs(t: FeishuTurnTransport, attachments: FeishuTurnAttachments): Promise<ResolvedInputs> {
-  const canReadLocalFiles = t.canReadLocalFiles !== false;
+  const { canReadLocalFiles } = t;
   const images = [...attachments.primary.images];
   const files = [...attachments.primary.files];
   let referentBlock = "";
--- a/src/channels/feishu/invoke-turn.ts
+++ b/src/channels/feishu/invoke-turn.ts
@@ -17,11 +17,13 @@ import { log } from "../../log.ts";
 import {
   type BusyRetry,
   DEFAULT_BUSY_RETRY,
+  LocalFileAccessUnavailable,
   attachedFilesManifest,
   attributedFileName,
   backgroundImagesManifest,
   missingAttachmentsNote,
   streamTurnWithBusyRetry,
+  unreadableAttachmentsNote,
 } from "../invoke-turn-kit.ts";
 import { BUFFER_ATTACH_MAX } from "../context-buffer.ts";
 import type { FeishuBufferedRef } from "./context-buffer.ts";
@@ -54,6 +56,8 @@ export interface FeishuTurnTransport {
    *  (participant-model.md §5). The engine reads it once, at session creation; every later turn
    *  carries it inertly. */
   parentSession?: string;
+  /** Undefined preserves the legacy embedder posture; standard serving supplies an explicit fact. */
+  canReadLocalFiles?: boolean;
 }
 
 /** An attachment reference: the resource key inside its CARRYING message (the resource API addresses
@@ -220,6 +224,7 @@ async function walkReplyChain(
  * degrade independently.
  */
 async function resolveTurnInputs(t: FeishuTurnTransport, attachments: FeishuTurnAttachments): Promise<ResolvedInputs> {
+  const canReadLocalFiles = t.canReadLocalFiles !== false;
   const images = [...attachments.primary.images];
   const files = [...attachments.primary.files];
   let referentBlock = "";
@@ -263,6 +268,7 @@ async function resolveTurnInputs(t: FeishuTurnTransport, attachments: FeishuTurn
   }
 
   // Primary first and fail-fast: these are resources the current user explicitly pointed at.
+  if (!canReadLocalFiles && files.length > 0) throw new LocalFileAccessUnavailable();
   const imageRefs: ImageRef[] = [];
   for (const ref of images) imageRefs.push(await t.api.fetchImage(ref.msg, ref.key));
   const downloaded: DownloadedFile[] = [];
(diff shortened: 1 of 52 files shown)


kid7st (commented):
Thesis: this PR introduces `codingTools` — a config-level selection of pi's four machine-reaching built-ins — and propagates one resolved fact (`codingToolNames`, and its `canReadLocalFiles` projection) through prompt identity, skill findings, report lines and every chat channel's attachment path.

**`docs/cli.md` was not updated for the `diagnostics` change.** `info --json` now emits `AssemblyFinding` inside `diagnostics`, but the doc still describes that field as "skills and diagnostics". See the inline comment on `src/cli/commands/info.ts`.

**The three new channel tests do not cover the default error surface.** `test/telegram.test.ts`, `test/slack.test.ts` and `test/feishu.test.ts` each mount with a custom `onError` that leaks `failed.details`, so they assert on wording no default deployment ever shows. With the shipped default the user sees `"⚠️ Sorry, something went wrong…"` and the operator sees nothing. See the inline comment on `src/channels/invoke-turn-kit.ts`.

**`canReadLocalFiles` is derived ad hoc at six call sites.** `codingToolNames.includes("read")` is recomputed in `dev.ts`, `start.ts` (×3), `info.ts`, `open.ts`, `session-builder.ts` and `create.ts`. The concept exists as a named field on `ChannelContext` but not on `AgentAssembly`, which is where it is produced — put it there once and let consumers read it.

src/channels/feishu/invoke-turn.ts
```diff
@@ -143,6 +148,7 @@ async function resolveTurnInputs(t: FeishuTurnTransport, attachments: FeishuTurn
   }
 
   // Primary first and fail-fast: these are resources the current user explicitly pointed at.
+  if (!canReadLocalFiles && files.length > 0) throw new LocalFileAccessUnavailable();
```
kid7st: The referent's files are treated as user-attached primaries, so a thread reply hard-fails.

This function's own documented policy is explicit: a replied-to message is CONTEXT, "losing it must not cost the user their answer — every first message of a thread carries one, so a hard failure here would turn an ordinary platform edge into a lost turn." This guard runs *after* referent resolution has pushed `parsed.fileRefs` into `files`, so an agent without `read` loses the entire turn whenever the quoted message happens to be a file — and the message tells the user to "send the content as text" for a file they never sent. The new test `rejects a primary non-image referent without a reader…` encodes exactly that inverted semantics.

Guard `attachments.primary.files` only; referent files should degrade like every other unreadable referent (skip + note in the referent marker / `unreadableAttachmentsNote`).

src/channels/invoke-turn-kit.ts
```diff
@@ -69,6 +69,14 @@ export async function* streamTurnWithBusyRetry(
   }
 }
 
+/** A primary local-file attachment cannot be represented faithfully for an agent without a reader. */
+export class LocalFileAccessUnavailable extends Error {
```
kid7st, line 83: The actionable message is written for an operator and delivered to nobody.

With the default `onError`, `retryable: false` renders `"⚠️ Sorry, something went wrong. Try rephrasing, or check I have access to what you need."` — "mount the read coding tool" never reaches the end user. Nothing calls `log.warn` either, so the operator never sees it. A misconfigured agent silently answers "something went wrong" to every attachment. The three new tests all pass a custom `onError` that leaks `details`, so none covers the shipped path.

Log the condition operator-side at the throw site, and decide the user-facing wording deliberately: this is a configuration limitation, not an unknown failure, and `defaultErrorMessage` has no branch for it.

src/channels/telegram/invoke-turn.ts
```diff
@@ -120,7 +130,13 @@ export async function* invokeTurn(
   try {
     resolved = await resolveTurnAttachments(transport, attachments);
   } catch (e) {
-    yield { type: "failed", details: `could not load attachment: ${String(e)}`, retryable: true };
+    const unavailable = e instanceof LocalFileAccessUnavailable;
+    yield {
+      type: "failed",
+      details: unavailable ? e.message : `could not load attachment: ${String(e)}`,
+      retryable: !unavailable,
+      ...(unavailable ? { code: "attachment_unsupported" } : {}),
```
kid7st: `"attachment_unsupported"` is an undeclared code literal, copied three times (here, `slack/invoke-turn.ts:126`, `feishu/invoke-turn.ts:239`).

`agent.ts` states the convention and its reason: `SESSION_BUSY_CODE` / `ABORTED_CODE` exist as exported constants precisely so "a consumer that must branch on it should not string-match." This code has no constant, no doc, and no consumer — the channel emits it and the same channel consumes the event.

Either export it from `agent.ts` alongside the others with the consumer named, or drop it and let `retryable: false` be the whole signal.

src/channels/telegram/invoke-turn.ts
```diff
@@ -58,8 +62,10 @@ interface ResolvedAttachments {
 async function resolveTurnAttachments(t: TurnTransport, attachments: TurnAttachments): Promise<ResolvedAttachments> {
   const { api, botToken, chatId, filesDir } = t;
   const { primary, buffered } = attachments;
+  const canReadLocalFiles = t.canReadLocalFiles !== false;
+  if (!canReadLocalFiles && (primary.fileIds?.length ?? 0) > 0) throw new LocalFileAccessUnavailable();
   const images = await resolveImages(api, botToken, primary.imageFileIds);
-  const files = await resolveFiles(api, botToken, primary.fileIds, chatId, filesDir);
+  const files = canReadLocalFiles ? await resolveFiles(api, botToken, primary.fileIds, chatId, filesDir) : undefined;
```
kid7st: Dead ternary: line 66 already threw when `primary.fileIds` is non-empty, so the `: undefined` branch is unreachable, and `resolveFiles` over an empty list returns nothing anyway. Two readings of one rule two lines apart — drop the ternary.

src/engines/pi/create.ts
```diff
@@ -415,14 +454,16 @@ export async function createPiAgentFromDefinition(
   // Boot-time load: fail-visibly at startup on a broken directory, and give callers the snapshot to
   // report (skills/diagnostics/collisions). Serving does NOT close over it — see `live` below.
   const definition = await loadAgentDefinition(dir, { cwd: env.cwd, env });
+  // Deferred tools need their loader on every rung (idempotent — the workspace opener already applied
+  // it; a caller's own search_tools wins).
+  const tools = withSearchTool(options.tools ?? piDefaultTools());
+  const codingToolNames = options.codingToolNames ?? (options.tools === undefined ? [...CODING_TOOL_NAMES] : []);
```
kid7st: `codingToolNames` is inferred from whether `options.tools` was passed, not from what is mounted.

`options.codingToolNames ?? (options.tools === undefined ? [...CODING_TOOL_NAMES] : [])` records an L1/L2 caller doing `createPiAgentFromDefinition(dir, { tools: [...piDefaultTools(), myTool] })` as having zero coding tools: it gets the degraded "AI assistant" identity while mounting all four, plus a false `skills_require_file_reader` warning naming every skill.

Derive from the mounted set (`CODING_TOOL_NAMES.filter(n => tools.some(t => t.name === n))`) and keep `options.codingToolNames` only as the directory-opener's override.

src/engines/pi/create.ts
```diff
@@ -161,10 +191,16 @@ export function piBasePrompt(options: { tools?: MountedTool[]; persona?: string
   const toolsList =
     tools.length > 0 ? tools.map((t) => `- ${t.name}: ${(t.description ?? "").split("\n")[0]}`).join("\n") : "(none)";
   // Segment ① identity: an authored persona (persona.md) replaces the default engine identity line
-  // (core.md §11), keeping the tools list + guidelines below.
+  // (core.md §11), keeping the tools list + guidelines below. Preserve pi's coding identity only for
+  // the full coding surface; a partial/empty surface must not claim machine capabilities it lacks.
+  const inferredCodingNames = CODING_TOOL_NAMES.filter((name) => mounted.some((tool) => tool.name === name));
```
kid7st: `inferredCodingNames` is dead in every production path: both callers (`live()` below and `session-builder.ts`) always pass an explicit `codingToolNames`, so this branch only runs in `test/definition.test.ts`. It answers the same question as the assembly-site inference and answers it correctly — fixing that call site by inferring from the mounted tools makes this the single implementation instead of an unused sibling.

src/engines/pi/report.ts
```diff
@@ -35,11 +61,12 @@ const lastFindings = new Map<string, string>();
  * `dir` must be the RESOLVED definition root (`LoadedDefinition.dir`), or two spellings of one path
  * become two memos and warn twice.
  */
-export function reportFindingsIfChanged(dir: string, def: Findings): void {
-  const sig = findingsSignature(def);
+export function reportFindingsIfChanged(dir: string, def: Findings, assembly: readonly AssemblyFinding[] = []): void {
```
kid7st: `assembly` defaults to `[]` in a memoized single-door reporter. This module's doc rules out a "record without printing" variant because "a memo entry that trusts some other caller to have printed would silently swallow findings" — an omitted third argument is that variant: it writes a signature without the assembly findings, so a later caller that does pass them re-warns, and an earlier one suppresses them. Every production caller already passes it; make the parameter required.

src/host/node.ts
```diff
@@ -26,6 +26,9 @@ export type Routes = Record<string, ChannelHandler>;
 export interface ChannelContext {
   agent: Agent;
   stateRoot: string;
+  /** Whether the assembled agent can read absolute local attachment paths. Undefined preserves the
+   *  legacy embedder posture (assume available); standard directory serving always supplies a fact. */
+  canReadLocalFiles?: boolean;
```
kid7st: `canReadLocalFiles?: boolean` is a tri-state whose unsafe value is the default: `undefined` means "assume it can read", i.e. hand the model dead paths — the exact failure this PR set out to remove. The three transports then each re-derive the rule as `x !== false` independently.

Every in-repo producer supplies the fact; the optionality exists only for external embedders. Make it required on `ChannelContext` (this surface is pre-1.0), or default it once at the `loadChannels` seam so no transport re-reads the tri-state.

src/cli/commands/info.ts
```diff
@@ -92,7 +103,7 @@ export async function runInfo(dirArg: string, opts: InfoOptions): Promise<void>
           stateRoot,
           sessionsDir,
           authPath,
-          diagnostics: definition.diagnostics,
+          diagnostics: [...definition.diagnostics, ...assemblyFindings],
```
kid7st: `diagnostics` now mixes two concepts. It was `SkillDiagnostic[]` (per-skill parse problems, `path` = the SKILL.md); it now also carries `AssemblyFinding`, whose `path` is the definition dir and whose `code` is outside the skill-diagnostic vocabulary. A machine consumer of `info --json` branching on `code` or resolving `path` gets a second entity in one field.

Emit it as its own key (`assemblyFindings`) and update `docs/cli.md`, which still describes this as "skills and diagnostics".

src/scaffold/templates/persona.md
```diff
@@ -2,11 +2,11 @@
...
 
-Your definition is this directory: `persona.md`, `skills/`, `tools/`, and the config beside them. Your WORKSPACE is the directory you were started in — the project you work on, and where your `read` / `write` / `edit` / `bash` tools operate. It may be this same directory, or the one containing it; `fastagent info` prints both. If the workspace has an `AGENTS.md`, it is project context — read it to learn the project's conventions.
+Your definition is this directory: `persona.md`, `skills/`, `tools/`, and the config beside them. Your WORKSPACE is the directory you were started in — the project you work on. It may be this same directory, or the one containing it; `fastagent info` prints both. Use only the tools actually listed in your system prompt; `codingTools` may narrow or remove `read` / `write` / `edit` / `bash`. If the workspace has an `AGENTS.md`, it is project context — follow it without assuming a file tool is available.
```
kid7st: The default scaffold now hedges capabilities the default scaffold has. `init` produces an agent with all four coding tools, yet the persona it writes says "`codingTools` may narrow or remove read/write/edit/bash", "If `read` is mounted, read `skills/writing-great-skills/SKILL.md`" (line 9) and "edit this file when an editing tool is mounted" (line 10) — conditional prose about a config the file cannot see, paid for by every default author, and it drops the concrete statement of where the workspace tools operate.

The system prompt already lists the mounted tools. Keep the definite text and leave narrowing authors to adjust the persona they own.

src/channels/feishu/invoke-turn.ts
kid7st: Fixed. The guard now runs on the ask's own files, before a referent's join the list; a quoted file with no reader is skipped and counted into the same `unreadableAttachmentsNote` every other degraded resource uses.

The test that encoded the inverted semantics is rewritten as `a quoted file the agent cannot read costs the note, not the turn`: the turn RUNS, nothing is downloaded, and the model is told what it did not get.

src/channels/invoke-turn-kit.ts
kid7st: Fixed, both audiences.

Operator: `LocalFileAccessUnavailable` logs at construction — channel label, the `read` diagnosis, and the `codingTools` knob. That is the only place that knows, and nothing else was telling them.

User: `defaultErrorMessage` now branches on the code and says "I can't open file attachments — please paste the content as text." You were right that this is a configuration limitation rather than an unknown failure; asking the user to rephrase a file is a dead end.

Tests cover both without a custom `onError`.

src/channels/telegram/invoke-turn.ts
kid7st: Took the first option: `ATTACHMENT_UNSUPPORTED_CODE` is exported from `agent.ts` beside `SESSION_BUSY_CODE`, with the consumer named.

It has a real consumer now — `defaultErrorMessage` branches on it to name the limitation. That is what earned it a constant; as a bare literal with nothing reading it, your "drop it" option would have been the right one.

src/channels/telegram/invoke-turn.ts
kid7st: Dropped.

src/engines/pi/create.ts
kid7st: Fixed as suggested: derived from the mounted set via `codingToolNamesIn()`, with `options.codingToolNames` kept as the directory-opener's override.

Test: `does not claim a caller passing \`read\` lacks a file reader` — falsified by restoring the empty default, which brings the warning back.

src/engines/pi/create.ts
kid7st: Right — fixing the assembly site made this the single implementation. `codingToolNamesIn()` is now the one reading, used by both, instead of a correct implementation sitting unused next to an incorrect one.

src/engines/pi/report.ts
kid7st: Made it required. Every production caller already passed it, and the default was exactly the "record without printing" variant the module's single door rules out.

src/host/node.ts
kid7st: Made it required on `ChannelContext`. The three transports now read it directly instead of each re-deriving `x !== false`.

Took the first option rather than defaulting at `loadChannels`: a default there would still be a policy chosen far from where the fact is known, and every in-repo producer already supplies it.

src/cli/commands/info.ts
kid7st: Split into `assemblyFindings`, and `docs/cli.md` updated — it described this as "skills and diagnostics" and now names the third thing.

Test asserts the finding is in the new key and NOT in `diagnostics`.

src/scaffold/templates/persona.md
kid7st: Reverted to the definite text. All three conditionals are gone, and the concrete statement about where the workspace tools operate is back.

Agreed on the principle: the system prompt already lists what is mounted, and an author who narrows owns the persona they are narrowing.

PR #333, opened 2026-08-14
feat(models): custom model endpoints via the agent's own models.json

Sukitly:
## What

Point an agent at a self-hosted model or your own gateway by declaring the endpoint in a
`models.json` next to `fastagent.config.*`:

```json
{
  "providers": {
    "mygw": {
      "baseUrl": "https://llm-proxy.internal/v1",
      "api": "openai-completions",
      "apiKey": "$MYGW_API_KEY",
      "models": [{ "id": "deepseek-v4-pro" }]
    }
  }
}
```

```ts
export default defineConfig({ model: "mygw/deepseek-v4-pro", deploy: { secrets: ["MYGW_API_KEY"] } });
```

Works on every path: `dev`, `start`, `invoke`, `chat`, `deploy`'s pre-flight, and L2 embedding.
Custom endpoints are additive — built-in providers stay available. Giving an existing provider a new
`baseUrl` retargets it through a proxy while keeping its models, pricing and compat metadata.

## Why this shape

**The capability is pi's, not new.** The schema is pi's own `models.json` (`docs/models.md`),
unchanged — so the full field reference, every `compat` flag and per-model override transfer, and the
docs point there instead of duplicating it.

**What was actually rejected was the file's LOCATION, not its format.** `createPiModelRuntime` pinned
`modelsPath: null` because pi's file lives at `~/.pi/agent/models.json` — builder-machine state that
`deploy` does not bake, so an endpoint defined there would work locally and vanish on deploy. A
`models.json` inside the agent dir is definition data: it travels with the artifact. The
machine-global file stays unread, with no fallback — a fallback is exactly the "works here, breaks
there" failure the original decision was protecting against.

`docs/configuration.md` listed custom model providers under "deliberately not config" alongside
custom session stores and distributed leases. That grouping was wrong: those are *code*, an endpoint
is *data*. Declarative endpoints now live in `models.json`; `providers` (embedding) remains the
injection point for a provider that needs code, e.g. minting a token per request.

## Two upstream behaviors this must not pass through

| Behavior | Why it cannot be transparent |
|---|---|
| `modelsStorePath` defaults to `<dirname(modelsPath)>/models-store.json` | With the file in the agent dir, pi's generated catalog cache lands in the author's definition dir — and `deploy` bakes that tree into the image. Pinned to the state root. |
| A malformed `models.json` does not throw | `ModelRuntime.create` resolves with the built-ins and parks the reason in `getError()`. Unread, a typo resurfaces later as a bare `unknown model`. Surfaced as a throw, which matches `preflight`'s documented "throws on a real fault" contract. |

## Structure

`createPiModels` (pi-ai `builtinModels`) and `createPiModelRuntime` (`ModelRuntime`) were two model
hubs, and only the second can see `models.json`. The directory rungs now assemble against
`ModelRuntime`, which `implements Models`, so this removes a duplicate hub rather than adding an
abstraction.

`models.json` is a DIRECTORY concept, so the split follows placement: L1 `createPiAgent` — assembly
from typed parts, no directory to read — keeps the synchronous built-ins path and its public
signature; L2 `createPiAgentFromDefinition`, already async, builds the runtime and hands it down.
Verified before switching: `ModelRuntime`'s built-in provider set is identical to `builtinModels()`
(38 each, no difference either way), so no model spec changes meaning, and its default auth context
is the same function fastagent passed explicitly.

Two callers stay on the built-ins by design: `fastagent models` answers "what does FastAgent
support", not "what does this agent use" (`fastagent info` answers that), and login verification only
checks stored credentials, which a `models.json` endpoint does not use.

## Verification

`npm run lint && npm run typecheck && npm test` green (1337 tests).

Four assumptions about upstream were probed before implementing, not inferred from types: loading
from a non-default `modelsPath`, `$ENV_VAR` interpolation outside the TUI, the `modelsStorePath`
default, and built-in provider set equality. The serving-path test was checked to fail without the
change (`unknown model "mygw/deepseek-v3"`).

Also run end to end against a real OpenAI-compatible gateway: single turn, a tool-calling turn, and
`start` + `/invoke` SSE — including `thinking` deltas, which is the only way to confirm a `compat`
flag is right.

## Not included

- `fastagent models` is not directory-aware (documented).
- `fastagent login` does not offer custom providers — such an endpoint authenticates from its own
  `apiKey` (`$ENV` / `!command`), so there is no credential to store.
- No dedicated tests for the `deploy` pre-flight and `reportAuth` dir-awareness; both are pass-the-
  runtime changes covered by the end-to-end run.

### DIFF
--- a/src/cli/commands/info.ts
+++ b/src/cli/commands/info.ts
@@ -6,9 +6,11 @@ import {
   defaultSessionsDir,
   loadConfig,
   resolveAuthPath,
+  resolveModel,
   resolveModelSpec,
   resolveSessionsDirOverride,
 } from "../../engines/pi/config.ts";
+import { createPiModelRuntime } from "../../engines/pi/models.ts";
 import { resolveStateRoot, workspaceHint } from "../../paths.ts";
 import { resolveAgentTools } from "../../engines/pi/create.ts";
 import { loadAgentDefinition } from "../../engines/pi/definition.ts";
@@ -70,6 +72,21 @@ export async function runInfo(dirArg: string, opts: InfoOptions): Promise<void> 
   const sessionsDir = resolveSessionsDirOverride(opts.sessionsDir) ?? defaultSessionsDir(stateRoot);
   const authPath = resolveAuthPath(agentDir, opts.authPath); // flag > FASTAGENT_AUTH_PATH > default — the one owner
 
+  // RESOLVE the spec, do not just echo it: a spec is only real once its provider/model exist in the
+  // agent's own surface (built-ins + its models.json), which is exactly what a custom endpoint changes.
+  // Reporting a healthy-looking spec that `dev`/`start` then reject is the failure this pre-empts.
+  // Reported as DATA rather than thrown — a broken agent is what `info` is for — and read-only: the
+  // runtime reads models.json without creating anything (its catalog cache is written on refresh, and
+  // there is none here).
+  const modelError = modelSpec
+    ? await createPiModelRuntime({ agentDir, authPath })
+        .then((models) => {
+          resolveModel(models, modelSpec);
+          return undefined as string | undefined;
+        })
+        .catch((error: unknown) => (error as Error).message)
+    : undefined;
+
   if (opts.json) {
     console.log(
       JSON.stringify(
@@ -78,6 +95,7 @@ export async function runInfo(dirArg: string, opts: InfoOptions): Promise<void> 
           workspace,
           configPath: configPath ?? null,
           model: modelSpec ?? null,
+          modelError: modelError ?? null,
           thinkingLevel: config.thinkingLevel ?? null,
           context: definition.contextFiles.map((f) => f.path),
           persona: definition.persona !== undefined,
@@ -106,12 +124,15 @@ export async function runInfo(dirArg: string, opts: InfoOptions): Promise<void> 
   // One padded label writer: hand-spaced labels drifted out of alignment the moment a longer one
   // (agent/workspace/selfSchedule) joined the report.
   const line = (label: string, value: string): void => console.log(`${`${label}:`.padEnd(13)} ${value}`);
+  /** A continuation under the previous line, aligned to the same column (no label, so no bare colon). */
+  const cont = (value: string): void => console.log(`${"".padEnd(13)} ${value}`);
   line("agent", agentDir);
   line("workspace", workspace);
   const hint = workspaceHint({ agentDir, workspace });
   if (hint) line("hint", hint);
   line("config", configPath ?? "(none)");
   line("model", modelSpec ?? "(not set — pass --model, set FASTAGENT_MODEL, or config.model)");
+  if (modelError) cont(`⚠ does not resolve: ${modelError}`);
   if (config.thinkingLevel) line("thinking", config.thinkingLevel);
   line("context", definition.contextFiles.map((f) => f.path).join(", ") || "(none)");
   line("persona", definition.persona ? "persona.md" : "(none)");
--- a/src/cli/shared.ts
+++ b/src/cli/shared.ts
@@ -20,7 +20,13 @@ import {
   rewriteConfigModel,
 } from "../engines/pi/config.ts";
 import { LoginCancelled, type LoginIO, type LoginMethod, type LoginResult, loginFlow } from "../engines/pi/login.ts";
-import { createPiModels, probeApiKey, probeAuthSource, providerAuthStatuses } from "../engines/pi/models.ts";
+import {
+  createPiModelRuntime,
+  createPiModels,
+  probeApiKey,
+  probeAuthSource,
+  providerAuthStatuses,
+} from "../engines/pi/models.ts";
 import { formatAuthReport } from "./auth-view.ts";
 import { log } from "../log.ts";
 import { openExternalUrl } from "../open-url.ts";
@@ -80,10 +86,13 @@ export function parseBind(value: string | undefined): string | undefined {
   return bindAddress(trimmed); // a name never travels past this point — see bind.ts
 }
 
-/** Report which source provides the model's credentials, surfacing a remediation hint at startup. Non-blocking. */
-export async function reportAuth(modelSpec: string, authPath: string): Promise<void> {
+/** Report which source provides the model's credentials, surfacing a remediation hint at startup. Non-blocking.
+ *  Probes through the AGENT's model surface (`agentDir` carries its models.json), so a custom endpoint is
+ *  reported like any built-in rather than as an unknown provider. */
+export async function reportAuth(agentDir: string, modelSpec: string, authPath: string): Promise<void> {
   const provider = providerOf(modelSpec);
-  const source = await probeAuthSource(createPiModels({ authPath }), modelSpec);
+  const models = await createPiModelRuntime({ agentDir, authPath }).catch(failStartup);
+  const source = await probeAuthSource(models, modelSpec);
   // Only when nothing satisfies auth do we read the store (refresh-FREE) to tell "nothing stored" from
   // "stored but unusable" — see formatAuthReport for why. store.read warns on a corrupt file itself.
   const stored =
@@ -119,7 +128,9 @@ export async function resolveFirstRunModel(
   if (!isInteractive()) return; // CI/deploy: the opener throws the actionable missing-model error
 
   const authPath = resolveAuthPath(agentDir, options.authPath);
-  const models = createPiModels({ authPath });
+  // The picker lists the AGENT's surface: built-ins plus whatever its models.json declares, so a
+  // self-hosted endpoint is pickable on first run instead of being invisible until hand-set.
+  const models = await createPiModelRuntime({ agentDir, authPath }).catch(failStartup);
   const chosen = await pickWithCredentials(models, authPath);
   if (chosen === undefined) return; // cancelled (or auth probe failed): the caller raises its clear missing-model error
   process.env.FASTAGENT_MODEL = chosen; // this process + any spawned dev worker inherits it
@@ -234,6 +245,8 @@ async function verifyApiKeyLogin(
   authPath: string,
   spec?: string,
 ): Promise<"ok" | "rejected" | "unknown"> {
+  // Built-ins only: `login` itself offers built-in providers (login.ts), and a models.json endpoint
+  // authenticates from its own `apiKey` (env/command), so there is no stored credential to verify here.
   const models = createPiModels({ authPath });
   const model = spec ? resolveModel(models, spec) : models.getProvider(provider)?.getModels()[0];
   if (!model) {
--- a/src/deploy/preflight.ts
+++ b/src/deploy/preflight.ts
@@ -18,7 +18,7 @@ import { resolveAuthPath } from "../engines/pi/config.ts";
 import { type ResolvedPlacement, resolveSecretsDir, resolveStateRoot } from "../paths.ts";
 import { inspectChannels } from "../engines/pi/channel.ts";
 import { discoverScheduleFiles } from "../schedule/discover.ts";
-import { createPiModels, probeAuthSource } from "../engines/pi/models.ts";
+import { createPiModelRuntime, probeAuthSource } from "../engines/pi/models.ts";
 import { CHANNEL_KINDS, type ChannelKind } from "../scaffold/add-channel.ts";
 import { exists } from "../paths.ts";
 import { detectRuntime, readPackageJson } from "../runtime.ts";
@@ -194,9 +194,13 @@ export async function preflightDeploy(input: {
   }
 
   // Probe auth from the SAME project-level file the opener/login use — not the global default, which would
-  // miss a `fastagent login` credential and falsely report "none configured".
+  // miss a `fastagent login` credential and falsely report "none configured". Through the AGENT's model
+  // surface too (its models.json travels into the image), so a custom endpoint is not read as an unknown
+  // provider — this probe feeds the gate that decides whether `--run` may proceed.
   const authPath = resolveAuthPath(agentDir, authPathFlag);
-  const modelAuth = modelSpec ? await probeAuthSource(createPiModels({ authPath }), modelSpec) : undefined;
+  const modelAuth = modelSpec
+    ? await probeAuthSource(await createPiModelRuntime({ agentDir, authPath }), modelSpec)
+    : undefined;
 
   // Container facts (shared by every host) + the warnings that follow. The facts describe the AGENT —
   // its package.json/runtime/lockfile drive the image's install step — never the workspace's (the bake
--- a/src/deploy/preflight.ts
+++ b/src/deploy/preflight.ts
@@ -18,12 +18,13 @@ import { resolveAuthPath } from "../engines/pi/config.ts";
 import { type ResolvedPlacement, resolveSecretsDir, resolveStateRoot } from "../paths.ts";
 import { inspectChannels } from "../engines/pi/channel.ts";
 import { discoverScheduleFiles } from "../schedule/discover.ts";
-import { createPiModelRuntime, probeAuthSource } from "../engines/pi/models.ts";
+import { createPiModelRuntime, modelCredentialCarry, probeAuthSource } from "../engines/pi/models.ts";
 import { CHANNEL_KINDS, type ChannelKind } from "../scaffold/add-channel.ts";
 import { exists } from "../paths.ts";
 import { detectRuntime, readPackageJson } from "../runtime.ts";
 import { fastagentVersion } from "../version.ts";
 import { type ContainerInput, isGeneratedDockerfile, isGeneratedDockerignore } from "./container.ts";
+import { isEnvKey } from "./secrets.ts";
 
 /** A stderr line the CLI prints (`[fastagent] warn: …` / `[fastagent] note: …`). Host-neutral advisories. */
 interface DeployMessage {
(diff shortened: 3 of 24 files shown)


kid7st (commented):
The core direction looks right, but I found one dev-loop blocker and two consistency issues:

1. **`fastagent dev` does not watch `models.json`.** `createPiModelRuntime()` loads the file once during worker assembly, while `devWatchIgnored()` currently ignores every root file except `fastagent.config.*` and `package.json`. Editing/adding `models.json` therefore leaves the running worker on stale model configuration. More importantly, if a malformed `models.json` prevents the worker from starting, fixing that same file does not trigger the supervisor's recovery path. Please add `models.json` to the root allowlist, `WATCHED_HINT`, and `test/dev-supervisor.test.ts`.

2. **`fastagent info` does not validate what the docs say it resolves.** The new configuration docs tell users to run `fastagent info` to confirm what the agent resolved, but `runInfo()` only prints `resolveModelSpec()`; it never loads `models.json`, checks that the provider/model exists, or surfaces a malformed file. Thus `info` can report a healthy-looking custom spec immediately before `dev`/`start` fail. Either make `info` load and resolve against the agent-local runtime (without violating its read-only guarantee), or narrow the documentation to say it only displays the selected spec.

3. **The documented same-ID precedence is reversed.** `docs/embedding.md` says an injected `providers` entry with a matching ID overrides `models.json`. `ModelRuntime.registerNativeProvider()` actually installs the native provider as the base and then composes the `models.json` config over it; Pi documents this as “models.json overrides above registered native providers.” Please correct the docs/options comments and add a same-ID precedence test rather than promising replacement semantics.

CI is green and the main runtime/cache/error-handling design otherwise looks sound.

Sukitly:
All three confirmed and fixed in 66e6dfc. Thank you — #1 was a regression this branch introduced, and #3 was a claim I made without testing it.

**1. `dev` watch.** Right, and worse than stale config: because a malformed `models.json` now fails assembly, the worker dies and the edit that repairs it was not an edit that restarts — the author is stranded holding a correct file. Added to the root allowlist, `WATCHED_HINT`, and `test/dev-supervisor.test.ts`.

Rather than spelling the name in the watcher, it moved to `paths.ts` as `AGENT_MODELS_FILE`, which is the rule the config names already follow ("the config NAMES come from paths.ts, not a regex spelled here") and for exactly this reason: the restart scope must not be able to drift from what the worker loads.

**2. `info`.** Agreed the docs wrote a cheque `info` did not cash. I took the first option — `info` now resolves the spec against the agent-local runtime:

```
model:        mygw/not-declared
              ⚠ does not resolve: unknown model "mygw/not-declared" (provider "mygw" / id "not-declared" not in registry); run `fastagent models` to list available specs
```

`--json` gains `modelError: string | null`. Read-only is preserved (no state root, no sessions dir; the catalog cache is written on refresh and there is none here) and it stays exit-0 with the verdict as data, since diagnosing a broken agent is what `info` is for. Note this also closes the same hole for built-in providers — `info` never validated those specs either. Covered in `test/cli.test.ts`.

**3. Precedence.** You are right, and I had it backwards. Probed it rather than trusting either reading:

```
models.json: dup → https://FROM-MODELS-JSON/v1     (contextWindow unset → default 128000)
injected:    dup → https://FROM-NATIVE-PROVIDER/v1 (contextWindow 111)
→ resolved: baseUrl https://FROM-MODELS-JSON/v1, contextWindow 128000
```

The file composes over the injected provider, descriptor included. `docs/embedding.md` and the `providers` option comment now say so, with a test pinning the direction.

Worth stating why this is the desirable direction and not just upstream's accident: where a deployed agent's traffic goes is a property of its definition, not of the program that embedded it. The sharp edge is that a same-id entry *silently* replaces an injected endpoint, so both the doc and the option comment now tell callers to use a distinct id when they mean both to exist.

`npm run lint && npm run typecheck && npm test` green (1339).

src/deploy/preflight.ts
```diff
@@ -194,9 +194,13 @@ export async function preflightDeploy(input: {
   }
 
   // Probe auth from the SAME project-level file the opener/login use — not the global default, which would
-  // miss a `fastagent login` credential and falsely report "none configured".
+  // miss a `fastagent login` credential and falsely report "none configured". Through the AGENT's model
+  // surface too (its models.json travels into the image), so a custom endpoint is not read as an unknown
+  // provider — this probe feeds the gate that decides whether `--run` may proceed.
```
kid7st: The probe is now endpoint-aware, but the secret it discovers still cannot travel. For a models.json provider `probeAuthSource` returns the literal label `"configured API key"`, so `isEnvKey` is false downstream and the gateway key is neither listed in the runbook nor carried by `--run`; `needsModelCredential` then stops the deploy with "run `fastagent login`", which cannot serve this provider. This comment's claim that the probe "feeds the gate that decides whether `--run` may proceed" is accurate and is precisely the problem — the gate is now fed a value it cannot act on. Direction: extract the `$VAR` names declared in models.json (`apiKey`, `headers`) and pass them as required secret names into `deploymentSecrets`/`assembleSecrets`.

kid7st (commented):
x

src/cli/shared.ts
```diff
@@ -80,10 +86,13 @@ export function parseBind(value: string | undefined): string | undefined {
   return bindAddress(trimmed); // a name never travels past this point — see bind.ts
 }
 
-/** Report which source provides the model's credentials, surfacing a remediation hint at startup. Non-blocking. */
-export async function reportAuth(modelSpec: string, authPath: string): Promise<void> {
+/** Report which source provides the model's credentials, surfacing a remediation hint at startup. Non-blocking.
```
kid7st: Two problems on this function.

1. Remediation is misassigned for a custom endpoint. When `$MYGW_API_KEY` is unset, upstream `resolve` throws → `probeAuthSource` returns undefined → `formatAuthReport` prints `(none found)` with the hint "run `fastagent login`, or set the provider's API key in .env". Neither branch fixes it: the variable is named in `models.json`, and `login` cannot see this provider. Same responsibility error as the deploy gate. Direction: carry the provider's declared key source into `formatAuthReport` so the hint names the actual env var.

2. The docstring still says "Non-blocking", but the body now does `await createPiModelRuntime(...).catch(failStartup)` — a process exit. Correct the sentence or drop it.

src/cli/commands/info.ts
```diff
@@ -70,6 +72,21 @@ export async function runInfo(dirArg: string, opts: InfoOptions): Promise<void>
   const sessionsDir = resolveSessionsDirOverride(opts.sessionsDir) ?? defaultSessionsDir(stateRoot);
   const authPath = resolveAuthPath(agentDir, opts.authPath); // flag > FASTAGENT_AUTH_PATH > default — the one owner
 
+  // RESOLVE the spec, do not just echo it: a spec is only real once its provider/model exist in the
+  // agent's own surface (built-ins + its models.json), which is exactly what a custom endpoint changes.
+  // Reporting a healthy-looking spec that `dev`/`start` then reject is the failure this pre-empts.
+  // Reported as DATA rather than thrown — a broken agent is what `info` is for — and read-only: the
+  // runtime reads models.json without creating anything (its catalog cache is written on refresh, and
+  // there is none here).
+  const modelError = modelSpec
```
kid7st: `modelError` conflates two distinct faults, and hides one of them entirely.

The value is rendered as `⚠ does not resolve: …`, but it also catches `createPiModelRuntime`'s throw — i.e. a models.json PARSE error surfaces as a SPEC resolution error. And the whole block is gated on `modelSpec`, so an agent with no model set and a malformed models.json — which will kill `dev`/`start` at assembly — gets a clean `info` report. That inverts the command's own stated contract ("fix only what it reports", per the schedules comment above).

Direction: build the runtime unconditionally and report a load failure as its own field/line (`models.json: <error>`), keeping `modelError` for spec resolution only.

src/engines/pi/models.ts
```diff
@@ -4,11 +4,23 @@
...
  */
+import { join } from "node:path";
 import { type Api, type Model, type Models, type Provider, defaultProviderAuthContext } from "@earendil-works/pi-ai";
 import { builtinModels } from "@earendil-works/pi-ai/providers/all";
 import { ModelRuntime } from "@earendil-works/pi-coding-agent";
 import { type FastagentAuthOptions, fastagentCredentialStore } from "./auth.ts";
 import { type InteractiveLoginKind, interactiveLoginKind } from "./login.ts";
+import { AGENT_MODELS_FILE, resolveStateRoot } from "../../paths.ts";
+
+/** The DEFINITION-LOCAL custom-endpoint file, in pi's own models.json schema (see pi's docs/models.md):
```
kid7st: This doc block documents `AGENT_MODELS_FILE`, which is declared in `paths.ts` — and being the leading comment of `export interface CreatePiModelsOptions`, it is that interface's hover text that it becomes. Its content also restates the `paths.ts` docstring and `docs/configuration.md`. Direction: keep the rationale at the single declaration in `paths.ts`; leave at most a one-line pointer here.

src/engines/pi/models.ts
```diff
@@ -37,20 +49,50 @@ export function createPiModels(options: CreatePiModelsOptions = {}): Models {
...
+    ...(agentDir
+      ? { modelsStorePath: join(options.stateRoot ?? resolveStateRoot(agentDir), "models-store.json") }
+      : {}),
     allowModelNetwork: false,
   });
+  // A malformed models.json does NOT throw upstream — `create` resolves with the built-ins and parks the
+  // reason in getError(). Left unread, a typo'd endpoint would silently degrade to "provider not in
+  // registry" at model-resolution time, i.e. the silent fallback this codebase forbids. The upstream
+  // message already names both the reason and the file, so it is surfaced verbatim.
+  const error = runtime.getError();
```
kid7st: `getError()` is not the parse-error channel this comment describes. Upstream aggregates the config error, per-provider `compositionErrors`, and `availabilityError`. Throwing on the union means a composition failure in a provider the agent never selects aborts startup for every command — under a message this comment promises "names both the reason and the file", which for a composition error it does not.

Also ordering: `registerNativeProvider` runs on the line below, after the check, so a composition error introduced by an injected provider is never read.

Direction: move the check below the registration loop, and either narrow to the config error or word the throw so a non-parse cause is not mislabelled as a malformed file.

test/models.test.ts
```diff
@@ -94,3 +100,143 @@ describe("probeApiKey (the post-login quick-fail check)", () => {
...
+    expect(model.contextWindow).toBe(65536);
+    expect(model.maxTokens).toBe(16384);
+    // Built-ins are kept alongside, so a custom endpoint is additive, never a replacement.
+    expect(runtime.getProvider("anthropic")).toBeDefined();
+
+    // `apiKey: "$ENV"` must interpolate on the SERVING path too (pi documents it for the TUI): the key
+    // stays out of the file, which is what lets models.json be committed and baked into an image.
+    process.env.FASTAGENT_TEST_GW_KEY = "sk-from-env";
+    try {
+      expect(await probeAuthSource(runtime, "mygw/deepseek-v3")).toBeDefined();
```
kid7st: This assertion cannot fail for the scenario the test name promises. `probeAuthSource` returns the constant label `"configured API key"` for any configured key, so `toBeDefined()` would also pass if pi treated `"$FASTAGENT_TEST_GW_KEY"` as a LITERAL key — the `$` interpolation, which is the stated subject ("its key comes from the environment"), is untested. There is also no negative case, though it is the branch authors actually hit: with the variable unset, upstream `resolveConfigValueOrThrow` throws and the probe swallows it to undefined.

Direction: assert the resolved credential (`(await runtime.getAuth(model))?.auth.apiKey === "sk-from-env"`) and add the unset-env case.

src/cli/shared.ts
```diff
@@ -119,7 +128,9 @@ export async function resolveFirstRunModel(
   if (!isInteractive()) return; // CI/deploy: the opener throws the actionable missing-model error
 
   const authPath = resolveAuthPath(agentDir, options.authPath);
-  const models = createPiModels({ authPath });
+  // The picker lists the AGENT's surface: built-ins plus whatever its models.json declares, so a
+  // self-hosted endpoint is pickable on first run instead of being invisible until hand-set.
+  const models = await createPiModelRuntime({ agentDir, authPath }).catch(failStartup);
```
kid7st, line 133: The first-run picker now lists models.json providers, but the post-pick auth policy below cannot serve them. Verified against a real `models.json` provider: `interactiveLoginKind()` returns `"api_key"` (pi builds `auth.apiKey.login` for a file-declared provider), so `pickWithCredentials` never takes the `login === "none"` branch whose message is the correct remedy ("set its API key env var"); it takes the inline-login branch instead, and `loginFlow` resolves against `builtinProviders()` → throws `unknown provider "mygw"`, which surfaces as `login for "mygw" failed: ... — model saved; run \`fastagent login\` to fix auth`. That remedy is a flow which structurally cannot handle the provider. Second effect: when the env var is unset, `models.getAuth()` THROWS (`Failed to resolve API key ... from environment variable: MYGW_KEY`), so `providerAuthStatuses` classifies a perfectly fine but not-yet-exported endpoint as `broken` rather than `unconfigured`. Direction: classify a provider that is not in `builtinProviders()` as login `"none"` (and give that branch a models.json-aware remedy naming the env var), so the picker's hint predicts what picking actually does.

src/cli/shared.ts
```diff
@@ -80,10 +86,13 @@ export function parseBind(value: string | undefined): string | undefined {
   return bindAddress(trimmed); // a name never travels past this point — see bind.ts
 }
 
-/** Report which source provides the model's credentials, surfacing a remediation hint at startup. Non-blocking. */
-export async function reportAuth(modelSpec: string, authPath: string): Promise<void> {
+/** Report which source provides the model's credentials, surfacing a remediation hint at startup. Non-blocking.
+ *  Probes through the AGENT's model surface (`agentDir` carries its models.json), so a custom endpoint is
+ *  reported like any built-in rather than as an unknown provider. */
+export async function reportAuth(agentDir: string, modelSpec: string, authPath: string): Promise<void> {
```
kid7st, line 92: Two things drifted with the new signature. (1) The docstring still says "Non-blocking", but line 94 now ends in `.catch(failStartup)` — this function can `process.exit(1)`. It is currently unreachable only because every caller opens the agent first; the companion text no longer describes the code. (2) `reportAuth` takes `agentDir` and rebuilds a SECOND `ModelRuntime` (re-reads models.json, re-creates the credential store) for a surface the opener already built and threw away — two hubs per boot that can disagree if the file changes in between. The opener is the one that knows the assembled surface; direction is to thread the `Models` it built (or expose it on the opener's result) into `reportAuth` rather than have the reporter re-derive it from the directory.

src/cli/commands/info.ts
kid7st, line 81: `modelError` is computed only when `modelSpec` is set, so an agent with a malformed `models.json` and no model configured reports nothing wrong — in exactly the command the docs point at to diagnose a broken agent, and for exactly the failure mode this PR makes fatal for `dev`/`start`. Also, when the spec IS set, a parse error is rendered under the model line as `⚠ does not resolve: <parse error>`, which attributes a file-level fault to the spec. Direction: build the runtime unconditionally when `models.json` exists and report its failure as its own line/JSON field (a `models.json` fact), with spec resolution as a separate check on top.

kid7st (commented):
Thesis: `models.json` next to the config becomes definition-local model-endpoint data, and every directory rung (`dev`/`start`/`invoke`/`chat`/`info`/preflight) now assembles against `ModelRuntime` so it is in scope.

**`deploy --run` hard-gates an agent whose model is a models.json endpoint (missing change, `src/deploy/secrets.ts`)**

Probed against a real endpoint declared as `apiKey: "$MYGW_KEY"`: `probeAuthSource` returns the label `"configured API key"`, not the env-var name. Downstream in `assembleSecrets`, `isEnvKey("configured API key")` is false and a models.json-only agent has no `auth.json`, so it falls into `needsModelCredential = true` — and `runFlyDeploy`/`runRailwayDeploy`/`deploy/docker/run.ts` stop with `no model credential — run \`fastagent login\`, or set a provider API key in .env`. Both remedies are wrong for this agent: `fastagent login` cannot serve a custom provider (`loginFlow` resolves against `builtinProviders()`), and the key already IS in the environment and already travels via `deploy.secrets` → `extraSecrets`. `docs/configuration.md`'s promise ("list the NAME in `deploy.secrets`, so `deploy` carries the value to the host") is therefore contradicted on the `--run` path. The credential model has three states (env key / stored login / none) and a models.json endpoint is a fourth — self-carrying — that it cannot express; the PR extends the model surface into preflight without extending it. Direction: make preflight report that the resolved model's provider carries its own `apiKey` (e.g. a fact alongside `modelAuth`) and treat that as satisfying the credential gate, with the declared `deploy.secrets` name being the thing that must have a local value. The PR's "no dedicated tests for the deploy pre-flight ... covered by the end-to-end run" does not cover it: the e2e run was `start` + `/invoke`, not `deploy --run`.

Related: `test/models.test.ts` asserts only `expect(await probeAuthSource(runtime, "mygw/deepseek-v3")).toBeDefined()`. The SHAPE of that label is what deploy branches on, so asserting the value (`"configured API key"`) is what would have surfaced the gate above.

Sukitly:
Confirmed and fixed in b608b3b. This one was a real hole, and the diagnosis was exactly right — thank you for probing `--run` specifically, since my end-to-end run was `start` + `/invoke` and would never have touched it.

**The root cause is the probe, not the gate.** `probeAuthSource` answers "is this authenticated here", which is not the question deploy asks ("how does the credential *reach the host*"). It flattens three distinct cases into one display label:

| models.json `apiKey` | `probeAuthSource` | `getProviderAuthStatus` |
|---|---|---|
| `"$GATE_KEY"` | `configured API key` | `{configured: true, source: "environment", label: "GATE_KEY"}` |
| `"sk-literal"` | `configured API key` | `{configured: true, source: "models_json_key"}` |
| `"!echo sk"` | `configured API key` | `{configured: true, source: "models_json_command"}` |

The second column is what preflight was reporting, so `isEnvKey` rejected it and — with no `auth.json` — every models.json agent fell into `needsModelCredential`.

So rather than teaching the gate about a fourth state, preflight now asks the right question (`modelCredentialCarry`). For the env case `modelAuth` becomes **the variable name**, which is the shape the existing machinery already speaks: the runbook lists it, `assembleSecrets` carries the value, and a missing local value gets the `.env` remediation instead of the login one. Nothing new threads through that path.

A consequence worth flagging: the author no longer needs `deploy.secrets` for the model key at all — deploy infers it. `docs/configuration.md` now says so, and reframes `deploy.secrets` as the place for what cannot be inferred (a key used only in `headers`, or a value assembled from several variables like `"${A}_${B}"`, which has no single name to carry).

Your "fourth state" reading is right for the other two rows, though: a key written *into* the file travels with the file, so there is nothing to carry **and nothing to gate** — gating there would print two remedies that are both impossible. That is `modelKeyInDefinition`, and it only suppresses the gate; it never invents a secret.

**On the test.** You are right that asserting `toBeDefined()` was the miss — the *shape* of that label is what deploy branches on. It now pins both the label and the carry:

```ts
expect(await probeAuthSource(runtime, "mygw/deepseek-v3")).toBe("configured API key");
expect(modelCredentialCarry(runtime, "mygw/deepseek-v3")).toEqual({
  envVar: "FASTAGENT_TEST_GW_KEY",
  inDefinition: false,
});
```

Plus preflight tests for both directions and an `assembleSecrets` test for the suppressed gate. All three were checked to fail without the fix (`git stash` the two source files → 3 failures, restore → green).

Rebased onto the branch's merge of main. `npm run lint && npm run typecheck && npm test` green (1369).

PR #335, opened 2026-08-16
refactor(pi): session creation is one atomic primitive

kid7st:
Follow-up to #328's review. Independent of #330, which is where this work was first written.

## Why

#328's review found the JSONL fork was not atomic: pi's `create` and `fork` publish the file before appending entries, while `openOrCreate` short-circuits on EXISTENCE — so a crash mid-inheritance left a discoverable child that read as "the decision was taken" forever, holding whatever half it had. That was fixed by adding a draft realm and an atomic rename, exposed as **three** verbs on the backend seam (`createDraft` / `forkDraft` / `publish`).

Three verbs put the sequencing in the callers as a rule rather than in the type as a structure, and a rule is only as good as the next caller. Building on that seam produced two further defects of exactly that shape: a birth mark appended after the session was already discoverable, and a second creation for the same id reachable from a catch path.

## What

One verb, one contract:

```ts
createAtomically(id, from, fill): Promise<Session>
```

**`id` resolves to a complete session or to nothing, never to a half.** Retrying is therefore always safe, so a failed attempt needs no cleanup and the fallback is simply the next attempt — "exactly one creation" stops depending on how many times callers call. `createInheriting` keeps only the semantics: where to branch, and what the newborn is born with.

## What comes with it

**A path base, defined because the refactor needs one.** Not a pre-existing defect — `main` derives the publish target from the parent session's metadata path, which pi has already resolved, so a relative `dir` works there today (verified). But `createAtomically` must also create sessions with no parent, and a parentless draft has no parent path to borrow, so the target has to be derived from the store's own root. That makes the root's meaning load-bearing, and this store has two path bases: pi resolves a relative `sessionsRoot` against `cwd`, while every direct `fs` call here resolves against `process.cwd()`. `dir` is therefore resolved once at the entry, by pi's rule (`NodeExecutionEnv.absolutePath` is `resolve(cwd, path)`). The behaviour is unchanged for every in-tree caller, all of which pass absolute paths; what changes is that a relative `dir` now has a stated meaning rather than an accidental one.

**A defect that does close: `encodeSessionId` is now injective.** Flagged during #328's review and deferred, it now sits in the path that carries `parentSession`, where a collision is not two rooms sharing a file but a thread inheriting from the wrong one. The previous form padded to a *minimum* of two hex digits, so an escape run could be re-split: the id `U+0100`, and the id `U+0010` followed by the digit `0`, both encoded to `%100`. Every escape now has a self-describing width — `%XX`, or `%uXXXX` above one byte — so no two inputs can produce one output. ASCII ids, which is every id the built-in channels mint, encode exactly as before: **existing session files keep their names.**

## Tests

Three new, all mutation-verified: restoring the variable-width encoding fails the same-file test; publishing off the raw `dir` option fails the relative-dir test, which pins the base the refactor now depends on; and staging creation in the real directory instead of the draft realm fails the staging test.

That last one closes a gap review found rather than a defect: the property this branch exists for had no test. Removing the staging passed the whole inheritance suite, because publish's rename then targets the path the file already has — a no-op, the same end state, and only the crash window differs. It is a structural check that the mechanism ran; proving crash safety would need a failure injected between the draft and the rename, which would mean a seam in production code that exists only for the test.

The rest of the inheritance suite — atomic publish, invisible drafts, exactly one file after a failure — passes unchanged against the new seam.

The staging realm is `.drafts`, and it holds creations reached through the inheritance path: the fork, and that path's empty fallback. An ordinary create with no parent writes straight into the store — an empty session is complete the moment its file exists, so there is no window to protect.


kid7st (commented):
Thesis: `SessionBackend.createAtomically` replaces a three-verb draft protocol with one primitive whose contract is that an id resolves to a complete session or to nothing — so retrying is safe, the fallback is just the next attempt, and "exactly one creation" stops depending on callers sequencing anything.

## Finding — the published ports' documented shape does not match what they accept

`docs/api-reference.md` (the "Sessions and leases" block) is not in this diff, which is the problem:

```ts
interface PiSessionStore {
  openOrCreate(sessionId: string): Promise<Session>;
}
function jsonlSessionStore(options: { dir: string; cwd?: string }): PiSessionStore & PiSessionReader;
```

**`dir` gains a resolution rule here, and it is undocumented.** This PR's own justification names external callers as the audience — *"jsonlSessionStore is a published injection port whose signature asks nothing of `dir`, which is who this is for"* — and after this change a relative `dir` has a definite meaning: resolved against `cwd` (itself defaulting to `process.cwd()`), never against the process directory when the two differ. Nothing in the reference says so, so the behaviour fixed for external callers is not stated to them.

**`openOrCreate` is missing its second parameter** — pre-existing, from #328, not introduced here. The implementation is `openOrCreate(sessionId, inherit?: SessionInheritance)`, and `SessionInheritance` is not exported from `/pi`. An external implementer can still write the method (the parameter type comes from contextual typing — checked), but cannot name the type. A published port whose parameter type is internal is half-published.

Direction: add one line under that block stating the `dir` rule; update the `PiSessionStore` signature to carry `inherit?`; export `SessionInheritance` from `/pi` alongside the port that references it.

## Checked and not filed

Recorded so the next reviewer does not re-derive them — each was a real suspicion, refuted by measurement rather than by argument:

| Suspicion | Measured |
|---|---|
| A persistent fork failure accumulates drafts under `.inherit-tmp`, making "bounded by crash count" wrong | Tearing a parent journal and inheriting from it five times leaves `drafts=0` — `fork` fails while reading the source, before the target file exists |
| Nothing proves the parentless path publishes out of the draft realm now that it goes through it | Mutating it to return the draft unpublished fails the existing "a failed inheritance leaves exactly one file" test |
| The in-memory backend has no draft realm, so a throwing `fill` leaves a discoverable half and the retry double-creates | `InMemorySessionRepo.create` is keyed by id and overwrites (count stays 1), and invoke holds the session lease before reaching the store, so no concurrent observer exists |

kid7st:
Review response — the finding is fixed in 709dccf, and a second problem surfaced while fixing it.

**The documented port now matches the real one.** `openOrCreate` carries `inherit?`, `SessionInheritance` is exported from `/pi` (a published port whose parameter type is internal is only half published), and `dir`'s resolution rule is stated: relative means *inside the workspace this store serves*, not inside the process directory, with `cwd` also scoping lookups.

**Unrelated to the review, found by widening the diff:** two session journals from a local mutation run had been committed on this branch and pushed, inside `.state/`. They entered through `git add -A` and survived two reviews of this branch because both diffed `-- src/ docs/` instead of the whole tree. Removed, and `.state` is now ignored — it is where an agent keeps machine state by definition. Squash-merging means `main` never sees them either way, but the branch should not carry them.

kid7st:
Correction to this PR's own description, found by checking a claim I had only reasoned about.

The body said this change closes **two** pre-existing defects. Only one of them is real. `main` derives the publish target from the parent session's metadata path — already resolved by pi — so a relative `dir` works there today; I ran it to be sure (`inherited=true discoverable=true underWorkspace=true`). The path-base work is not a repair of something broken on `main`: `createAtomically` has to create parentless sessions too, and a parentless draft has no parent path to borrow, so the refactor needs a defined root of its own. It is a consequence of the change, handled inside it, not a defect it fixes.

`encodeSessionId` stands as written — a real defect, flagged during #328's review and deferred.

Body updated. The code is unchanged.

kid7st (commented):
Thesis: session creation collapses from a three-verb draft protocol callers must sequence into one `createAtomically` primitive whose contract is "`id` resolves to a complete session or to nothing", plus the path base and injective id encoding that primitive now requires.

5 inline findings, all on `src/engines/pi/sessions.ts`.

src/engines/pi/sessions.ts
```diff
@@ -485,7 +513,23 @@ export function jsonlSessionStore(options: {
   };
 }
 
-/** Injective filename-safe encoding: [A-Za-z0-9._-] verbatim, the rest %-escaped. */
+/**
+ * Filename-safe encoding, INJECTIVE: `[A-Za-z0-9._-]` verbatim, everything else `%XX` (one byte) or
+ * `%uXXXX` (above it). Two different ids must never name one file — they would share a conversation,
```
kid7st: The doc claims a filesystem property the function cannot provide. Injectivity of the string function is not "two different ids never name one file": `[A-Za-z0-9._-]` passes through verbatim, and the default filesystem on macOS (APFS) and Windows is case-insensitive, so `"Room"` and `"room"` are two distinct ids naming one file — exactly the failure named here ("a thread inheriting from the wrong room"), reachable through a custom `route()`.

Direction: fold case into the escaping (lowercase-only alphabet, uppercase escaped), or state that the guarantee holds only on case-sensitive filesystems.

src/engines/pi/sessions.ts
```diff
@@ -320,91 +320,109 @@ async function markInheritanceWindow(child: Session): Promise<void> {
...
 }
 
-/** The backend surface inheritance needs — two repos, one decision table. */
-interface InheritRepo<M> {
+/**
+ * The backend surface session creation needs. Its one contract: **`id` resolves to a complete
+ * session or to nothing, never to a half** — `openOrCreate` short-circuits on EXISTENCE, so a
+ * newborn discovered before its entries, repair and window mark are written reads as "the decision
+ * was taken" forever. pi's own `create`/`fork` publish first and append after, which is why this
+ * seam exists. Retrying is therefore always safe.
```
kid7st: "Retrying is therefore always safe" is stated unconditionally but only holds for SEQUENTIAL retry. Two concurrent `openOrCreate` calls for the same unknown id still publish two files with the same id (filenames are timestamped, so `rename` cannot collide); afterwards `find(m => m.id === id)` picks one by list order and the other's turns are written to the loser. Nothing in this store serializes creation — it relies on the channel turn-queues.

Direction: scope the claim to "retrying after a failure is safe", or name the serialization the invariant depends on.

src/engines/pi/sessions.ts
```diff
@@ -430,30 +448,40 @@ export function jsonlSessionStore(options: {
...
+  // scans only `<root>/<encodedCwd>`, and a cwd-less `list()` scans `<root>/*/​*.jsonl`, one level,
+  // which `.inherit-tmp/<encodedCwd>/*.jsonl` sits below. This is what makes creation atomic on a
+  // backend that has no transactions: everything before the rename is invisible and repeatable,
+  // everything after it is complete. A crash mid-creation therefore leaves only invisible garbage
+  // (bounded by crash count; the next attempt starts a fresh draft — never resumed, so staleness
+  // cannot poison anything).
   const draftRepo = new JsonlSessionRepo({
     fs: new NodeExecutionEnv({ cwd }),
-    sessionsRoot: join(options.dir, ".inherit-tmp"),
+    sessionsRoot: join(root, ".inherit-tmp"),
```
kid7st: `.inherit-tmp` no longer names what it holds. After this refactor every creation stages here — parentless drafts and the empty fallback included (`startEmpty` → `createAtomically(id, undefined, …)`). The comment above already calls it "the draft realm"; the directory name still says inheritance, and the test filters on that literal.

Direction: rename to `.drafts`. Drafts are never resumed, so the on-disk name change costs nothing.

src/engines/pi/sessions.ts
```diff
@@ -430,30 +448,40 @@ export function jsonlSessionStore(options: {
...
+  // Resolve ONCE, by pi's rule (`NodeExecutionEnv.absolutePath` is `resolve(cwd, path)`): the direct
+  // fs calls below resolve against process.cwd() instead, and a relative `dir` would straddle both.
+  const root = resolve(cwd, options.dir);
+  const repo = new JsonlSessionRepo({ fs: new NodeExecutionEnv({ cwd }), sessionsRoot: root });
+  // The draft realm: a sibling root INSIDE the store root but OUTSIDE every lookup — `list({ cwd })`
+  // scans only `<root>/<encodedCwd>`, and a cwd-less `list()` scans `<root>/*/​*.jsonl`, one level,
+  // which `.inherit-tmp/<encodedCwd>/*.jsonl` sits below. This is what makes creation atomic on a
+  // backend that has no transactions: everything before the rename is invisible and repeatable,
+  // everything after it is complete. A crash mid-creation therefore leaves only invisible garbage
+  // (bounded by crash count; the next attempt starts a fresh draft — never resumed, so staleness
```
kid7st: The garbage bound stated here is narrower than the code's. A HANDLED failure leaks a draft too: `draftRepo.fork` can create the draft file and then throw on a torn parent, and a throw from `fill` or from `rename` unwinds into `createInheriting`'s catch, which starts empty and never unlinks. That path repeats without a process restart, so "bounded by crash count" is not the real bound.

Direction: say "a crash or a failed attempt", or unlink the draft on the handled failure path.

src/engines/pi/sessions.ts
```diff
@@ -320,91 +320,109 @@ async function markInheritanceWindow(child: Session): Promise<void> {
...
     }
-    const draft = await repo.forkDraft(parentMeta, id, at ?? leaf.id);
-    await reconcileInterruptedToolCalls(draft);
-    await markInheritanceWindow(draft);
-    return await repo.publish(draft, parentMeta);
+    log.debug(`[fastagent] parent session "${parentId}" is empty — nothing to inherit`);
   } catch (error) {
-    // A failure ANYWHERE above leaves no discoverable child (drafts live outside the store's
-    // lookups), so falling back to create cannot produce a second file under the same id.
     log.warn(`[fastagent] could not fork parent session "${parentId}" (${String(error)}) — starting empty`);
```
kid7st: The `try` now spans `createAtomically`, so `rename` failing (EACCES/ENOSPC) or `published session vanished` degrades to an empty session under the message `could not fork parent session "X"`. That attributes a store-level fault to the parent's data — different owner, different fix.

Direction: keep the parent-read work in the `try` and let publish-level failures surface as themselves, or word the warn so the wrapped error is not pre-diagnosed.

src/engines/pi/sessions.ts
kid7st: Refuted as a behaviour, accepted as wording — fixed in cd7c1ae.

Measured: two ids differing only in case produce two files and cannot see each other's content (files=2, lowerSeesUpper=false), on this machine's case-folding APFS. A session file is `<timestamp>_<id>.jsonl`, so distinct sessions never share a name, and identity is resolved by the `id` a session's metadata carries — a case-sensitive string compare in `find`, not a filesystem lookup.

What was true is that the comment said 'two different ids must never NAME ONE FILE', which invited exactly this reading. It now states the guarantee it actually makes: two ids must never encode alike, because the encoded id is what openOrCreate matches on.

src/engines/pi/sessions.ts
kid7st: Accepted as an incomplete claim; fixed in cd7c1ae. The sentence now says re-attempting AFTER A FAILURE, which is the sequential thing it was about, and names what answers the other question: the single-writer lease, taken in invoke before anything reaches a store.

One correction to the finding: I could not reproduce the duplicate. Two concurrent openOrCreate calls for one unknown id gave files=1 twice — same millisecond means the same filename and the second create overwrites; staggering them makes the second find the first. The mechanism is plausible from the code (both list() completing before either create()), but it is undemonstrated, and the lease excludes it on the serving path. The in-process lease does leave two PROCESSES sharing a sessions directory unprotected, which is the multi-instance tier the SPEC already places outside this store.

src/engines/pi/sessions.ts
kid7st: Accepted, renamed to `.drafts` in cd7c1ae, comments and test literals with it. Every creation stages there now, so a name about inheritance described only the case it was introduced for.

src/engines/pi/sessions.ts
kid7st: Conclusion accepted, mechanism corrected; fixed in cd7c1ae. Measured: a torn parent leaks nothing (draftsLeaked=0) — pi's fork opens and reads the source before it creates the target, so a read failure leaves no file. What does leak is a throw from `fill` or from the rename, after the draft exists, and those are handled failures that repeat without a restart. So the bound was indeed wrong; the comment now says a crash OR a handled failure, and states that nothing unlinks them.

Not doing the unlink: it would add a cleanup path for residue that only appears on write failures and is already inert (drafts are never resumed).

src/engines/pi/sessions.ts
kid7st: Accepted; the warn no longer pre-diagnoses. One correction: this is not something the refactor introduced — `main` also has `publish` inside the same `try` under the same message, so a rename failure has always been reported as a bad parent. The behaviour stays (degrading to an empty session keeps the turn alive, and context is not the ask); only the attribution changes, since the block spans reading the parent AND writing the child.

kid7st:
One gap the review's mutation pass exposed, worth calling out separately because it is about this branch's headline property rather than any single line.

**The atomicity had no test.** Staging drafts in the real directory instead of the draft realm — that is, removing the mechanism this PR exists for — passed all nineteen inheritance tests. Publish's rename then targets the path the file already has, so it is a no-op, the end state is identical, and only the crash window differs. Nothing observed it.

The first assertion I wrote for this was worse than none: it went into the test that plants garbage by `mkdirSync`-ing `.drafts` itself, so it asserted its own setup and passed under the mutation too. It now lives in a test that never touches that directory, and it fails when the staging is removed.

What it buys, stated honestly: it stops the mechanism from being removed silently. It is not proof of crash safety — that needs fault injection between draft creation and publish, which would mean a seam in production code that exists only for the test. Still declining that one.

kid7st (commented):
Second pass, over `cd7c1ae`. Thesis unchanged: one `createAtomically` primitive whose contract is "`id` resolves to a complete session or to nothing".

Withdrawing my previous `encodeSessionId` case-collision finding — filenames carry the timestamp and identity is matched on metadata `id`, so a case-folding filesystem does not merge two ids. The corrected doc states this.

2 inline findings.

src/engines/pi/sessions.ts
```diff
@@ -430,30 +452,43 @@ export function jsonlSessionStore(options: {
...
-  // exactly one atomic rename, already repaired and window-marked.
+  // Resolve ONCE, by pi's rule (`NodeExecutionEnv.absolutePath` is `resolve(cwd, path)`): the direct
+  // fs calls below resolve against process.cwd() instead, and a relative `dir` would straddle both.
+  const root = resolve(cwd, options.dir);
+  const repo = new JsonlSessionRepo({ fs: new NodeExecutionEnv({ cwd }), sessionsRoot: root });
+  // The draft realm: a sibling root INSIDE the store root but OUTSIDE every lookup — `list({ cwd })`
+  // scans only `<root>/<encodedCwd>`, and a cwd-less `list()` scans `<root>/*/​*.jsonl`, one level,
+  // which `.drafts/<encodedCwd>/*.jsonl` sits below. This is what makes creation atomic on a backend
+  // that has no transactions: everything before the rename is invisible and repeatable, everything
+  // after it is complete. EVERY creation stages here, forked or empty — hence `.drafts`, not a name
```
kid7st: "EVERY creation stages here, forked or empty" is false, and it is the stated justification for the `.drafts` name. `openOrCreate` without `inherit` calls `repo.create({ id, cwd })` directly, in both backends — the ordinary new session (every channel's first session with no parent, i.e. the common case) is never staged and never reaches `createAtomically`. What actually stages is "creation reached through `createInheriting`", including its empty fallback.

Direction: say that, or route the no-inherit create through `backend.createAtomically(id, undefined, …)` so the sentence becomes true and there is one creation path instead of two.

test/session-inherit.test.ts
```diff
@@ -45,13 +45,18 @@ async function buildParent(store: PiSessionStore, id: string, n: number): Promis
...
 
     const child = await store.openOrCreate("thread", { parentSession: "room" });
     const seen = await contextText(child);
     expect(seen).toContain("q1 ask");
     expect(seen).toContain("a2 answer");
+    // Staged outside the store's lookups, then published — not written in place. Without this, moving
+    // the draft realm INTO the real directory passes every other test: publish's rename becomes a
+    // no-op on the same path, so only the crash window differs and nothing else observes it. A
+    // structural check that the mechanism ran, not proof of crash safety.
+    expect(existsSync(join(dir, ".drafts"))).toBe(true);
```
kid7st: The branch's headline property is asserted inside a test whose name is about something else: `it("a new session naming a parent forks it once, then lives independently")` now also carries the only check that publishing goes through the draft realm. The name does not spec it, so the protection is invisible when reading the suite — and the next person tidying this test deletes the one assertion the refactor rests on.

Direction: give it its own `it("a newborn is staged outside the store's lookups, then published")` — `buildParent` + one `openOrCreate` + the `.drafts` check.

src/engines/pi/sessions.ts
kid7st: Accepted, and confirmed before fixing: after a plain `openOrCreate` the draft realm does not exist; after an inheriting one it does (afterPlainCreate=false, afterInheritFallback=true). The sentence was false, and it came from the previous round's finding, which I took at its word instead of running it — the same mistake this review keeps catching.

Fixed in b295f3e. The comment now says what stages (creation reached through createInheriting: the fork, and that path's empty fallback) and why the other path needs nothing: an empty session is complete the moment its file exists, so no reader can catch it half-made.

Declining the second direction. Routing plain creates through createAtomically would make the sentence true by adding a rename to every ordinary session — the common case — to protect a window that does not exist there. Two paths is the right shape; what was missing is saying so.

The name stays `.drafts`, on an argument that does not depend on the false one: it names the mechanism rather than today's only caller, so a second caller staging there would not turn it into a lie. Say the word if you would rather have `.inherit-tmp` back.

test/session-inherit.test.ts
kid7st: Accepted, fixed in b295f3e — `it("a newborn is staged outside the store's lookups, then published")`, with buildParent + one openOrCreate + the realm check, and the reasoning about what it does and does not prove kept with it. Mutation-verified in its new home: staging in the real directory fails it and nothing else.

The diagnosis is worth keeping: I had put it in the fork-once test because that was where a store with a parent already existed, which is a convenience, not a spec. A test name is the spec, and a guard outside its spec is one tidy-up away from gone.

PR #338, opened 2026-08-18
feat(pi): prove the SPEC MUSTs on pi's own AgentSession class

kid7st:
Phase 1 of moving the serving path off `AgentHarness`. This is a **proof, not a migration**: it adds a second L0 next to `createPiAgentFromHarness` and runs the existing SPEC conformance suite against it. Nothing serves on it yet.

## Why

pi 0.84 promoted an unimplemented lane-based `AgentHarness` to pi-agent-core's default export and removed the working one (`prompt`/`compact`/`abort`/`events.on` all throw `HarnessNotImplemented`; see #337, which holds us at 0.83). The deeper problem it exposed: **pi does not consume `AgentHarness` itself** — its TUI, RPC and SDK all run on `AgentSession`. We were the only consumer of a surface with no dogfooding pressure behind it.

`docs/design/conformance-levels.md` §2 already named the destination — `per-invoke × session`, the "portable AND extension-hosting" cell. This PR makes it executable.

## What it proves

All four Agent-side MUSTs hold on `AgentSession` in the `per-invoke` posture, MUST 6 (portability) included — every turn builds a fresh session over the same jsonl, with nothing shared in-process but the directory:

```
✓ MUST 1 single terminal                     25ms
✓ MUST 1+2 failures are events               10ms
✓ MUST 3 cancel + cleanup                    10ms
✓ §5 JSON-serializable                       17ms
✓ MUST 6 portable across instances           22ms
```

Measured behaviour behind the implementation (probed before writing it):

| Question | Answer |
|---|---|
| Event stream | `agent_start → turn_start → message_start/update/end → turn_end → agent_end → agent_settled`; `message_update` / `tool_execution_*` fields are identical to the harness ones |
| Does `prompt()` throw? | No — a provider error and an abort both resolve normally, so the terminal comes from the last assistant message's `stopReason`, exactly as on the harness path |
| Abort | `stopReason: "aborted"`, content truncated at the interruption point |
| Cross-instance history | `SessionManager.create(cwd, dir, {id})` / `list` / `open(path)` is sufficient |

Per-turn cost on a 20-turn session (directory scan + jsonl read included): services build once at 16ms, then **p50 0.59ms / p95 2.08ms** to bind a session — confirming the "~1 ms" figure in conformance-levels.md §2.

One free upgrade: `AgentSession` retries a failed assistant request itself (`auto_retry_start`), which the harness never did. It maps onto SPEC's existing `retrying` event. Whether serving keeps it enabled is a phase-2 policy call — the failure subject turns it off so its ~14s backoff does not pace the suite.

## Scope

Wired: the concurrency floor, the event stream, cancellation.

Not wired: observation plane (`SessionObserver` / `RunControls` / rich `SessionEvent`), tool activation bridge, auto-compaction, session inheritance. The fastagent assembly (definition, tools, prompt) is not connected either — the test supplies a bare `services`.

So `createPiAgentFromHarness` remains the only serving path, and the new module is deliberately absent from `src/pi.ts`.

Known debt, stated in the module header with its retirement condition: `toAgentEvent` translates pi events straight to SPEC `AgentEvent`s, which is the second parallel translation `docs/design/session-control.md` §6 forbids. It goes away when this L0 grows the observation plane — which is also when the harness L0 goes away.

## Verification

`npm run lint && npm run typecheck && npm test` — 106 files, 1368 tests, all green.

### DIFF
--- a/src/engines/pi/invoke-session.ts
+++ b/src/engines/pi/invoke-session.ts
@@ -10,10 +10,14 @@
  *
  * SCOPE, deliberately narrow: the concurrency floor, the event stream, and cancellation. The
  * observation plane (SessionObserver / RunControls / the rich `SessionEvent` vocabulary), the tool
- * activation bridge and auto-compaction are NOT wired — this translates pi events straight to SPEC
- * `AgentEvent`s, which is the second parallel translation design §6 forbids. That is the price of a
- * throwaway proof; the harness L0 remains the one true path until this one grows the observation
- * plane and replaces it.
+ * activation bridge, auto-compaction and session inheritance are NOT wired.
+ *
+ * WHICH L0 SERVES: {@link createPiAgentFromHarness}, still — this one is reachable only from its
+ * conformance test (deliberately absent from `src/pi.ts`), because a serving path needs the pieces
+ * above. Its one known debt is {@link toAgentEvent}: translating pi events straight to SPEC
+ * `AgentEvent`s is the second parallel translation `docs/design/session-control.md` §6 forbids. It
+ * retires the moment this L0 grows the observation plane — the rich `SessionEvent` layer comes back
+ * with it, and the harness L0 goes away.
  */
 import type { AgentSession, AgentSessionEvent } from "@earendil-works/pi-coding-agent";
 import type { AssistantMessage } from "@earendil-works/pi-ai";
@@ -71,7 +75,16 @@ function terminalFromState(session: AgentSession): AgentEvent {
     const message = messages[i];
     if (message?.role === "assistant") return toTerminal(message as AssistantMessage);
   }
-  return { type: "failed", details: "the turn produced no assistant message", retryable: false };
+  // Unreachable on a settled run: pi records an assistant message for every outcome, error and abort
+  // included. Reaching it means the engine broke its own contract — say so with what the state held,
+  // so the report names the engine rather than the turn.
+  return {
+    type: "failed",
+    details:
+      `the engine settled the run without an assistant message ` +
+      `(${messages.length} message(s), last role: ${messages.at(-1)?.role ?? "none"})`,
+    retryable: false,
+  };
 }
 
 export function createPiAgentFromSession(options: CreatePiAgentFromSessionOptions): Agent {
--- a/src/engines/pi/invoke-session.ts
+++ b/src/engines/pi/invoke-session.ts
@@ -42,9 +42,11 @@ export interface CreatePiAgentFromSessionOptions {
 function toAgentEvent(event: AgentSessionEvent): AgentEvent | null {
   switch (event.type) {
     case "message_update": {
+      // An empty delta is not output: it moves no consumer's state, and treating it as output would
+      // spend the silent window that auto-retry is allowed to use (see runOnSession).
       const ev = event.assistantMessageEvent;
-      if (ev.type === "text_delta") return { type: "text", delta: ev.delta };
-      if (ev.type === "thinking_delta") return { type: "thinking", delta: ev.delta };
+      if (ev.type === "text_delta") return ev.delta === "" ? null : { type: "text", delta: ev.delta };
+      if (ev.type === "thinking_delta") return ev.delta === "" ? null : { type: "thinking", delta: ev.delta };
       return null;
     }
     case "tool_execution_start":
@@ -66,27 +68,22 @@ function toAgentEvent(event: AgentSessionEvent): AgentEvent | null {
 
 /**
  * The turn's outcome. `prompt()` resolves void and never throws for an engine-side failure (measured:
- * a provider error and an abort both resolve normally), so the terminal comes from the last assistant
- * message's `stopReason` — the same read {@link toTerminal} performs on the harness path.
+ * a provider error and an abort both resolve normally), so the terminal comes from the assistant
+ * message the run ended on — the same read {@link toTerminal} performs on the harness path.
  *
- * Bounded to THIS turn (`from` = the message count before `prompt()`): the session is durable and
- * carries every previous turn, so an unbounded reverse scan would answer a run that produced nothing
- * with the PREVIOUS turn's assistant message — reporting `completed` for a turn that never ran.
+ * That message is taken from the EVENT STREAM, not from `session.state.messages`. Session state is
+ * mutable mid-turn: compaction replaces the array, and overflow recovery splices the last assistant
+ * message out of it outright (`state.messages = messages.slice(0, -1)`). Any index into it is a
+ * turn boundary that the engine is free to invalidate, while a `message_end` payload is a fact that
+ * already happened. Auto-compaction runs its own model call outside the agent's event stream, so it
+ * cannot masquerade as the turn's answer here.
  */
-function turnTerminal(session: AgentSession, from: number): AgentEvent {
-  const messages = session.state.messages;
-  for (let i = messages.length - 1; i >= from; i--) {
-    const message = messages[i];
-    if (message?.role === "assistant") return toTerminal(message as AssistantMessage);
-  }
-  // Unreachable on a settled run: pi records an assistant message for every outcome, error and abort
-  // included. Reaching it means the engine broke its own contract — say so with what the state held,
-  // so the report names the engine rather than the turn.
+function engineProducedNothing(): AgentEvent {
+  // Unreachable on a settled run: pi ends every outcome, error and abort included, with an assistant
+  // message. Reaching it means the engine broke its own contract — name the engine, not the turn.
   return {
     type: "failed",
-    details:
-      `the engine settled the run without an assistant message ` +
-      `(${messages.length - from} message(s) this turn, last role: ${messages.at(-1)?.role ?? "none"})`,
+    details: "the engine settled the run without ending an assistant message",
     retryable: false,
   };
 }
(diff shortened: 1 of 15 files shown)


kid7st:
Follow-up review with `gpt-5.6-sol` found a real defect the earlier pass missed, now fixed in 49fc721.

`terminalFromState()` scanned the whole transcript backwards for the run's assistant message. The session is durable and carries every previous turn, so a run that settled without producing anything answered with the **previous** turn's assistant message — reporting `completed` for a turn that never ran (SPEC MUST 1). The harness L0 never had this shape: `prompt()` returns the message there, so the read could not reach history.

The scan now starts at the message count captured before `prompt()`. Covered by `test/invoke-session.test.ts` with a stub session — a healthy engine always records an assistant message, so the case is unreachable through a real one (verified the test fails on the previous commit with `expected 'completed' to be 'failed'`).

kid7st:
Two more defects from a follow-up review pass with `gpt-5.6-sol`, both fixed here:

- **4c8acab** — the cancel latch was read right after the session was built, leaving prompt-option resolution (image resizing) outside the guard. A consumer cancelling in that window knocked on an armed door with an idle session behind it, and the turn started anyway: a model call for a caller that already walked away (MUST 3). One gate now sits after the last await before `session.prompt()`. The regression test holds option resolution open so the cancellation lands inside it.
- **0ff216c** — a throw while preparing the prompt escaped the generator and broke iteration (MUST 2). It is a `failed` event now.

The same MUST 2 defect exists on the harness L0 and is fixed separately in #339, since that is the serving path.

`rv.sh local` is clean on both branches.

kid7st:
## Clean-up pass (structure, not behaviour)

Reviewed the branch for structure and did three things:

**1. The dependency direction was backwards** — `invoke-session.ts` was importing the harness L0's internals (lease, terminals, event queue, prompt prep), and two comments existed only to explain why those became exported. None of those parts care which pi class runs the turn, so they move to `turn-kit.ts`, which both L0s sit on. `invoke.ts` drops 721 → 510 lines and its header now says what it is.

**2. The cancellation protocol was copied per L0** — `cancellableStream` joins `abortFirstIterator` in `collect.ts`, so a turn generator receives the latch and publishes its door through one wrapper instead of 20 lines of identical plumbing in each L0.

**3. Naming** — `turn-kit` is *pi-class*-neutral, not `engine-neutral`. This repo reserves the latter for code with no engine import at all (`src/agent.ts`); `turn-kit` speaks pi's message and image types. Fixed in the module header, `AGENTS.md` and `core.md`.

## One more defect, fixed at the protocol boundary

Review found that a generator parked in an await can still reach a `yield` on its way out — an error path yielding a terminal — and that yield satisfies the pending `next()`, handing a terminal to a consumer that already walked away (MUST 3 says a cancelled turn has no terminal).

The reflex fix is a latch check before that `yield`. That would have been the third such check across two L0s, so instead `abortFirstIterator` answers it once: after teardown starts, `next()` resolves `done` regardless of what the generator emitted. Both teardown paths also stop being near-copies. Covered by a race test that fails without the change.

## Merge order

#339 fixes the same class of defect on the harness path. Merging **#339 first** is cleaner; verified locally that the two merge without conflict and the suite stays green (1374 tests). After both land I will add the harness-side regression test for the cancel window — `toPiPromptOptions` is module-local today, and only this PR's `turn-kit` extraction makes it mockable.

`rv.sh local` on this branch is clean apart from the harness-path finding that #339 already carries.

kid7st:
Rebased on `main` now that #339 has landed, and closed the loop it left open.

#339 moved the harness L0's cancel latch after prompt preparation but could not test it — `toPiPromptOptions` was module-local, so nothing could hold that window open. This branch's `turn-kit` extraction makes it reachable, so the harness path now has the same regression test as the session path: park the generator inside preparation, cancel, assert the model call never starts. Verified it fails with the latch read back at its old position.

Both parked-window tests also stopped counting microtasks — they wait for a signal raised from inside the window instead. The harness factory has several awaits before preparation, so `await Promise.resolve()` was enough for the session path and silently wrong for this one.

`rv.sh local`: **No findings.** 1375 tests green.

src/engines/pi/invoke-session.ts
```diff
@@ -0,0 +1,177 @@
...
+      options = await toPiPromptOptions(prompt);
+    } catch (error) {
+      yield errorToTerminal(error);
+      return;
+    }
+    if (wasCancelled()) {
+      await abort();
+      return;
+    }
+    const before = session.state.messages.length;
```
kid7st: Problem: `turnTerminal` uses the pre-prompt array length as a turn boundary, but `AgentSession.prompt()` can auto-compact before or after the model call and replace `agent.state.messages` with a shorter/reindexed array. After such a compaction, a successful current-turn assistant message can sit below `from`, so this L0 reports `failed: ...without an assistant message` for a successful turn.

Direction: Capture the current turn’s final assistant `message_end` from the subscribed event stream and derive the terminal from that stable event. Do not use an index into mutable session state as the turn identity.

src/engines/pi/invoke-session.ts
```diff
@@ -0,0 +1,177 @@
...
+
+/**
+ * SPEC events from pi's session stream. `auto_retry_start` has no harness counterpart: an
+ * `AgentSession` retries a failed assistant request itself, which the SPEC already has a word for.
+ */
+function toAgentEvent(event: AgentSessionEvent): AgentEvent | null {
+  switch (event.type) {
+    case "message_update": {
+      const ev = event.assistantMessageEvent;
+      if (ev.type === "text_delta") return { type: "text", delta: ev.delta };
```
kid7st: Problem: The L0 forwards `text_delta` immediately while leaving `AgentSession` auto-retry enabled. A retryable attempt may stream partial text, then `auto_retry_start` removes that failed message and a later attempt streams the replacement; SPEC consumers such as `collect` only append deltas, so they return the failed attempt’s partial text concatenated with the successful answer. The `retrying` event cannot retract already-emitted output, and the conformance failure subject disables auto-retry, so the tests do not exercise this behavior.

Direction: Disable `AgentSession` auto-retry for this L0 until the AgentEvent protocol can represent replacement/reset, or buffer attempt output and emit only the attempt that becomes the final answer.

kid7st (commented):
This change adds a per-invoke AgentSession L0 as an executable SPEC-conformance proof, while extracting the pi-class-neutral turn mechanism shared with the existing harness L0.

src/engines/pi/invoke-session.ts
kid7st: Confirmed and fixed in 793e9c4. The engine rewrites that array in three places, not just compaction: `_prepareRetry` and overflow recovery both run `state.messages = messages.slice(0, -1)`, so the boundary index can be invalidated even without a compaction. The terminal now comes from the `message_end` payload. Auto-compaction cannot forge it — it runs its own model call outside the agent event stream (verified: only `compaction_start`/`compaction_end` are emitted around it).

src/engines/pi/invoke-session.ts
kid7st: Fixed in 793e9c4, keeping the resilience where it is free. The retry is refused only once output has been emitted; a failure in the silent window still retries (both directions are now tested).

Two things surfaced while fixing it:

- Refusing was not enough — it had to be **cancelled**. pi emits `auto_retry_start` *before* creating the controller that makes its backoff abortable, so an abort from inside the listener is a no-op and the turn still pays 2s and burns a provider call on an answer it must discard. The cancel is deferred one microtask; the test asserts the failure returns in under 1s.
- Your suggested alternative, `setAutoRetryEnabled(false)`, turned out to be unusable: it persists to the machine-global `~/.pi/agent/settings.json`. My earlier conformance subject called it and silently disabled auto-retry for this machine's pi TUI. Reverted, and nothing in src or test writes user settings now.

src/engines/pi/invoke-session.ts
```diff
@@ -0,0 +1,200 @@
...
+      retriedAfterOutput = event.errorMessage;
+      // Not synchronously: pi emits this event BEFORE creating the controller that makes its backoff
+      // abortable, so an abort from inside the listener would find nothing to cancel and the turn
+      // would still pay the full delay and burn a provider call on an answer we must discard.
+      queueMicrotask(() => void abort());
+      return;
+    }
+    const projected = toAgentEvent(event);
+    if (!projected) return;
+    if (projected.type !== "retrying") emittedOutput = true;
```
kid7st: Problem: `emittedOutput` is set by every projected event except `retrying`, so a completed `tool_started` / `tool_ended` pair also closes the internal-retry window. If the model request after a successful tool call fails transiently before emitting answer text, `AgentSession` can safely retry that model request using the persisted tool result; this L0 aborts it instead and returns `failed{retryable:true}`. A caller retry can then rerun the whole prompt and repeat the tool side effect—the opposite of the guard’s stated purpose, which is only to prevent concatenating a failed attempt’s answer deltas.

Direction: Track only replaceable assistant-attempt deltas as the retry blocker, not tool lifecycle events. Add a tool-call → silent transient model failure → successful internal retry test that proves the tool executes once and the replacement answer completes.

kid7st (commented):
This change adds a per-invoke AgentSession L0 as an executable SPEC-conformance proof and extracts the pi-class-neutral turn mechanism shared with the harness L0.

src/engines/pi/invoke-session.ts
kid7st: Correct, and the consequence was worse than the problem the guard exists for. Fixed in fa8cb7c.

The flag now flips only on `text` / `thinking` deltas — the only output a retry can duplicate, because only those were already streamed as deltas the SPEC cannot retract. pi's retry discards the failed assistant message and resumes from the persisted tool result, so tool lifecycle events are never replayed by it.

Added the test you described: tool call → silent transient failure → internal retry succeeds. It asserts `toolRuns === 1` and that the replacement answer completes. Verified it fails on the previous commit (`expected +0 to be 1`).

One incidental find while writing it: the subject was built with `noTools: "all"`, which drops custom tools too, so the tool never ran at all. Now `"builtin"` — the coding tools stay off, a subject's own tool stays mounted.

PR #344, opened 2026-08-20
chore(deps): pi 0.84 — the hold is over

kid7st:
The hold on pi is over: the serving path no longer runs on `AgentHarness`, so the reason to pin 0.83 is gone.

## What the upgrade cost

**Four type errors, all renames** — down from the ~40 that made the original bump a migration rather than a bump:

| | |
|---|---|
| `SessionTreeEntry` | moved to pi-coding-agent as `SessionEntry`. That is the package we were describing anyway: the tool runtime and the control plane read a v3 jsonl entry, not an agent-core one. |
| `AuthInteraction` | became `ProviderAuthInteraction`, with `signal` required. A caller that passes none now gets one that never fires — the same "no cancellation" the optional field meant. |

pi's entry type is exported as **`PiSessionEntry`**, not `SessionEntry`: this package already publishes a `SessionEntry`, the neutral one the control plane answers `entries()` with. Two shapes, two names. Likewise `SessionManager` is exported under its own name rather than kept as `Session` — an old concept's name over a new concept is worse than either.

## Verified

`lint + typecheck + 1346 tests` on 0.84.2, and end to end against a real provider — a definition served over HTTP, a telegram-shaped session id, and a thread inheriting from it:

```
turn 1  session "-1001234567890"                    → "noted" (remembers 58)
turn 2  session "thread-x", parentSession "-100…"   → "58"
```

## Dependabot

The `>=0.84` ignore is removed. Its stated release condition was "when the harness L0 is gone" — #341 removed it.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -9,19 +9,6 @@ updates:
       npm-dependencies:
         patterns:
           - "*"
-    ignore:
-      # pi 0.84 promoted the unfinished lane-based AgentHarness to pi-agent-core's default export
-      # and removed the working one: prompt/compact/abort/events.on all throw HarnessNotImplemented.
-      # The serving path (src/engines/pi/invoke.ts) runs on that class, so it cannot move yet.
-      #
-      # The condition is OURS, not upstream's. pi does not consume AgentHarness itself - its TUI,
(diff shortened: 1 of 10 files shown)


kid7st (commented):
No findings.

PR #345, opened 2026-08-20
feat(pi): the definition carries its own extensions/ (chat runs them; serving says why it cannot)

### LINKED ISSUE #285: Definition should carry its own extensions, the way it carries skills

kid7st:
Closes #285.

Load `<agentDir>/extensions/` as part of the definition and run them in **`fastagent chat`**.
Serving discovers them, warns, and runs without them — because pi's extension runtime cannot
currently be shared safely by concurrent sessions. That split is the substance of this PR.

## The split the flag was really making

`noExtensions` protects reproducibility — a served agent must not inherit the authoring machine's
`~/.pi` extensions and slash commands. That is worth keeping. But the flag was suppressing two
different things at once:

| | before | after |
|---|---|---|
| machine-global `~/.pi/extensions` | suppressed | suppressed |
| the definition's own `extensions/` | suppressed (collateral) | **loaded in `chat`** |

pi honours `additionalExtensionPaths` even under `noExtensions`, so `chat` gets both properties at
once without a host-kind branch or a "local mode".

## Why serving does not run them

This is what most of the review on this PR was about, and it took six rounds to see whole.

pi's extension machinery is built for **one process serving one session** — a terminal. Its own
source calls the object holding a session's actions "the shared runtime", and every `AgentSession`
overwrites it on construction:

```js
// Copy actions into the shared runtime (all extension APIs reference this)
this.runtime.sendMessage = actions.sendMessage;
this.runtime.appendEntry = actions.appendEntry;
```

Serving is the opposite shape: one process, concurrent turns, conversations with nothing to do with
each other. With two turns in flight, the second to start redirects those actions to itself — so an
extension calling `pi.sendMessage()` during the first turn can deliver **into someone else's
conversation**.

The same sharing shows up from other angles, and each one was found separately before the pattern
was clear:

| symptom found | what it actually was |
|---|---|
| a turn's `session_shutdown` cleared a timer a concurrent turn had just opened | one module instance, per-turn lifecycle events |
| reloading per turn "for isolation" — three concurrent turns still shared one instance | reload mutates shared state; it cannot isolate |
| building a loader per turn didn't help either | the module cache is process-global |
| editing `persona.md` rebuilt every extension | `reload()` starts with `clearExtensionCache()` |

Six defects, one cause: a single-session extension model wired into a multi-session server. Serving
therefore discovers, warns, and continues — a missing feature that says so beats a silent wrong
answer.

**This is narrowly fixable upstream.** pi already has an uncached loader path that builds a fresh
module per call (jiti with `moduleCache: false`) and takes the runtime as an argument — that *is*
per-session isolation. It is not exported, and the deep path is blocked by the package `exports`.
When that entry point exists, serving can run extensions with the guarantees `chat` has today. The
options type and `docs/configuration.md` both record what is needed.

## Not the shape the issue proposed

The issue suggested `AgentDefinition.extensions: Extension[]`, by analogy with `skills`. The analogy
breaks: a skill is **data** (content inline, serializable), an extension is **code** that pi loads
with jiti and binds to its own eventBus and runtime. The definition carries *paths*; pi does the
loading.

They are entry-point **files**, not the directory — handing pi a directory fails with
`Cannot find module` into a `LoadExtensionsResult.errors` entry nothing was reading. Discovery
follows pi's own rules (direct `*.ts`/`*.js`, or a subdirectory `index.*`); the `package.json` `pi`
manifest form is warned about rather than skipped in silence.

## Failures are read now, not dropped

- a load error (`LoadExtensionsResult.errors`) — pi collects and carries on, sound for a TUI that
  displays them, silent for a host that never looks;
- a throwing handler — pi fans errors to registered listeners and has none by default;
- a subdirectory in the unsupported manifest shape;
- **every symlink**, announced without guessing whether it meant to be an extension. A symlinked
  entry is refused (it resolves on the authoring machine and is missing in the container), and the
  message says how to silence it when the file was never an extension. Guessing by name looked
  tidier until `audit.ext -> some/dir` — a directory candidate to pi — got dropped in silence.

## Two things deliberately not done

**No `uiContext`.** pi defaults it to a no-op, so an extension calling `select`/`confirm`/`input`
gets an immediate "no answer" rather than hanging. (This also corrects the premise of #289.) Binding
a stand-in would mean implementing 20+ members of `ExtensionUIContext`, most of them TUI geometry.

**A throwing extension handler does not fail the turn.** It warns and the turn continues, which is
pi's own fan-out-and-continue semantics. Making it terminal would let one buggy handler take a chat
session down.

## What on-PR review changed

Review found two ways "chat runs them fully" was not yet true, and fixing them uncovered a third
bug nobody had reported:

- **An extension could not define the model.** `registerProvider()` is pi's documented way to add
  providers, but extensions do not execute until the services are built — and the model was resolved
  before that. A definition whose configured model came from its own extension failed with a bare
  `unknown model`, after warning about credentials for a provider that did not exist yet. The model
  is now resolved per session, after the services.
- **The `tools` allowlist froze the tool set at build time.** pi lets an extension register from
  `session_start`, a command, or any handler; those names cannot be in a startup snapshot, so
  `refreshTools()` filtered them straight back out. `noTools: "builtin"` gives the guarantee the
  allowlist was really there for — the machine's `defaultTools` setting cannot add pi's own copies
  on top of ours — without freezing anything.
- **(unreported) Chat was running pi's `read`, not ours.** Allow-listing the default tool *names*
  mounts pi's own implementations; fastagent's is `createReadTool({ imageProcessor })`. Chat had
  been running without image reading. Every fastagent tool is now passed as a `customTool`, exactly
  as serving does.

Two of my own fixes were wrong and are recorded here because the mistake is instructive:

- I added a `bindExtensions()` call, having measured that `session_start` never fired. It fires:
  `InteractiveMode.bindCurrentSessionExtensions()` does it. The measurement drove
  `buildAgentSessionRuntime` — the assembly, not the host — so it proved something true of the test's
  own call path and false of `fastagent chat`. The call would have emitted `session_start` twice per
  chat.
- I then stated the active tool set explicitly, on the theory that `noTools: "builtin"` left tools
  inactive. It does not; pi activates this agent's tools and leaves only its own extra helpers
  (`grep`, `find`, `ls`) inactive — the same set serving offers. Stating it explicitly handed the
  model all three. The existing session-builder assertion caught it.

Both properties now have tests that fail if the behaviour regresses.

## Validation

- [x] `npm run lint`, `npm run typecheck`, `npm test` (108 files, 1364 tests)
- [x] Rebased onto pi 0.84.2 + the single-engine `main`
- [x] `rv.sh local` — **No findings.** Ten rounds local plus one on-PR pass; each behavioural fix
      was falsified by reverting it and confirming the test goes red

## Checklist

- [x] Docs were updated (`docs/configuration.md`, `docs/overview.md`)

### DIFF
--- a/src/engines/pi/session-builder.ts
+++ b/src/engines/pi/session-builder.ts
@@ -323,14 +323,17 @@ export async function buildAgentSessionRuntime(
     // session.bindExtensions() with the TUI's uiContext, abort handler and command actions — binding
     // here too would emit session_start twice per chat, so an extension opening a resource on start
     // would open two.
-    // Deferral emulation: pi's session starts with everything active — narrow it by SUBTRACTING
-    // the deferred names from whatever is active (robust to pi mounting tools of its own; an
-    // exact-set-equality gate would silently stop narrowing the day pi adds one). Applied on EVERY
-    // build including /resume: pi's chat session does not record activations (its SessionContext has
-    // no activeToolNames), so "restore prior activations" is not implementable here — deferral stays
-    // consistently ON and a resumed conversation re-discovers via search_tools (documented divergence
-    // from serving, where activations persist in the session). The deferred SET comes from the shared
-    // assembly (one definition of "deferred"), never recomputed here.
+    // Deferral emulation: pi starts THIS agent's tools active — its own four default names, which
+    // our customTools have replaced, plus every custom and extension tool. (Tools pi mounts but does
+    // not activate, like grep/find/ls, stay inactive, which is the same set serving offers.) So
+    // narrow by SUBTRACTING the deferred names from whatever is active, rather than stating a set:
+    // an exact-set-equality gate would silently stop narrowing the day pi activates one more.
+    //
+    // Applied on EVERY build including /resume: pi's chat session does not record activations (its
+    // SessionContext has no activeToolNames), so "restore prior activations" is not implementable
+    // here — deferral stays consistently ON and a resumed conversation re-discovers via search_tools
+    // (a documented divergence from serving, where activations persist in the session). The deferred
+    // SET comes from the shared assembly (one definition of "deferred"), never recomputed here.
     if (deferredToolNames.length > 0) {
       const active = result.session.getActiveToolNames();
       if (deferredToolNames.some((n) => active.includes(n))) {
--- a/src/engines/pi/session-builder.ts
+++ b/src/engines/pi/session-builder.ts
@@ -43,7 +43,7 @@ import {
 } from "@earendil-works/pi-coding-agent";
 import { reportExtensionErrors } from "./agent-session-factory.ts";
 import { resolveModel } from "./config.ts";
-import { assembleSystemPrompt, piBasePrompt, piDefaultTools } from "./create.ts";
+import { assembleSystemPrompt, piBasePrompt } from "./create.ts";
 import { canonicalPath, loadAgentDefinition, loadExtensionPaths } from "./definition.ts";
 import { createPiModelRuntime, probeAuthSource } from "./models.ts";
 import { log } from "../../log.ts";
@@ -126,29 +126,16 @@ export async function buildAgentSessionRuntime(
     // agentDir carries the agent's own models.json (custom endpoints); stateRoot keeps pi's generated
     // catalog cache out of the definition dir. See createPiModelRuntime.
     const modelRuntime = await createPiModelRuntime({ authPath, agentDir, stateRoot });
-    // MIGRATION HINT (deliberate breaking change): chat historically used pi's own `~/.pi` auth;
-    // it now reads the agent's credential file like every other command. Probe the RESOLVED
-    // model's provider through the normal resolution path (stored credential OR env var — an
-    // env-authed user is fine and must not be warned): only when that provider has no usable auth
-    // AND pi's old file exists does the bare provider error get its cause named.
-    if (
-      (await probeAuthSource(modelRuntime, modelSpec)) === undefined &&
-      existsSync(join(getAgentDir(), "auth.json"))
-    ) {
-      log.warn(
-        `[fastagent] no credentials for ${modelSpec} in ${authPath} — this runtime no longer reads ` +
-          `pi's ~/.pi auth; run \`fastagent login\` (or /login in the TUI) to store credentials for this agent`,
-      );
-    }
-    const model = resolveModel(modelRuntime, modelSpec);
     const env = new NodeExecutionEnv({ cwd });
     const definition = await loadAgentDefinition(agentDir, { cwd, env });
     reportFindingsIfChanged(definition.dir, definition);
     // Assembly-time, like serving's: this whole function is memoized, so the scan and its warnings
     // happen once per runtime rather than per session rebuild (/new, /resume, fork).
     const extensionPaths = await loadExtensionPaths(agentDir, { cwd, env });
-    const defaultNames = piDefaultTools().map((t) => t.name);
-    const customTools = tools.filter((t) => !defaultNames.includes(t.name));
+    // ALL of them, pi's four included: fastagent's `read` is `createReadTool({ imageProcessor })`
+    // (create.ts), and letting pi mount its own instead would silently drop image reading in chat.
+    // Serving already does it this way.
+    const customTools = tools;
     // Adapt fastagent's AgentTool to pi's ToolDefinition (`parameters` is plain JSON-Schema; pi accepts
     // it). Each execute runs inside the turn context with the CURRENT session's activation bridge — the
     // assembly is memoized across /new//resume/fork rebuilds while the session changes, so the bridge
@@ -169,10 +156,9 @@ export async function buildAgentSessionRuntime(
         if (!bound) throw new Error("tool executed before its session was built (lifecycle invariant broken)");
         return turnContext.run(
           { cwd, sessionManager: bound.sessionManager, tools: bound.activation },
-          // pi's per-turn TOOL context (5th parameter) is read only by its default coding tools, and
-          // those are filtered out of `customTools` above; fastagent's own take theirs from
-          // `turnContext` (AsyncLocalStorage). So this env exists to satisfy the shape, and it is the
-          // chat cwd's — the same root a default tool would have got, had one reached here.
+          // pi's per-turn TOOL context (5th parameter) is read only by its default coding tools;
+          // fastagent's own take theirs from `turnContext` (AsyncLocalStorage). This env satisfies
+          // the shape for both, and is the chat cwd's — the same root pi would have handed them.
           () => t.execute(id, params, signal, undefined, { env }) as Promise<unknown>,
         );
       },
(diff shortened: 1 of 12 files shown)


src/engines/pi/session-builder.ts
```diff
@@ -245,6 +251,10 @@ export async function buildAgentSessionRuntime(
         // ~/.pi extensions, slash commands, global AGENTS.md, APPEND_SYSTEM.md) so this runtime runs
         // the same agent that gets served, not the authoring machine's pi setup on top.
         noExtensions: true,
+        // ...except the definition's OWN extensions/: pi honours additionalExtensionPaths even under
+        // noExtensions. Without this the same definition would carry its extensions when served and
+        // lose them in chat — one artifact, two behaviours.
+        ...(extensionPaths.length > 0 ? { additionalExtensionPaths: extensionPaths } : {}),
```
kid7st: 问题：extension 直到 `createAgentSessionServices()` 才加载，但 `modelSpec` 已在 `session-builder.ts:143` 解析完成。`pi.registerProvider()` 的职责是让 extension 在启动期注册模型/provider；当前顺序使配置使用 extension 提供的模型时直接报 `unknown model`，同时还会在 provider 注册前输出错误的缺凭据警告。因此“chat runs extensions in full”缺失了 extension 的模型注册语义。

建议方向：先创建 services 并完成 extension provider 注册，再执行 auth probe 和 `resolveModel`；增加一个由 extension 注册 provider、并由 `fastagent.config` 选择其模型的 chat-path 测试。

src/engines/pi/session-builder.ts
```diff
@@ -270,13 +280,20 @@ export async function buildAgentSessionRuntime(
...
+      .getExtensions()
+      .extensions.flatMap((extension) => [...extension.tools.keys()]);
     const result = await createAgentSessionFromServices({
       services,
       sessionManager,
       sessionStartEvent,
       model,
       thinkingLevel,
-      tools: [...defaultNames, ...customTools.map((t) => t.name)],
+      tools: [...defaultNames, ...customTools.map((t) => t.name), ...extensionToolNames],
```
kid7st: 问题：这里把启动时已有的 extension tool 名称固化为 pi 的 `tools` allowlist。pi 明确支持在 `session_start`、command 或其他 handler 中调用 `pi.registerTool()`，但后注册的名称不在 allowlist，`refreshTools()` 会立即把它过滤掉；extension 看似注册成功，模型和 `pi.getAllTools()` 都看不到该工具。这与“chat runs them fully”不一致。

建议方向：不要用启动快照限制 extension tool registry；保留 FastAgent 自己的 builtin/custom tool 选择，同时让 extension 后续注册的工具进入 registry。增加一个启动后注册工具并断言 `getAllTools()` 可见的测试。

kid7st (commented):
本次改动把定义内的 `extensions/` 定义为随 artifact 携带的 pi 代码入口：只在单 resident session 的 `chat` 中运行，serving 仅发现并明确告警。

src/engines/pi/session-builder.ts
kid7st: Confirmed and fixed in bf7dbb23.

The ordering was worse than described: nothing ever called `bindExtensions()`, so `session_start` never fired in chat at all — an extension registered what it registered at import time and then went deaf. That is now called (with the error listener attached), and services are built **before** `resolveModel`, so a model registered via `registerProvider()` resolves.

Test: `resolves a model registered by an extension's registerProvider()` — it failed with exactly the `unknown model "acme-proxy/acme-1"` you predicted.

src/engines/pi/session-builder.ts
kid7st: Confirmed and fixed in bf7dbb23.

Verified with an extension registering one tool at import time and one from `session_start`: only the first appeared. Dropped the allowlist in favour of `noTools: "builtin"`, which gives the guarantee the allowlist was actually there for (the machine's `defaultTools` setting cannot add pi's own copies on top of ours) without freezing the set.

This also uncovered a bug nobody had reported: allow-listing pi's default tool *names* mounts **pi's** `read`, not fastagent's `createReadTool({ imageProcessor })` — chat had been running without image reading. Chat now passes every fastagent tool as a `customTool`, exactly as serving does.

Tests: `fires session_start, so a tool registered from a handler reaches the session` and `uses fastagent's own read tool, not pi's copy of it`.

PR #349, opened 2026-08-21
refactor(channels): an attachment is a fact, not an assumption about the reader

kid7st:
Removes the capability flag that #331 threaded through eight files, by removing the assumption that
made it necessary.

## One half-sentence

The attachment manifest said:

```
[attached files — read them with your tools: …]
```

"read them with your tools" is an assumption about the reader, and an assumption has to be verified.
That verification is where all of this came from:

| | |
|---|---|
| `canReadLocalFiles` | threaded config → CLI → serve → `ChannelContext` → 3 transports → invoke-turn |
| `codingToolNames.includes("read")` | the same derivation, written **9 times** |
| `LocalFileAccessUnavailable`, `ATTACHMENT_UNSUPPORTED_CODE`, `unreadableAttachmentsNote`, a `defaultErrorMessage` branch | the machinery for refusing when the assumption failed |

## Images never needed any of it

```ts
interface ImageRef { mimeType: string; data: string; }   // a value
interface DownloadedFile { path: string; name: string; size: number; }   // a reference
```

`ImageRef` carries `data`, so a channel hands over a value and the engine decides what to do with
it. **No channel anywhere asks whether the model has vision** — zero occurrences of any such check.

A file carried a path — a reference, which only means something if the holder can dereference it.
That is the entire difference, and it is why the agent's tool configuration leaked into every
channel.

## What replaces it

The manifest states name, size and path. Nothing else changes hands.

An agent with a file tool decides for itself whether to open one and how much of it to read (`read`
takes offset/limit — better than any threshold we could have picked). An agent without one answers
that it cannot: visible, and the direct consequence of its own `codingTools`, which is what
configuring a capability means.

A considered and rejected middle: inlining small text files below some size threshold, so a
tool-less agent could still read them. It trades one guess for another — the channel would be
deciding *whether the agent needs the content* instead of *whether it can read it* — and the number
would have been mine, not measured. If usage later shows tool-free attachment reading matters, it
can be added against evidence.

## The same assumption, one size smaller

`report.ts` had a pre-flight version of it: at assembly, warn when a definition lists model-visible
skills while mounting no `read`. Removed too, along with `AssemblyFinding`, the `assemblyFindings`
key in `info --json`, and `reportFindingsIfChanged`'s third parameter.

Its answer was already wrong in a case #331 itself established. Authors may define their own tool
named `read` — `disabledBuiltinNames` subtracts exactly that name from the exclusion list — so
`codingTools: false` plus `tools/read.ts` is a perfectly readable agent that this check warned about
anyway. Every new way to have a reader is another way for a pre-flight guess to be wrong.

What it bought was hearing at startup rather than at first use, and it only ever warned, so nothing
was gated on it. An agent asked to use a skill it cannot open says so — same visibility, one turn
later, no guess.

**Public surface note:** `info --json` no longer emits `assemblyFindings`. `diagnostics` keeps its
original meaning (per-skill parse problems), and `docs/cli.md` is updated.

## Validation

- [x] `npm run lint`, `npm run typecheck`, `npm test` (108 files, 1376 tests)
- [x] `rv.sh local` — **No findings.** It caught two tests that asserted removed behaviour, one of
      which had been passing for an unrelated reason
- [x] Net **−408 lines**

## Checklist

- [x] Docs updated (`docs/configuration.md`, `docs/api-reference.md`)

### DIFF
--- a/docs/api-reference.md
+++ b/docs/api-reference.md
@@ -283,7 +283,6 @@ Costs and behavior to know:
 interface ChannelContext {
   agent: Agent;
   stateRoot: string; // resolved state root (FASTAGENT_STATE_DIR > <root>/.state), absolute
-  canReadLocalFiles: boolean; // required: whether the agent has a reader for absolute local paths
 }
 type ChannelModule = (ctx: ChannelContext) => Routes;
 interface LongConnection {
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -75,7 +75,7 @@ Prints the assembled surface without starting a server:
 - model source,
 - config path,
 - persona presence and context files (`AGENTS.md`),
-- skills, their diagnostics, and any assembly findings (`assemblyFindings` — definition-wide, e.g. skills with no reader),
+- skills and their diagnostics,
 - the resolved coding-tool names (`read`/`bash`/`edit`/`write`), plus authored tools and collisions,
 - channel files,
 - schedules (name + next fire instant; a broken schedule file is reported here, not first at `dev`),
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -294,10 +294,14 @@ export default defineConfig({
 ```
 
 Model-visible skills are loaded on demand from their `SKILL.md` paths, and chat-channel non-image
-attachments are downloaded to local paths. Both need the built-in `read` tool; FastAgent warns when
-model-visible skills have no reader, and channels reject an explicitly attached file rather than hand
-the model an unreadable path. Use `false` only when the agent needs neither capability (images still
-travel inline through vision).
+attachments are downloaded to local paths. Both are read with the built-in `read` tool.
+
+Skills get a startup warning when they are listed to the model but `read` is absent — an assembly
+that advertises a capability it cannot deliver is contradicting itself, and only the assembly can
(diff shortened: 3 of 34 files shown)


kid7st (commented):
This change makes attachment manifests state resolved file facts without channels deciding whether the Agent can read them, and also removes the analogous skills-reader preflight.

**PR description — ‘What survives, and why’**

Problem: The description says `report.ts` still owns the reader derivation, now accepts tool names, and removes duplicated checks from six callers. The head instead deletes `definitionAssemblyFindings`, `AssemblyFinding`, `info --json.assemblyFindings`, and every caller. The PR therefore describes the opposite of the final behavior and omits a public JSON surface change.

Direction: Remove or rewrite this section to state that the skills-reader preflight and the `assemblyFindings` JSON field are gone, and explain the final decision.

**`src/engines/pi/create.ts:473-474, 523-527`**

Problem: The `codingToolNames` comments still say it drives ‘skill findings’ and that lacking a reader warns that skills cannot be read, although this PR removes that warning completely. The comments now assign `codingToolNames` a responsibility it no longer has and describe behavior that cannot occur.

Direction: Remove the reader/skill-finding claims and retain only the current responsibilities: prompt identity and disabled built-in-name calculation.

PR #351, opened 2026-08-22
refactor(tools): make the coding tool set an agent invariant

kid7st:
## Summary

Directory agents always mount pi's complete coding-tool set: `read`, `grep`, `find`, `ls`, `bash`, `edit`, `write`.

The `codingTools` config option and everything that served it — validation, resolver branches, session denylist plumbing, scaffold examples, tests — are gone. Supplying `codingTools` is now an unknown-config-key error.

`config.tools` and discovered `tools/` still append authored tools. If an authored tool reuses a coding-tool name, the built-in wins and the collision is reported.

## Why

Coding tools are basic agent capabilities, not an isolation boundary. A built-in allowlist cannot constrain authored `tools/` or channel code — both are ordinary executable code that imports whatever it likes — so it narrows a blast radius rather than closing one. Deployments that need isolation must sandbox the whole agent process.

That leaves one rule for directory agents instead of a boolean/array option whose meaning shifted with its shape.

## Lower-level API

`createPiAgentFromDefinition` defaults to the seven; an explicit `tools` replaces that coding set. `createPiAgent` starts from the tools its caller passes. In both, deferred tools may still append `search_tools`.

Omitted coding built-ins are now refused at the **registry**, not merely left inactive. pi's `noTools: "builtin"` keeps its copies off the active set but leaves them registered, so a `search_tools` activation could resurrect a `bash` the caller had excluded. The shared assembly path now computes the omitted names and passes them as a denylist, covering both rungs.

## Breaking

`CodingToolName` is no longer exported from `@fastagent-sh/fastagent/pi`. It described the config option's value type and has no public API left to describe.

## Cleanup

`codingToolNames` was threaded from the tool resolver through `AgentAssembly` and `createPiAgentFromDir` into three CLI reports, where it had become a constant equal to the full set. The reports read the constant directly; the field is dropped from both return types. `codingToolsLabel` went with it (its `(none)` branch was unreachable once every caller passed a constant), as did the `piDefaultTools` alias for `piAllCodingTools`.

## Validation

- [x] `~/.pi/agent/scripts/check.sh` — lint, typecheck, 107 test files / 1371 tests
- [x] Local semantic review clean
- [x] Removing one name from `CODING_TOOL_NAMES` fails 6 tests across `definition`/`tool`/`cli`, confirming the invariant is asserted at the CLI's JSON output and not only at the constant
- [x] Docs, scaffold template, and API reference updated


kid7st (commented):
本次改动把工具配置重定义为：固定挂载 pi 的只读基线，再由配置选择附加的 mutating tools；它位于目录 agent 的 config → tool resolution → serving/chat assembly 边界。

`docs/cli.md:79`

问题：`fastagent info` 的文档仍把 resolved coding-tool names 定义为 `read`/`bash`/`edit`/`write`，但本次改动后默认报告七项，`codingTools: false` 也报告四项只读基线。CLI 文档描述的输出结构已与实现不一致。

建议方向：把该处更新为完整七项，并说明报告值是“固定只读基线 + 已启用的 mutating tools”。

src/engines/pi/config.ts
```diff
@@ -35,10 +52,18 @@ export interface FastagentConfig {
...
+   * ({@link BASELINE_TOOL_NAMES}: `read`, `grep`, `find`, `ls`).
+   *
+   *  - unset / `true` — all of them: `bash`, `edit`, `write`
+   *  - `false` / `[]` — none; the agent can read and search, and that is all
+   *  - an array — exactly those, e.g. `["edit"]` for an agent that writes files but runs nothing
+   *
+   * Authored tools (`config.tools` + discovered `tools/`) and conditional built-ins (`search_tools`,
+   * `wake`) are independent of this.
+   */
+  codingTools?: boolean | MutatingToolName[];
```
kid7st: 问题：`codingTools` 现在只控制 `bash`、`edit`、`write`，而 `read`、`grep`、`find`、`ls` 这些同属 coding tools 的能力永远存在。名称仍表示“编码工具集合”，但 `false` 并不会关闭编码工具，数组也不接受大部分编码工具名；调用方会被字段字面含义误导。

建议方向：把配置项改名为 `mutatingTools`，并同步 CLI 报告字段和文档；`CodingToolName` 继续表示完整七项集合，`MutatingToolName` 表示该配置可选的三项。

kid7st: Keeping `codingTools` is deliberate: the confirmed public API is `codingTools: [names]`, with array semantics defined as the fixed read-only baseline plus the named mutating tools. Renaming it after that API decision would solve a naming preference by changing the contract. `MutatingToolName` already makes the accepted values explicit, while CLI `codingTools` intentionally reports the complete resolved surface. I updated the stale CLI documentation separately in 1a247600.

kid7st:
Addressed the review-body finding in 1a247600: `fastagent info` now documents the complete resolved set (the fixed read-only baseline plus enabled mutating tools).

src/engines/pi/config.ts
kid7st: Superseded by 98fb86fe: the config option is now removed entirely. Directory agents always mount the complete coding-tool set; isolation belongs at the sandbox boundary.

src/engines/pi/create.ts
```diff
@@ -584,9 +572,8 @@ export async function createPiAgentFromDefinition(
     // change without a restart, which is why this sits outside `live` above.
     extensionPaths: await loadExtensionPaths(dir, { cwd, env }),
     cwd,
-    // The disabled built-ins, refused at the registry — see PiAgentSessionFactoryOptions. A name an
-    // AUTHORED tool has taken is not excluded: with the built-in off, that name is the author's to
-    // reuse (documented), and pi's denylist works on names, so excluding it would delete their tool.
+    // Lower-level callers may pass an exact list. Keep omitted pi built-ins out of that session;
+    // a reused name stays because pi's denylist would otherwise remove the caller's tool too.
     excludedToolNames: disabledBuiltinNames(codingToolNames, tools),
```
kid7st: 问题：PR 明确保留“createPiAgentFromDefinition 的显式 tools 列表替换默认集合，并拒绝未列出的 pi built-ins”这一低层保证，但现有测试只检查 partial list 生成的 prompt，没有验证运行时 registry/activation 确实不能恢复被省略的 built-ins。disabledBuiltinNames 或 session factory 的 denylist plumbing 失效时，当前测试仍会全部通过。

建议方向：增加一个运行时测试，传 tools: [] 或 partial list，断言省略的 bash/write 不在 registry 中，且按名称激活也不能恢复它们。

kid7st (commented):
本次改动把目录 agent 的 coding-tool 语义改为固定挂载完整七项集合；目录 config 不再控制该能力，低层 L2 API 仍可用显式 tools 列表替换。

README.md:199、docs/embedding.md:152、docs/design/core.md:304、src/engines/pi/create.ts:64、src/engines/pi/session-builder.ts:14

问题：本次改动把目录 agent 的语义定为“七项 coding tools 固定挂载，并且全部绕过 ExecutionEnv”，但 companion text 仍分别声称 ExecutionEnv 管理默认 coding tools、默认集合只有四项、grep/find/ls 是否进入默认集合尚未决定，以及 chat 只按名称让 pi 重建工具。公开边界说明和实际实现相互矛盾。

建议方向：统一改成七项固定集合；明确它们都由 fastagent 作为 customTools 挂载并直接访问本地进程，ExecutionEnv 只负责 definition loading。

src/engines/pi/create.ts
kid7st: Fixed in 5cd5db79. The test now captures the real L2 session factory, asserts the runtime registry contains only the explicit `read` tool, and verifies activating `bash`/`write` by name cannot restore them. That test failed when the denylist plumbing was removed, proving `noTools: "builtin"` alone is insufficient.

kid7st:
Addressed the review-body companion-text finding in 5cd5db79 and 17d969db: README, embedding/API/design docs, and source comments now consistently state that all seven directory coding tools are mounted through `customTools`, bypass `ExecutionEnv`, and require process-level sandboxing for isolation.

docs/api-reference.md
```diff
@@ -172,24 +172,18 @@ function createPiAgentFromDir(
...
-order. Every directory-opening workflow (`dev`, `start`, `invoke`, `chat`, `tool`, and `info`) applies
-the same resolved surface. Model-visible skills and chat-channel non-image attachments require the
-built-in `read`; use `codingTools: ["read"]` when those capabilities remain needed. Conditional
-built-ins stay independent: deferred tools may add `search_tools`, and `selfSchedule` may add `wake`
-while serving. The lower-level `createPiAgent` API remains explicit: its `tools` option is the exact set
-supplied by the caller and has no `codingTools` option.
+Every directory-opening workflow (`dev`, `start`, `invoke`, `chat`, `tool`, and `info`) mounts the
+complete coding set. Conditional built-ins stay independent: deferred tools may add `search_tools`,
+and `selfSchedule` may add `wake` while serving. `createPiAgentFromDefinition` uses the complete set
+unless `tools` replaces it; the lower-level `createPiAgent` mounts only the `tools` explicitly passed.
```
kid7st: 问题：这里承诺 `createPiAgent` 只挂调用方显式传入的工具，但 L1 没有像 L2 一样传 `excludedToolNames`。本 PR 的 L2 registry 测试已经证明 `noTools: "builtin"` 只抑制默认激活，不会移除 pi built-ins；因此 L1 的省略项仍留在 registry，之后可按名称激活，和“only the tools explicitly passed”的低层语义不一致。

建议方向：在 `createPiAgent` 中也根据最终 `tools` 计算并传入 `omittedBuiltinNames`，并增加 L1 runtime registry 测试，断言省略的 built-ins 不在 `getAllTools()` 中且无法按名称恢复。

kid7st (commented):
本次改动把目录 agent 的 coding-tool 语义改为固定挂载完整七项集合；目录 config 不再控制该能力，L2 仍允许显式 tools 替换，L1 则应只挂调用方传入的工具。

docs/api-reference.md
kid7st: Fixed in 6b628b74 by moving omitted-builtin calculation into the shared `buildPiAgent`, so L1 and L2 use the same registry rule. The L1 test invokes a real loader that tries to activate `bash`; removing the denylist makes it fail with `activated = ["bash"]`.

kid7st: 问题：这里把 L1 描述为“只挂载显式传入的 tools”，并把低层列表称为 exact；但 L1/L2 都会经过 `withSearchTool`，只要列表里有 `deferred` tool，就会额外挂载未显式传入的 `search_tools`。同段前一句也承认这个例外，因此当前契约表述自相矛盾。

建议方向：把保证限定为“显式列表替换七项 coding-tool 默认集合，省略的 coding built-ins 无法恢复”；同时明确 `deferred` tool 会隐式加入 `search_tools`，并把相关测试名和注释中的 “exact” 改成这个更窄的保证。

kid7st (commented):
本次改动把目录 agent 的七项 coding tools 变成固定能力；只有 L1/L2 低层 API 保留显式列表替换 coding-tool 默认集合的语义。

docs/api-reference.md
kid7st: Fixed in f1546715. The contract and tests now say that explicit `tools` replace the coding defaults and omitted coding built-ins cannot be restored; they no longer claim the final registry is exactly the passed list. Deferred tools are explicitly documented as adding `search_tools`.

PR #355, opened 2026-08-24
feat: mount an agent in one call, and make /control/* browser-reachable

### LINKED ISSUE #352: /control/* is unreachable from any browser client: no CORS preflight handling
## What

`controlRoutes` registers no `OPTIONS` handler and sends no `access-control-*` headers, so
`/control/*` cannot be read from any browser context — a Web panel, an Electron renderer, or
anything else that reaches the plane through `fetch`.

## Why it bites every request, not just complex ones

The bearer token travels in `Authorization`, which is not a CORS-safelisted request header. Every
call to the plane is therefore preflighted, including a plain `GET /control/state`. The preflight is
`OPTIONS`, which no route registers, so `router` answers `405 method not allowed` — correctly, since
the path exists under other methods — with no `access-control-allow-origin`. The browser rejects it
before the real request is ever sent, and before the token is ever checked.

## Why it matters here

The design names the consumers this blocks:

> One transport serves every remote consumer — Web panel, desktop app, `fastagent attach`

`fastagent attach` is unaffected, because Node's fetch does not enforce CORS. The other two are
exactly the ones that cannot connect, and they are much of the reason the transport exists.

It is not a local-only problem either. A GUI client is never same-origin with the plane: a dev
server while developing, `file://` once packaged, and a different host entirely against a
deployment.

## Suggested shape

Answer `OPTIONS` with 204, and send on every response:

```
access-control-allow-origin: *
access-control-allow-headers: authorization
access-control-allow-methods: GET, POST, OPTIONS
```

`*` is the right value rather than a concession. Authorisation on this plane is the token — never
the origin, never a cookie — so an origin that cannot present the token gets 401 either way.
Restricting it would also be unimplementable for a deployment, which cannot know the origins of the
GUIs that will manage it. That is the same asymmetry §14 already settles with "the bearer token is
the only auth the framework owns".

## Verified

Reproduced against a stub implementing this plane's shape, from an Electron renderer in both
positions a packaged client takes: an `http://localhost` dev origin and a `file://` production
origin. Both fail at preflight before the token is checked; both pass once `OPTIONS` plus the three
headers are added, with no other change.

kid7st:
Closes #352.

## Summary

`/control/*` is reachable from a browser for the first time. Getting there took a structural change rather than a header patch, and that turned out to be the actual fix.

**The plane owns the `/control` prefix as a mount**, instead of being a set of routes that happen to share it. CORS belongs to every reply that *leaves* the plane — and three of those are produced where no route runs: an unknown path under the prefix, a method a path does not serve, and a handler that throws. As separate routes those came from the host, outside anything the plane could decorate, so each surfaced as a bare "network error" in a browser and each had to be found separately. Owning the prefix makes them the plane's own answers, applied at the single exit they share.

```
access-control-allow-origin: *
access-control-allow-headers: authorization, content-type
access-control-allow-methods: <what this path serves> + HEAD where GET is served + the requested method
```

`fastagent attach` was never affected — Node's fetch does not enforce CORS — which is why the gap stayed invisible while blocking every browser client.

## Why the values are what they are

- **`*`** — authorisation here is the bearer token, never the origin and never a cookie, so an origin that cannot present it gets 401 either way. A deployment also cannot know the origins of the GUIs that will manage it (§14's asymmetry).
- **`content-type`** — only three content-types are CORS-safelisted and `application/json` is not among them, so a browser POSTing to `dispatch`/`invoke` names it in the preflight. Allowing only `authorization` (as the issue suggested) leaves precisely the plane's two WRITE routes unreachable while every read works. Invisible to a probe that omits the header, since `fetch` then defaults to safelisted `text/plain`.
- **The requested method, always** — CORS headers on a 404 are worthless if the browser never sends the request. Preflight is a gate applied *before* the request exists: refusing there (unknown path, unsupported method) means the real request is never sent and the client sees the same opaque error this PR removes. Every preflight under the prefix is answered `204`, and the plane's own `404`/`405` carries the explanation.

## Dispatch is a map lookup, not a router

A deployment mounts a handful of **literal** paths — one per channel, plus health, plus the plane's prefix. For literal paths, "which handler serves this request?" is `map.get(pathname)`, and "would these two fight over a request?" is string equality.

A routing library answers the first question through a pattern language we do not use, and its extra semantics — decode-before-match, HEAD fallback, wildcard precedence, registration order — then have to be *predicted* by every collision check. That prediction is what drifted: eight separate rounds of this review were one unpredicted detail each. The behaviour is now the twelve lines that implement it.

`hono` is gone; `@hono/node-server` stays for what it is genuinely good at — replacing 60 lines of hand-written `node:http`↔Fetch bridging, the most failure-prone code in the repo. `overrideGlobalObjects: false` keeps it there: the default swaps `globalThis.Request`/`Response` process-wide, which reaches every line of an embedding app — and breaks code here first, since a channel holding a `Response` captured before mount fails an `instanceof` check and is answered 500 by the very boundary meant to protect it.

**No channel changes.** `ChannelHandler`, `Routes` and `serveNode()` keep their signatures — SPEC §11 fixes the gateway contract as `(Request) => Response`, channel directories ship hand-written modules with it, and an embedding app may already run Express/Fastify. No channel, scaffold template, or user `channel.ts` changes.

## The route language

A key is `"METHOD /path"` or `"/path"`, and the path is literal. Validation exists only to tell an author their handler will not run — an unmatched key is simply never found — so it covers the spellings someone plausibly writes *believing they work*:

| Refused | Because |
|---|---|
| anything `URL` rewrites (`?`, `#`, `.`/`..`, `%2e`, `\`) | the request arrives under the rewritten path, so the key is unreachable **and** compares as a different string, hiding the conflict. Asked of `URL` rather than by listing spellings |
| two keys naming one route (`"/x"` and `"GET /x"`) | registration order would pick a winner in silence |
| a route inside a mount, or two overlapping mounts | the mount answers it, including on paths it does not serve |

Every rule that survived has the same shape: breaking it costs **someone else**, and a channel goes dark whose author did nothing wrong. Rules that only protect an author from their own key were cut — `:param` and `*` are ordinary characters here, so such a key merely never matches, and an unusual method is the same. Refusing those trades a working capability for a lecture.

`HEAD` is answered from the `GET` route without the content (RFC 9110), body cancelled rather than left to a streaming producer with no reader; an explicit `HEAD` route is allowed and wins. Paths are matched **as they arrive**, without percent-decoding — decoding would undo the normalisation `URL` performs, turning `%2F..%2F` back into `/../`.

## Layering corrected along the way

The CORS gap was hard to see because the plane's replies were terminated by code in a file named for a channel: the serving bridge lived in `src/channels/http.ts`, so `host/node.ts` imported from `channels/` — the dependency pointing backwards.

Fixing that exposed a second misnaming. `src/host/node.ts` claimed a product dimension it was not a member of. In `core.md` §1, **Host/runtime** means "supplies process, storage, credentials, and deployment" — that is `src/deploy/` (docker, fly, railway, agentcore), which already calls itself host-neutral and adds one directory per host. `serveNode` supplies none of those; it is the same in-process mechanism *every* one of those hosts runs. Its file header additionally labelled itself "K-axis", a term this codebase uses for the durable-execution backend — a third, unrelated dimension.

So the pieces are split by what makes them change:

| | | |
|---|---|---|
| `src/channel.ts` | the Channel contract (`ChannelModule`, `Routes`, `LongConnection*`) | beside `agent.ts`/`session.ts` — the trigger side of §1, pure types |
| `src/channels/serve.ts` | path rule, dispatch, totality boundary, node:http binding | shared ground for every deploy target, not one of them |
| `src/host/` | — | deleted |

A WebSocket ingress needing `LongConnection` no longer pulls `node:http` and an HTTP framework in behind a type import. `channels/http.ts` drops to 125 lines and imports no `node:` builtin at all. `core.ts` re-exports both, so **no user-visible import path changes**.

The three contracts (`agent.ts`, `channel.ts`, `session.ts`) are now pinned at **zero package dependencies**, with `channels/serve.ts` pinned at exactly one. That guard exists because the first version of it did not catch a framework import into the contract.

## Embedding is held by real frameworks now

`docs/embedding.md` tells authors to mount fastagent inside an app they already run, and every one of those snippets crosses the Node/Fetch seam through code we do not own. It was verified by hand, which is not a regression net. `test/embedding.test.ts` runs the documented mounts against real Express and Fastify (the sole reason they are devDependencies).

Writing it immediately found a defect: `serveNode` took a `ChannelHandler` while `nodeListener` demanded `(req) => Promise<Response>`, so `nodeListener(router(routes))` — the whole agent on the app's own server — was a type error.

It also pinned down the one genuine conflict in embedding. Node's request is a one-shot stream, so a body parser registered *ahead* of the mount consumes it and nothing reaches the agent — and webhook channels verify signatures over the raw bytes, so rebuilding from a parsed body is not a choice we can make for the author. The fix is mount order, now the documented form, and the failure names it:

```
[serve] POST /hook: the request body was already read by upstream middleware (e.g. express.json())
— mount fastagent BEFORE the body parser, or scope the parser away from this route.
```

That guard is deliberately silent unless it is *certain* (a positive `content-length`): an empty POST and an empty chunked request both arrive drained and indistinguishable from an eaten one, and both were rejected before this was fixed. A diagnostic that rejects valid traffic is worse than none.

## The product was missing its entry

Splitting the layers made the gap obvious: the library exported *parts* — discovery, a router, a control plane, a Node binding — and the only code that knew how to assemble them into a running agent was the CLI. "I have an agent directory, mount it in my app" had exactly one supported path, and it was `fastagent start`.

Assembling it by hand is also easy to get wrong in a way nothing catches. Our own code got it wrong: the AgentCore branch composed a router and forgot the control plane's mount.

```ts
const service = await createAgentService("./my-agent");
app.use(service.handler);        // channels, health, /control — all of it
await service.close();
```

`dev` and `start` are now callers of it, so the CLI and an embedder run the same assembly rather than two that drift. It binds no port, installs no signal handler, and never calls `process.exit` — a library embedded in someone's app does not get to make those decisions. A test pins the four public entries as unable to reach `src/cli/`, where that decision lives.

## The surface says the product, not its parts

With one entry that assembles everything, the pieces it assembles stopped earning their place in the public API:

| | |
|---|---|
| `router`, `PrefixMount`, `createControlPlane` | internal — how a service is composed, not something a caller reproduces |
| `AgentService.mounts` | **removed** — added for "a host that must re-wrap them", used by no production caller. A field kept for a hypothetical caller is a field nobody maintains |

Tests already imported these from their modules, so nothing there moved. #356 tracks applying the same standard to the exports that predate this PR (`createScheduler`, `loadChannels`, `loadTools`, …).

## Behaviour changes

| | before | after |
|---|---|---|
| `HEAD` on a `GET` route | 405 | 200, no content (RFC 9110) |
| route key language | undocumented, unvalidated | literal path, refused with the reason |
| prefix ownership | — | a mount, passed beside the routes |
| percent-encoded request path | — | a distinct path; never decoded |
| channel route overlap | string equality on whole keys | `routeKeysConflict` / `pathUnderPrefix`, one rule for every caller |

## Validation

- [x] `~/.pi/agent/scripts/check.sh` — lint, typecheck, 109 test files / 1440 tests
- [x] Local semantic review clean
- [x] SPEC MUST 3 verified through the adapter: client disconnect → `req.signal` abort → `ReadableStream.cancel()` → `iterator.return()` → invoke cancellation
- [x] The `pump` contract tests were rewritten from white-box (a hand-built fake `ServerResponse`) to black-box (real socket, real disconnect) — the requirements were never about `pump`, they were about the host: never crash, never leak, always surface
- [x] Preflight is asserted the way a **browser** gates it (does the response let the real request through?), not through Node's fetch, which does not enforce it
- [x] Every change reverted independently turns the suite red — each dispatch rule, each refusal, each guard, in both directions where a guard could misfire


src/channels/control.ts
```diff
@@ -40,6 +40,47 @@ export interface WireEvent {
...
+ * The plane's routes, made reachable from a browser: CORS headers on EVERY response (a 401 or 400
+ * without them reaches the client as an opaque "network error", hiding the very answer it needs),
+ * plus an unauthenticated `OPTIONS` per path — a preflight carries no token, which is its entire
+ * purpose. Paths are DERIVED from the table, so a route added later is covered without anyone
+ * remembering to; `router` does exact matching, so each one needs its own entry.
+ */
+function browserReachable(routes: Routes): Routes {
+  const wrapped: Routes = {};
+  for (const [key, handler] of Object.entries(routes)) {
+    wrapped[key] = async (req) => withCorsHeaders(await handler(req));
```
kid7st: `browserReachable` only decorates handlers that successfully return a `Response`; it is not the whole-plane response boundary its name and documentation promise. Two paths still become opaque browser network errors: if a valid handler rejects (for example `commands()` throws), `nodeListener` synthesizes a bare 500 outside this wrapper; and if a client uses the wrong method, this preflight advertises both GET and POST, then `router` synthesizes a bare 405 outside the wrapper. The new test only exercises declared methods and handlers that return normally, so it misses both leaks. Make the CORS boundary cover handler rejection and router-generated method errors: at minimum convert rejected handlers into a logged, CORS-bearing 500, derive preflight methods per path, and provide a CORS-bearing 405 fallback. Add one rejecting-handler case and one wrong-method browser request to the test.

kid7st (commented):
This change makes the control-plane route table browser-reachable by deriving unauthenticated preflight routes and decorating control responses with CORS headers.

src/channels/control.ts
kid7st: Fixed in 34b1ac65 — two of the three, and the third deliberately not.

**Handler rejection → CORS-bearing 500.** Correct, and reachable: `commands()` on an unreadable definition is an existing tested path. `browserReachable` now has its own totality boundary — logs the error (the host's catch no longer sees it) and keeps the internal message out of the body. New test asserts 500 + headers + the log line + that the message stays internal.

**Wrong-method 405.** Fixed at the cause rather than the symptom: `access-control-allow-methods` is now derived **per path**, so a POST-only route no longer advertises GET. The browser refuses that request itself, with an accurate diagnosis, and the unreadable 405 never happens. The preflight sweep now asserts that advertised methods equal the methods the path actually mounts.

**A CORS-bearing 405 fallback: not added.** With per-path methods I could not construct a browser scenario that reaches it — anything carrying `Authorization` preflights first, and the preflight now tells the truth; same-origin requests skip CORS entirely; non-browser clients (`attach`) do not enforce it. That makes it a guard for a case I cannot make happen, which this repo treats as worse than no guard, since it reads as covered without being exercised. If you have a concrete path that reaches it, I will take it with a test.

Verified by reverting each change independently: plane-wide methods, and removing the catch, each turn the suite red.

src/channels/serve.ts
```diff
@@ -0,0 +1,265 @@
...
+  return ka.method === undefined || kb.method === undefined || ka.method === kb.method;
+}
+
+/**
+ * Would these two route paths answer the same request? A literal pair collides when equal; a prefix
+ * mount collides with anything beneath it, including a path it does not itself serve — the request
+ * would still reach the mount's own 404 instead of the other channel. Method-blind: {@link
+ * routeKeysOverlap} is what callers comparing whole keys should ask.
+ */
+export function routePathsOverlap(a: string, b: string): boolean {
```
kid7st: Problem: collision detection compares raw path strings, while Hono decodes percent-encoded unreserved characters before matching. For example, `"/control/*"` and `"GET /%63ontrol/mine"` are reported as non-overlapping, but a request to `/%63ontrol/mine` is handled by the prefix route and the literal handler is permanently unreachable. The same mismatch makes `GET /control/%63apabilities` succeed while its browser preflight returns 404, because `planeApp` looks up the undecoded `URL.pathname` in `methodsByPath`. This breaks both core promises of the change: no silent route shadowing and browser reachability for every route the server actually serves.

Suggested direction: use one canonical path representation for registration, overlap checks, and control-plane method lookup. The narrowest fix is to reject non-canonical percent-encoded route keys and canonicalize request pathnames with the same decoding semantics Hono uses; add one encoded-alias collision check and one corresponding CORS preflight check.

kid7st (commented):
This change makes `/control/*` a prefix-owning sub-application and centralizes route matching, collision detection, CORS, and Node/Fetch serving around that route model.

src/channels/serve.ts
kid7st: Confirmed and fixed in the last two commits — both halves were real, and verified against the matcher rather than reasoned about.

**Decoding.** `/%63ontrol/state` does reach `/control/state`, while `new URL(c.req.url).pathname` keeps the raw spelling — so the plane was describing a route it had just served as an unknown path. Fixed by reading `c.req.path` (the decoded path the matcher itself used) for the CORS method list, the 404/405 split, and the error log. Test asserts an encoded spelling gets both a 200 with headers and a 204 preflight.

**Registration.** Rather than canonicalising, percent-encoded route keys are now refused: requests are decoded before matching, so such a key is unreachable by any ordinary request *and* invisible to the overlap check — the same dead-code shape as an explicit HEAD route, which this PR already refuses. `?`/`#` in a key are refused for the same reason (they never reach the matcher; a registered `/x?y` answers nothing, confirmed).

One more thing this turned up, adjacent to your point: `bodyAlreadyRead` treated `transfer-encoding` as evidence of content. An empty chunked body is legal, arrives already drained, and is indistinguishable from an eaten one — a hand-written `Transfer-Encoding: chunked` + `0\r\n\r\n` request got a 500. The guard now answers only when certain (a positive `content-length`), since being wrong there makes the diagnostic itself the outage. Regression test drives a raw socket.

kid7st (commented):
This change makes `/control` a prefix-owning sub-application and moves route matching, collision checks, and the Node/Fetch adapter behind a literal-route plus `PrefixMount` model.

**PR description — “One rule, one answer” / “Behaviour changes”**

Problem: The PR description still presents prefixes as a `/*` spelling inside `Routes` and refers to the now-nonexistent `routePathsOverlap`. The final implementation’s core semantic is different: `Routes` accepts literal paths only, while prefix ownership is represented separately by `PrefixMount`. As written, the squash commit’s durable description will preserve the wrong public API model.

Suggested direction: Rewrite those sections and the behaviour table to describe the final `PrefixMount` model; remove the `/*` route-key language and replace the stale `routePathsOverlap` references with the actual `pathUnderPrefix` / `routeKeysConflict` responsibilities.

**`docs/channels.md` “Routes”; `docs/channel-development.md` route-key section**

Problem: Both primary channel-authoring guides still define a valid key as arbitrary `METHOD /path`, but the new boundary rejects `HEAD`, `ALL`, Fetch-forbidden methods, patterns, percent-encoding, query/fragment syntax, and URL-normalized paths. These docs now describe a larger accepted input language than the implementation, so an author following them can hit an unexplained startup failure.

Suggested direction: State the literal-path grammar, method restrictions, and implicit `HEAD`-from-`GET` behavior in both places, or make both sections point to the single complete rule in `docs/api-reference.md`.

src/channels/control.ts
```diff
@@ -40,6 +45,85 @@ export interface WireEvent {
...
+    if (methods.has("GET")) methods.add("HEAD");
+    return [...methods, "OPTIONS"].join(", ");
+  };
+
+  return async (req) => {
+    const path = new URL(req.url).pathname;
+    const known = methodsByPath.has(path);
+    const answer = async (): Promise<Response> => {
+      // A preflight carries no token — that is its entire purpose — so it is answered before auth.
+      if (known && req.method === "OPTIONS") return new Response(null, { status: 204 });
```
kid7st: Problem: The plane promises that browser clients can read its 404/405 responses, but preflight prevents those responses from being requested. An unknown path answers its `OPTIONS` request with 404, and a known path under an unsupported method returns 204 without that requested method in `Access-Control-Allow-Methods`. Because `Authorization` forces preflight, the browser rejects both cases before sending the real request and exposes the same opaque network error this PR is meant to remove. The tests at `test/control-http.test.ts:196-200` use Node `fetch`, which does not enforce that gate, so they do not verify their stated browser scenario.

Suggested direction: Let preflight proceed for every request under the owned `/control` prefix and allow the requested method (or the plane-wide finite method set), then keep the actual request responsible for returning CORS-bearing 404/405. Add preflight assertions for an unknown path and an unsupported method before asserting that the resulting error response is readable.

kid7st (commented):
This change makes `/control` a prefix-owning sub-application and moves literal-route dispatch, CORS, collision checks, and the Node/Fetch bridge behind that model.

**PR description — “Hono replaces the hand-written HTTP layer”, “One rule, one answer”, and the refused-route table**

Problem: The durable PR description still describes superseded implementations. The final code uses a custom `Map` router rather than Hono routing, treats percent-encoded spellings as distinct paths, permits explicit `HEAD`, and no longer rejects `ALL`, `CONNECT`/`TRACE`/`TRACK`, or other non-token methods in `assertRouteKey`. It also no longer has the stated `c.req.path` decoded-path fix. These are core API semantics, so the squash commit message would document behavior the merged code does not have.

Suggested direction: Rewrite those sections from the final implementation: `router` is the literal-path/prefix dispatcher, `@hono/node-server` is only the Node/Fetch bridge, percent-encoded paths remain encoded, explicit `HEAD` is supported, and route validation rejects only the spellings currently enforced by `assertRouteKey`.

src/cli/serve.ts
```diff
@@ -284,84 +305,45 @@ export function readyAddressLines(host: string | undefined, boundPort: number, b
...
   const stop = (exitCode: number): void => {
     if (stopping) return;
     stopping = true;
-    abort.abort();
+    // Bounded: shutdown must not hang on a channel that will not close, so the deadline fires
+    // regardless. No drain — an in-flight turn is cut, which is the existing contract.
     const deadline = setTimeout(() => process.exit(exitCode), 1_000);
-    void hosted
-      .close()
+    void Promise.resolve(hooks.onShutdown?.())
```
kid7st: Problem: `stop()` waits for `hooks.onShutdown()` before calling `hosted.close()`. `surface.close()` waits for every long connection to close, and during that wait the listening socket still accepts new requests; `closeAllConnections()` only cuts connections that already exist at that instant. The process can therefore accept and dispatch new work after its scheduler/channels have begun shutting down, for up to the one-second deadline. This contradicts the stated no-drain shutdown semantics and regresses the previous stop-accepting-first behavior.

Direction: call `hosted.close()` synchronously when `stop()` begins, then force-close existing connections, and await both server close and `onShutdown` under the existing deadline.

src/cli/serve.ts
```diff
@@ -265,6 +271,21 @@ export function assertTunnelBindable(host: string | undefined, tunnel: boolean,
...
  * still needs one to dial, which is what `clientHost` gives (loopback for a wildcard, itself otherwise).
  */
+/**
+ * The "we are serving" report: the supervisor message `dev`'s watcher waits for, the addresses, and
+ * what mounted. One function because both commands must say the same thing at the same moment —
+ * after readiness, never at socket bind.
+ */
+export function reportServing(surface: AgentSurface, host: string | undefined, boundPort: number): void {
+  process.send?.({ type: "ready", port: boundPort, routeChannels: surface.channels.routes });
+  const builtinInvoke = Object.keys(surface.routes).some((key) => key === "POST /invoke");
```
kid7st: Problem: the presence of `POST /invoke` does not mean this is the built-in invoke fallback. A channel may legally author that exact path with a different request protocol; this inference then prints the built-in JSON/SSE curl example for a handler that does not implement it. The old path used `routesFor().builtinInvoke`, but that semantic fact is discarded while constructing `AgentSurface`.

Direction: preserve `routed.builtinInvoke` on the surface or in internal reporting metadata and pass it to `readyAddressLines()` directly; do not infer route ownership from the path.

kid7st (commented):
This change makes `/control` a prefix-owning plane with one CORS/totality boundary, and introduces `openAgentSurface` as the shared directory-to-service assembly used by embedders and the CLI.

src/channels/serve.ts
```diff
@@ -0,0 +1,276 @@
...
+  if (method === "") throw new Error(describe('a leading space is not a method — write "/path" for any method'));
+  // A method is an HTTP token (RFC 9110). A non-token cannot arrive from any client, unlike the
+  // merely unusual ones above. Checked on what the author WROTE: upper-casing first would let `ß`
+  // through as `SS`, a token no client can send under the name the key declares.
+  const written = key.includes(" ") ? key.slice(0, key.indexOf(" ")) : undefined;
+  if (written !== undefined && written !== "" && !/^[A-Za-z0-9!#$%&'*+\-.^_`|~]+$/.test(written)) {
+    throw new Error(describe(`"${written}" is not a valid HTTP method`));
+  }
+  if (!path.startsWith("/")) throw new Error(describe('must start with "/"'));
+  const pattern = [":", "*"].find((ch) => path.includes(ch));
```
kid7st: 问题：`Routes` 被定义为 literal path，且校验注释声称只拒绝永远无法命中的 key，但这里把 `:` 和 `*` 一律当作 pattern 拒绝。它们都是能原样到达并被 `Map` 精确命中的合法路径字符，例如请求 `/files/*` 的 `pathname` 就是 `/files/*`。因此校验删除了 literal-path 概念本应表达的合法能力，且与自身“只拒绝 unreachable key”的边界冲突。

建议方向：删除这两个字符的拒绝规则；文档只说明它们没有参数/通配语义，并增加一个精确命中含 `:`/`*` 路径的测试。

src/surface.ts
```diff
@@ -0,0 +1,256 @@
...
+  let closing: Promise<void> | undefined;
+  const close = (): Promise<void> => {
+    // Awaits the connections rather than only signalling them: `close()` promises they are stopped,
+    // and a caller tearing down a test or a request-scoped surface needs that to be true on return.
+    closing ??= (async () => {
+      abort.abort();
+      scheduled.stop();
+      options.signal?.removeEventListener("abort", onAbort);
+      unannounce?.(); // a stale discovery file would point a client at a dead port
+      await Promise.allSettled(runs.map((run) => run.closed));
```
kid7st: 问题：`AgentSurface.close()` 的公开语义是“返回时 long connections 已停止”，而 `LongConnection.closed` 用 rejection 表达 terminal failure；这里用 `Promise.allSettled` 吞掉 rejection，导致关闭失败时 `close()` 仍成功返回。已用一个在 abort 后以 `Error("shutdown failed")` reject 的 channel 验证，实际输出是 `close resolved`。关闭责任被静默丢弃，调用方无法判断资源是否真的停止。

建议方向：让 `close()` await `Promise.all(runs.map(run => run.closed))` 并把关闭失败 reject 给调用方；如果需要聚合多个失败，抛出 `AggregateError`，不要把失败转换成成功。

kid7st (commented):
This change makes the control plane a prefix-owning mount and introduces openAgentSurface as the shared channels/control/schedules/long-connection assembly.

PR #357, opened 2026-08-26
test(agentcore): parse the CloudFormation template instead of matching substrings

kid7st:
The AgentCore CloudFormation template is generated as text and was only ever asserted with `toContain`. That cannot see the two ways it breaks worst.

**Indentation is semantic in YAML.** Shifting `Code:` three spaces left — which reparents it out of the Lambda — left all 85 existing assertions green.

**A reference is resolved by CloudFormation, not by us.** A typo'd `!Ref`, `!GetAtt`, or `${...}` inside `!Sub` passes every substring check and fails at deploy: the most expensive moment to learn, after the image is built and pushed.

## What the new tests answer

| Question | How |
|---|---|
| Does it parse — in every topology? | `yaml` with the CFN short tags, across 6 flag combinations (webhooks, schedules, selfSchedule, secrets) |
| Is each property under the resource that owns it? | `Code` on the Lambda, `FilesystemConfigurations` on the Runtime, and neither on the other |
| Does every reference resolve? | Walks the whole template — `Ref`, `GetAtt` and the ones hidden inside `Fn::Sub` strings — against resources + parameters + AWS pseudo-parameters |
| Do secrets reach the agent? | A `NoEcho` parameter that the Runtime never reads means the deployment takes the secret and the agent never sees it |

## Verification

Each test was checked the only way that counts — break the template, confirm it goes red, and confirm the existing suite does not:

| Injected fault | New tests | Existing 85 |
|---|---|---|
| `Code:` indented three spaces left | 5 failed | **all green** |
| `!Ref StateBucket` → `StateBuckett` | 1 failed | **all green** |
| `!GetAtt ExecutionRole.Arn` → `ExecutionRoleTypo` | 1 failed | **all green** |
| `!Sub "${Runtime.AgentRuntimeArn}"` → `RuntimeTypo` | 1 failed | — |
| Secret parameter declared, never wired into the runtime | 1 failed | — |
| `Outputs` referencing a renamed resource | 1 failed | — |

The last three come from review rounds on this PR: each caught a case where the test name promised more than the assertions delivered (references only walked `Resources`; "wires them into the runtime" only checked the parameter existed).

## The parser had to be made strict first

`yaml` does not throw on a tag it cannot resolve — it warns and returns the bare node. So `!Sub [template, vars]` (a valid CloudFormation form) arrives as a plain array with `Fn::Sub` gone, and any tag not in the custom set (`!If`, `!Join`) silently becomes its scalar. Either would leave "resolves every reference" green while checking nothing.

`parseTemplate` now fails on any parser warning or error, so a template using a form this test does not handle produces a red test naming it. Verified both ways: switching one `!GetAtt` to the two-argument `!Sub` → red; introducing `!If` → red. Handling those forms is deliberately left until a template needs them — the guard turns their arrival into a failure rather than silent loss of coverage.

Existing substring assertions stay — they check content this does not.

`yaml` is a devDependency, used only here.


kid7st (commented):
This change introduces structural CloudFormation template tests: the parsed template is meant to preserve resource ownership and guarantee that every generated reference resolves.

test/deploy-agentcore-template.test.ts
```diff
@@ -0,0 +1,172 @@
...
+import {
+  type AgentcorePlanInput,
+  type ScheduleFact,
+  MOUNT,
+  planAgentcoreDeploy,
+  scheduleResourceName,
+} from "../src/deploy/agentcore/plan.ts";
+
+/** CloudFormation's short tags are not standard YAML — resolve them to their long forms. */
+const CFN_TAGS = [
```
kid7st, line 31: Problem: `CFN_TAGS` only resolves scalar tags. CloudFormation’s valid two-argument `!Sub` form is a sequence (`!Sub [template, variables]` or its block equivalent); `yaml` emits an unresolved-tag warning, returns the bare array, and does not throw. `references()` then sees neither `Fn::Sub` nor its resource placeholders, so the test named “resolves every reference” can pass after a valid template change has silently removed references from the checker.

Suggested direction: Parse with `parseDocument`, fail on parser warnings/errors, and add sequence handling for `!Sub`. When checking that form, exclude placeholders supplied by its variable map and recurse into the map values for their real references.

PR #358, opened 2026-08-26
refactor(agentcore): give the posture its own assembly, not six branches in start

kid7st:
`start` carried the AgentCore posture as six inline branches — `mounted ? … : …`, a second control mount, a second scheduler start, a hand-composed router, a separate `unannounce`, and the collision rule invoked a second time by hand. #355 gave every other deployment one assembly (`mountAgentService`) and left this as "the one exception", which is where the branches came from.

They are now one branch, at the point where the posture is decided:

```ts
const service = await (isAgentcoreRuntime()
  ? mountAgentcoreService(opened, { wrapAgent, control, onStateReady })
  : mountAgentService(opened, { wrapAgent, control, closeTimeoutMs, onChannelClosed }));
// everything below is common
```

## Not one assembly — one product

AgentCore genuinely assembles differently, and the differences are not a flag: the adapter is the surface (`POST /invocations`), channels are discovered lazily because **the state mount at boot is pre-restore** (eager discovery would cache that emptiness and then clobber the restored files), the clock is external, resident connections cannot survive scale-to-zero, and shutdown has no drain. Folding those into `mountAgentService` would only move the `if` from the CLI into the library.

What they share is the *product*: both return an `AgentService`, so `serve`, the startup report, the tunnel and the shutdown hook stop asking which host they are on.

| | before | after |
|---|---|---|
| `start.ts` | 283 lines, 21 agentcore mentions | 206, 7 |
| `cli/serve.ts` | 251 lines, 11 mentions | 189, **0** |

`mountAgentcore` moved out of `cli/serve.ts` with it: an adapter mount is agentcore machinery, not a CLI concern.

## Ownership, re-derived rather than patched

Review found the assembly claiming a resource it cannot own: `setWakeupsSink` is a **process-global singleton**, and a service has `close()`. Three rounds of fixes each simulated multi-owner semantics the singleton does not have — compare-and-clear, then rollback-on-failure, then restore-the-previous.

The fix was to stop: this process *is* the AgentCore deployment, so the process entry installs the sink, exactly as before this refactor. The assembly takes `onStateReady` and owns nothing global. `wakeups.ts` ends up untouched.

The alternative — scoping the sink per state root — is the right long-term shape and is out of scope here.

## A leak the close() contract exposed

Giving the posture a `close()` made a pre-existing leak observable: `mountAgentcore` registered a process-global `onIdle` listener and discarded the unsubscribe `busy.ts` hands back. A closed adapter kept uploading its state snapshot on every later idle edge — including work belonging to whatever mounted next.

Same shape as the wake sink, opposite conclusion, because the owner differs: `onIdle` returns a disposer precisely because it is meant to be cancelled, and the listener belongs to the adapter, which belongs to the service. It now takes an `AbortSignal` that `close()` fires. The wake sink stays with the process because the process — not the service — is what "this deployment" means.

## Verification

New `test/agentcore-service.test.ts` (5), each checked by reverting the behaviour it covers:

| Reverted | Result |
|---|---|
| Control plane not mounted | red |
| Channel routes exposed as the service surface | red |
| Channels reported eagerly (the pre-restore emptiness) | red |
| `routes` reported as `{}` (startup line claiming nothing is served) | red |
| The idle listener left registered after close | red |

Explicitly **not** claimed: that `close()` clears the scheduler's timers. Fake timers would have to be installed before the assembly, which deadlocks its filesystem IO — noted at the call site rather than left reading as covered.

Behaviour equivalence checked where the merge changed a path: `builtinInvoke: false` keeps the curl example out of the startup line, and `onShutdown` (SIGINT/SIGTERM) now stops the scheduler that previously had its own two handlers.


src/channels/agentcore-service.ts
```diff
@@ -0,0 +1,172 @@
...
+    workspace,
+    // Unknown at boot by design — a channel list here would be the pre-restore emptiness.
+    channels: { routes: [], longConnections: [], builtinInvoke: false },
+    schedules: scheduled.schedules,
+    ready: Promise.resolve(), // nothing to open: no port of our own, no resident connections
+    ...(withControl.control ? { control: withControl.control } : {}),
+    announce(boundPort) {
+      unannounce = withControl.announce(boundPort);
+    },
+    async close() {
```
kid7st, line 111: `close()` stops the scheduler and removes the discovery file, but `mountAgentcore()` also registers a process-global `onIdle` listener in `src/channels/agentcore.ts:191` and discards its unsubscribe function. A closed service therefore remains retained and still uploads its state snapshot whenever later process work reaches idle; mounting another service in the same process can trigger stale cross-service state writes. I reproduced one PUT after restore → `close()` → a new idle edge. Make the adapter lifecycle expose a disposer or accept an `AbortSignal`, unregister the idle listener from `AgentService.close()`, and test that an idle edge after close performs no PUT.

kid7st (commented):
This change introduces `mountAgentcoreService`: an AgentCore-specific assembly that preserves lazy post-restore channel construction while presenting the same `AgentService` lifecycle to `start`.

PR #359, opened 2026-08-26
refactor(paths): one atomic write, and the permission it was silently missing

kid7st:
Four copies of "write a temp, then rename" had drifted apart:

| | temp name | permissions |
|---|---|---|
| `schedule/state.ts` | `<path>.tmp` | none |
| `channels/state.ts` | `<path>.tmp` | none |
| `service.ts` (control.json) | `<path>.<pid>.tmp` | `mode` + chmod after rename |
| `slack/bot-auth.ts` | `<path>.tmp` | `mode` + chmod on both |

The first two were byte-identical. `schedule/state.ts` even carried a note asking for exactly this extraction "when a third consumer appears" — there were four.

All four now call one `writeFileAtomic(path, data, mode?)` in `src/atomic-write.ts`. Slack's async
onboarding state keeps its own path — this is deliberately not an async API, and converting that
caller is a separate question from de-duplicating these four.

## What writing the test found

**The permission was not actually guaranteed.** `writeFileSync`'s `mode` option applies only when it CREATES the file, so a temp left behind by a crashed writer keeps its old permissions and the rename publishes them — verified: 0644 survives a `{ mode: 0o600 }` write. `control.json` IS the control-plane credential, and nothing tested its mode. It does now.

**The cleanup could bury the error that caused it.** `rmSync(tmp, { force: true })` only forgives ENOENT; when the temp is a *directory* it throws `ERR_FS_EISDIR` and replaces the write failure that explains what went wrong. That path is exercised by five existing channel tests, which inject failure by occupying the temp path with a directory. The test now asserts the surfacing error is the write (`syscall: "open"`), not the cleanup (`syscall: "rm"`).

**The two chmods were redundant with each other.** Reverting either alone left the suite green. Kept the one on the temp, before the rename: the final path is then never observable with the wrong permissions, which a chmod after the rename cannot promise.

## Deliberately not changed

The temp name stays `<path>.tmp` rather than the pid-carrying form two of the four used. The race it
addresses needs two processes writing one state root, which does not occur here: a deployment runs
one container, and `dev`'s supervisor respawns its worker only after the old one has EXITED
(`dev-supervisor.ts:125`) — these writes being synchronous, an exited process has none in flight.
Against that, the fixed name is the seam five existing channel tests use to inject a write failure.
The assumption is stated at the function, so a future second writer makes it a red flag rather than
a silent hazard.

## Verification

Each behaviour reverted independently:

| Reverted | Result |
|---|---|
| Non-atomic (write straight to the target) | 51 failed |
| The chmod | red |
| Cleanup that lets its own error escape | red |
| Parent-directory creation | red |

### DIFF
--- a/src/atomic-write.ts
+++ b/src/atomic-write.ts
@@ -1,0 +1,37 @@
+/**
+ * One spelling of "a reader sees the whole file or none of it", after four copies of it drifted
+ * apart: two identical, two with different temp names and different permission handling.
+ */
+import { chmodSync, mkdirSync, renameSync, rmSync, writeFileSync } from "node:fs";
+import { dirname } from "node:path";
+
+/**
+ * Write a file so a reader sees the whole thing or nothing: same-directory temp, then rename.
+ *
+ * The temp name is fixed (`<path>.tmp`), which is safe under this codebase's single-process state
+ * assumption and is also how several tests inject a write failure — occupying that path with a
+ * directory. `mode` is applied to the temp first, so the content is never briefly world-readable.
+ *
+ * The `chmod` is NOT redundant with the `mode` option: `writeFileSync` applies `mode` only when it
+ * CREATES the file, so a temp left behind by a crashed writer keeps its old, possibly loose
+ * permissions and the rename publishes them (verified: 0644 survives a `{ mode: 0o600 }` write).
+ * It runs on the temp, before the rename — the final path is then never observable with the wrong
+ * permissions, which a chmod after the rename cannot promise.
+ */
+export function writeFileAtomic(path: string, data: string, mode?: number): void {
+  mkdirSync(dirname(path), { recursive: true });
+  const tmp = `${path}.tmp`;
+  try {
+    writeFileSync(tmp, data, mode === undefined ? undefined : { mode });
+    if (mode !== undefined) chmodSync(tmp, mode);
+    renameSync(tmp, path);
+  } catch (error) {
+    try {
+      rmSync(tmp, { force: true });
+    } catch {
+      // `force` only forgives ENOENT: if the temp is a directory, rmSync throws its own error and
+      // would replace the write failure that actually explains what went wrong.
+    }
+    throw error;
+  }
+}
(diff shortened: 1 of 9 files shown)


src/atomic-write.ts
```diff
@@ -0,0 +1,37 @@
+/**
+ * One spelling of "a reader sees the whole file or none of it", after four copies of it drifted
```
kid7st: Problem: `writeFileAtomic` is presented as the single implementation extracted from four copies, but `src/channels/slack/bot-auth.ts`—one of the four consumers explicitly named in the PR description—still carries its own `writeFileSync`/`chmodSync`/`renameSync`/`rmSync` sequence. `src/channels/slack/onboarding-state.ts` also retains the asynchronous variant. The core ‘one spelling’ semantic therefore does not hold, and Slack credential writes do not receive this change's original-error preservation.

Direction: replace the duplicate synchronous `bot-auth.ts` save path with `writeFileAtomic`, and narrow this module's description to synchronous runtime writes so the intentionally asynchronous onboarding path is not claimed by the abstraction.

kid7st (commented):
This change defines `writeFileAtomic(path, data, mode?)` as the shared synchronous same-directory temp-write-and-rename primitive, including pre-publication permission enforcement and original-error preservation.

src/atomic-write.ts
kid7st: Split this: the duplicate consumer was real and is fixed; the unique-temp-name part is declined with a reason.

**Fixed** — `bot-auth.ts` now goes through `writeFileAtomic`, so all four synchronous consumers share one path and Slack's credential write gets the original-error preservation. The module's description narrowed to synchronous writes, leaving the async onboarding path explicitly outside the abstraction.

**Declined** — the fixed `<path>.tmp`. The race needs two processes writing one state root, and that does not occur: a deployment runs one container, and `dev`'s supervisor respawns its worker only after the old one has EXITED (`dev-supervisor.ts:125`) — these writes being synchronous, an exited process has none in flight. Against that, the fixed name is the seam five existing channel tests use to inject a write failure. Trading a working test seam for an unreachable race is a net loss.

The assumption is now stated at the function with the reasoning, so a future second writer makes this a red flag rather than a silent hazard.

PR #378, opened 2026-08-28
test(pi): check chat against serving's assembly, not against hard-coded values

kid7st:
One definition, two consumption shapes: the served `Agent` (`dev`/`start`/`invoke`) and chat's
resident `AgentSessionRuntime`. The split between them is deliberate and already documented — see
`AgentAssembly`'s doc in `open.ts` ("the agent assembly FRONT HALF — everything independent of how
pi consumes the definition … definition loading and prompt assembly stay per-consumer") and the
FIDELITY block heading `session-builder.ts`, which lists prompt / skills / tools / models / auth and
how each is injected to match serving.

This PR does not change that design. It adds one assertion the existing tests cannot make.

## What was already covered

`session-builder.test.ts` checks all five FIDELITY items — the config model, `thinkingLevel`
(already labelled `(fidelity)`), the tool surface, deferral emulation, and the credential file that
must not be `~/.pi`. Those tests assert chat's **absolute** values: the active set equals
`["bash","edit","find","grep","ls","read","write"]`, the level is `"high"`.

## The gap that leaves

An absolute assertion cannot notice that the OTHER side moved. Add a capability to serving — a new
built-in tool, a new config field reaching the assembly — and chat's tests still pass while chat
silently lacks it. That is the shape of the risk the history shows: of the last 14 commits touching
`session-builder.ts`, 13 changed two or more assembly files and 10 changed both `create.ts` and
`session-builder.ts`. Every capability so far was wired into both consumers by hand.

So these two tests derive the expectation from the SERVING assembly and check chat against it:
`served.toolNames + served.deferredToolNames` must all be mounted in chat, `served.modelSpec` must
be the model chat resolved, and a credential written where serving resolves `authPath` must be
visible to the hub chat ends up with. Nothing is hard-coded that serving could move.

## Verification

Each side broken once:

```
chat stops honouring config.thinkingLevel   → AssertionError: expected 'xhigh' to be 'high'
chat's deferral emulation disabled          → AssertionError: expected [...] to not include 'lazy'
```

1538 tests pass.

## One unrelated line, riding along

`roundCursor` (`src/cli/commands/attach.ts`) drops its `export`: the only caller is `attachRound`'s
own retry loop in the same file. It surfaced in the same scan; a public surface should have a
reason, and one without it is what nobody dares delete later. Nothing else in the PR touches
production code.

### DIFF
--- a/src/cli/commands/attach.ts
+++ b/src/cli/commands/attach.ts
@@ -652,7 +652,7 @@ function withCursor(error: unknown, cursor: string | undefined): Error {
 }
 
 /** The cursor a failed round reached, if it recorded one. */
-export function roundCursor(error: unknown): string | undefined {
+function roundCursor(error: unknown): string | undefined {
   return (error as { attachCursor?: string }).attachCursor;
 }
 
--- a/test/assembly-fidelity.test.ts
+++ b/test/assembly-fidelity.test.ts
@@ -1,22 +1,17 @@
 /**
  * ONE definition, two consumption shapes: the served `Agent` (dev/start/invoke) and chat's resident
- * `AgentSessionRuntime`. They share the front half (`resolveAgentAssembly` — placement, config,
- * model spec, tools, auth) and diverge after it ON PURPOSE: serving re-reads the definition per
- * invoke and assembles the whole prompt itself, chat takes a startup snapshot and lets pi append
- * skills and env. That divergence is the design; what must NOT diverge is what the definition SAYS.
+ * `AgentSessionRuntime`. The split is deliberate and documented where it is made — `AgentAssembly`
+ * in open.ts marks the shared front half, and session-builder.ts's FIDELITY block lists what chat
+ * injects to match serving: prompt, skills, tools, models, auth.
  *
- * This file is the check on that. It exists because the divergence has bitten once already — chat
- * authenticated against the machine-global `~/.pi` while serving used the project's own credentials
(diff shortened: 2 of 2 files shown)


kid7st:
Closing. The test is not wrong, but the increment does not carry a file.

`session-builder.test.ts` already covers all five FIDELITY items — the config
model, `thinkingLevel` (its case is even labelled `(fidelity)`), the tool
surface, deferral emulation, and the credential file that must not be `~/.pi`.
What this adds is deriving the expectation FROM the serving assembly instead of
hard-coding it, which only pays off in one scenario: a capability wired into
serving alone, where chat's absolute assertions stay true. Real, but narrow
enough that eight overlapping tests plus a new file is the wrong trade.

Worth recording, since the PR body argued otherwise for a while: the split
between the two paths is deliberate and already documented where it is made —
`AgentAssembly` in `open.ts` marks the shared front half and says definition
loading and prompt assembly stay per-consumer; `session-builder.ts` opens with a
FIDELITY block listing what chat injects to match serving. I reached the opposite
conclusion twice before reading either, by counting call sites of the same
function names instead of comparing what they were called WITH
(`assembleSystemPrompt` takes different arguments on each side, for a stated
reason). The answer was in two file headers the whole time.

The one line here that stands on its own — `roundCursor` dropping an `export`
it does not need — is left for whoever next touches `attach.ts`.

PR #385, opened 2026-08-29
fix(feishu): a signature says who, not when — bound the replay window

kid7st:
## Summary

A Feishu/Lark webhook signature was valid forever.

`verifySignature` hashes `timestamp + nonce + encryptKey + rawBody` and compares constant-time, but
nothing checked that the timestamp was recent — so a captured `{encrypt: …}` body plus its three
`x-lark-*` headers stayed a valid request indefinitely. The only thing standing between that and a
re-run turn (a re-sent message, a re-fired tool) was the `seen` ring, which is bounded at 2000 ids
and rolls over. The Slack ingress in this same repo has had a ±5 minute window since it landed; the
asymmetry was not deliberate and was not written down anywhere.

`signatureFresh` now bounds it with the same ±5 minutes. It is a separate predicate rather than a
branch inside `verifySignature` because the two failures need two diagnoses: a wrong encrypt key and
a clock that has drifted are different things to go fix, and one message covering both sends the
operator after the wrong one. The window is two-sided for the same reason — a skewed clock runs
either way, and a future timestamp is exactly as unverifiable as an ancient one.

Unaffected: `url_verification` (the platform sends it without signature headers, and that exception
is unchanged) and the plaintext no-Encrypt-Key mode (no signature to age).

## Validation

- [x] `npm run lint`
- [x] `npm run typecheck`
- [x] `npm test` (114 files, 1548 tests)

The ingress test used a hardcoded `"170"` timestamp, which is why nothing caught this; it now signs
with a live one and adds a correctly-signed-but-6-minutes-old request that must be refused. Reverting
`feishu.ts` alone turns that case red. `signatureFresh` gets its own unit coverage for both window
edges and for non-integer input.

## Checklist

- [x] Docs were updated when behavior or public APIs changed (`docs/feishu.md` security section)

### DIFF
--- a/docs/feishu.md
+++ b/docs/feishu.md
@@ -189,7 +189,7 @@ and resolves `closed`. There is deliberately no second public `close()` path. Se
 WebSocket authentication happens once while establishing the official-SDK connection. Webhook has two
 security modes, decided by the console's Encrypt Key setting and mirrored by `encryptKey`:
 
-- **Encrypt Key set (recommended):** ordinary events arrive AES-256-CBC encrypted with `X-Lark-Signature` headers. The channel verifies the signature over the raw body, decrypts, and **refuses plaintext events entirely** — accepting both would let a forger skip the stronger check.
+- **Encrypt Key set (recommended):** ordinary events arrive AES-256-CBC encrypted with `X-Lark-Signature` headers. The channel verifies the signature over the raw body, decrypts, and **refuses plaintext events entirely** — accepting both would let a forger skip the stronger check. A signature stays valid forever, so `X-Lark-Request-Timestamp` must also be within ±5 minutes of the server's clock; outside it the request is refused as a replay. Keep the host's clock in NTP sync.
 - **No Encrypt Key:** events arrive in plaintext and are authenticated by the Verification Token (constant-time compare).
 
 Request URL verification is the platform-documented narrow exception to event signatures. With an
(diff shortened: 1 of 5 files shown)


kid7st:
Closing without merging: this change bounds replay with the wrong primitive.

Feishu makes no promise about timestamp freshness — the platform documents no replay window, the official \`larksuite/node-sdk\` verifies the signature hash and never reads the clock, and nothing states whether a re-push re-signs. Any window value here is therefore a guess, and the cost is asymmetric: too tight silently 401s legitimate retries (the platform re-pushes a failed event at 15s / 5m / 1h / 6h), too loose stops nothing an attacker holding a signed ciphertext could not already do.

What the platform does specify is idempotency: dedupe repeats over the ~7.5h re-push span by event identity. The channel already does exactly that (`seen` ring keyed on `message_id`, which is stable across re-pushes of the same message), so this layer was defending a scenario with no evidence behind it while adding a real failure mode — a host clock drifting past the window rejects every event with nothing but a `log.warn`.

Two facts about the idempotency layer, recorded here rather than acted on, since neither has evidence demanding a change today:
- `seen` is capped at 2000 ids. Entries age out inside the 7.5h re-push span only above a sustained ~4.4 messages/minute; past that, a late 4th re-push can re-run a turn.
- If that window ever needs closing, the fix belongs in the dedup layer's retention, not in a clock check.

Issue #395, opened 2026-08-30
Verify Slack file-download host allowlist against a live workspace


kid7st:
`src/channels/slack/slack-api.ts` — `trustedDownloadUrl` allows `slack-files.com`, `slack.com`,
`slack-edge.com` (and subdomains) or the configured origin, and re-checks the host on every manual
redirect hop. The shape is right and load-bearing: the bearer token is re-sent on each hop, so the
allowlist is what stops the token from following a redirect off Slack.

The open question is whether Slack's `url_private` ever redirects file bytes to a non-Slack CDN host.
If it does, every affected file attachment fails with `refusing non-Slack file URL host …` — a
visible failure, but one the operator cannot fix.

This cannot be answered from the code or the docs. It needs a live workspace:

- share a file larger than ~1 MB in a channel the agent is in (the size where a CDN redirect is most
  likely), plus one small file as a control,
- run the agent against that workspace and check whether the download completes or logs
  `refusing non-Slack file URL host`,
- if it is refused, record the actual host from the log before changing anything.

Do NOT widen the allowlist speculatively — a guess here weakens the only thing keeping the bearer
token on Slack hosts.

From the round-2 full-repo review (finding C7, Info).

kid7st:
Still open — a live workspace is the only thing that answers it. Two things done around it.

**Indirect evidence, not an answer.** Public Slack-bot downloaders converge on the same three suffixes
we allow (`slack.com`, `slack-edge.com`, `slack-files.com`), and the field failures they report are the
*opposite* shape of the one feared here: apps that **drop** the bearer on the CDN hop get an HTML login
page back ([openclaw#63523](https://github.com/openclaw/openclaw/issues/63523),
[bolt-js#2585](https://github.com/slackapi/bolt-js/issues/2585)). No public report of `url_private`
redirecting to a host outside those three. Convergent, but nobody's log is our log — the allowlist stays
as it is.

**The one-command version of the check**, for whoever has a workspace (large file + a small control):

```sh
curl -sS -L -D - -o /dev/null -H "Authorization: Bearer $SLACK_BOT_TOKEN" '<url_private_download>' \
  | grep -iE '^(HTTP/|location:)'
```

Every `location:` host in the chain answers this issue. If one is off-Slack, paste it here before
anything is widened — and note that a widened allowlist alone would not be the fix: a presigned CDN URL
can reject the `Authorization` header we keep re-sending, so the hop would need the header dropped, not
just the host allowed.

[...]

PR #398, opened 2026-08-30
test(slack): cover the redirect hop of the file-download host guard

kid7st:
The allowlist in `trustedDownloadUrl` had a test for the *first* URL only. Its load-bearing half — the
manual redirect hop, where the bearer token is re-sent — had none: the guard could stop refusing
off-Slack hosts on redirect, or stop carrying the token to a Slack CDN host, with every test green.

One test now covers both hops: a `302` to `a.slack-edge.com` is followed with `Authorization` intact,
and a `302` to `cdn.example` is refused before the second fetch. Verified it fails without the guard
(neutering the host check turns it red).

No allowlist change — #395 (does `url_private` ever redirect off Slack?) still needs a live workspace.


kid7st:
Superseded by #399, which deletes the guard this test was written for.

The test asserted that the bearer token survives a redirect to a Slack CDN host — exactly the behaviour
#399 hands back to the platform. Keeping it would pin a decision we concluded was never ours to make.

PR #399, opened 2026-08-30
refactor(slack): file transfer trusts Slack, so stop policing its URLs

### LINKED ISSUE #395: Verify Slack file-download host allowlist against a live workspace

kid7st:
Closes #395.

## The check could not do its job

`trustedDownloadUrl` allowed `slack.com` / `slack-files.com` / `slack-edge.com` and re-checked the host
on every manual redirect hop. But every input to that decision comes from one party:

- `url_private` is read from **our own `files.info` call** over TLS (`invoke-turn.ts` takes only the
  `file_id` from the event payload — the URL never comes from the webhook),
- the bearer token is Slack's, issued to us by Slack,
- Slack's own docs define the field as *"require an authorization header of the form:
  `Authorization: Bearer A_VALID_TOKEN`"*.

So the allowlist was second-guessing a party we already trust unconditionally — we hand it the token,
follow its instructions, and feed its content to the model. If Slack ever put a hostile host in that
field, it could equally just use the token itself. The check stopped nobody, and cost a list of CDN
domains that only Slack can keep correct.

## What it did cost

#395 was open because nobody could answer whether `url_private` redirects off Slack. It could not be
answered from code or docs — it needed a live workspace — and if the answer had been yes, every
affected attachment would fail with `refusing non-Slack file URL host …`, a failure **no operator
could fix**. We were carrying an unanswerable question and an unfixable failure mode to run a check
that protected nothing.

Choosing `redirect: "manual"` is what created the need for a list in the first place: fetch already
answers "may this hop carry the token" (it drops `Authorization` cross-origin), and following
redirects by hand threw that answer away.

## Change

- `download()` is one `fetch` with the token and the existing cap/error handling. Redirects go back to
  the platform: where they lead is Slack's call, and whether a hop still carries the token is fetch's.
- The same allowlist is gone from `scaffold/slack-send.ts` (upload URL, bytes sent **without** a token
  — and those bytes are on their way to Slack either way). It matters more there: the scaffold is
  copied into every author's agent directory, so a bad pattern propagates.
- An attachment failure now derives `retryable` from `SlackApiError.status` instead of asserting one
  answer for all of them: a network error, exhausted 429 retries or a Slack 5xx is worth re-sending,
  while `access_denied` reads the same every time and "try again in a moment" is the wrong advice.
- The `fetch-url.ts` scaffold drops its SSRF note. A template's comments should teach fastagent, not
  general web security, and it is copied into the author's own directory.

Sweeping the repo for the same pattern found no third site: telegram splices Slack's counterpart —
`file_path`, a value Telegram chooses — straight into its download URL and checks nothing, and feishu
builds every resource URL itself. Slack was the outlier. What is ours stays untouched: signed ingress
on all five channels, attachment path sanitising, untrusted tool names rendered as code spans, the
control-plane bearer, and snapshot validation on the AgentCore state mount.

## Trade-off

If a Slack CDN hop ever *requires* the token rather than a signed URL, the download 401s. That is
diagnosable from the status and host in the log, and has an obvious fix — unlike guessing which
domains to add to a list.

### DIFF
--- a/src/channels/slack/invoke-turn.ts
+++ b/src/channels/slack/invoke-turn.ts
@@ -106,7 +106,8 @@ export async function* invokeSlackTurn(
   try {
     resolved = await resolveInputs(transport, attachments);
   } catch (error) {
-    yield { type: "failed", details: `could not load Slack attachment: ${String(error)}`, retryable: true };
+    // an unreadable attachment reads the same on every retry, so the user is told to change something
+    yield { type: "failed", details: `could not load Slack attachment: ${String(error)}`, retryable: false };
     return;
   }
   const prompt = { text: `${text}${resolved.promptSuffix}${MARKDOWN_INSTRUCTION}`, images: resolved.images };
--- a/src/channels/slack/invoke-turn.ts
+++ b/src/channels/slack/invoke-turn.ts
@@ -11,7 +11,7 @@ import {
   streamTurnWithBusyRetry,
 } from "../kit/invoke-turn-kit.ts";
 import type { SlackBufferedFileRef } from "./context-buffer.ts";
-import type { DownloadedSlackFile, SlackApi } from "./slack-api.ts";
+import { type DownloadedSlackFile, type SlackApi, SlackApiError } from "./slack-api.ts";
 
 const MARKDOWN_INSTRUCTION =
   "\n\n(Format your reply as standard Markdown. Slack renders it natively. Do not use HTML or Slack control-mention syntax such as <!here>, <!channel>, or <!everyone>.)";
@@ -106,8 +106,11 @@ export async function* invokeSlackTurn(
   try {
     resolved = await resolveInputs(transport, attachments);
   } catch (error) {
-    // an unreadable attachment reads the same on every retry, so the user is told to change something
-    yield { type: "failed", details: `could not load Slack attachment: ${String(error)}`, retryable: false };
+    // transient (network, exhausted 429 retries, Slack 5xx) is worth re-sending; an access or shape
+    // error reads the same every time, and "try again in a moment" is the wrong thing to tell the user
+    const retryable =
+      error instanceof SlackApiError && (error.status === 0 || error.status === 429 || error.status >= 500);
+    yield { type: "failed", details: `could not load Slack attachment: ${String(error)}`, retryable };
     return;
   }
   const prompt = { text: `${text}${resolved.promptSuffix}${MARKDOWN_INSTRUCTION}`, images: resolved.images };
--- a/src/channels/slack/scaffold/slack-send.ts
+++ b/src/channels/slack/scaffold/slack-send.ts
@@ -71,22 +71,6 @@ async function callSlack<T extends { ok?: boolean; error?: string }>(
   }
 }
 
-function trustedUploadUrl(value: string): URL {
-  const url = new URL(value);
-  const host = url.hostname.toLowerCase();
-  const slackHost =
-    host === "slack-files.com" ||
-    host.endsWith(".slack-files.com") ||
-    host === "slack.com" ||
-    host.endsWith(".slack.com") ||
-    host === "slack-edge.com" ||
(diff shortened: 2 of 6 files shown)


kid7st:
Swept the repo for the same pattern — policing a decision that belongs to the other side of a trust
boundary. All 15 `fetch` call sites, by where the URL comes from:

| Source of the URL | Sites | Check? |
|---|---|---|
| Our own `base + path` | feishu ×3, telegram ×2, the three send-tool scaffolds | n/a — we pick the host |
| A third party's API response | slack `url_private`, slack `upload_url`, telegram `file_path` | **only Slack had one** |
| Our own address | `wait-health.ts` | n/a |
| **The model** | `scaffold/templates/tools/fetch-url.ts` | none, by design |

Two things this settles:

**The two sites in this PR were the only ones.** Telegram splices `file_path` — a value Telegram
chooses — straight into the download URL and checks nothing. Feishu builds every resource URL itself.
Slack was the outlier; it now reads like its siblings.

**The one place where the URL is genuinely ours to police has no allowlist, and that is also right.**
In `fetch-url.ts` the URL comes from the model, so it is reachable by whoever can talk to the agent —
the real SSRF surface. It carries a comment telling the author to add an allowlist or block private
ranges *if they serve it publicly*, because only the author knows their exposure. So the repo had it
exactly backwards: a hardcoded list where the counterparty decides, and a judgement call where we do.

What is ours was already in place and stays: signed ingress on all five channels (constant-time
compare + replay window, cap-before-verify), attachment path/name sanitising, untrusted tool
names/args rendered as code spans, the control-plane bearer, snapshot validation on the AgentCore
state mount, and `http.ts` documenting that authentication belongs to the embedding host.

kid7st:
Also dropped the SSRF note from `scaffold/templates/tools/fetch-url.ts`, for the same reason one layer
up: it was the repo's only comment teaching general web security, in a file that gets copied into the
author's own directory.

A template's comments should teach *fastagent* — "the filename is the tool name, `tools/` is
auto-discovered, here is how to run it without a model". SSRF is not fastagent knowledge, and once you
start there is no line: XSS, secret handling, rate limits. The note also prevented nothing; it was a
disclaimer, and the author who needs it will not read it.

The `ponytail:` note on the tag stripper stays — that one marks a real ceiling in *this* code and its
upgrade path, which is the repo's existing convention (6 sites).

kid7st:
Review catch, fixed in `5a1a7a2`: flipping `retryable` to a constant `false` was wrong in the other
direction. `resolveInputs` fails on network errors (`status: 0`), exhausted 429 retries and Slack 5xx
just as much as on `access_denied` — the first three are worth re-sending, and those are the *more
common* production failures. It now reads `SlackApiError.status`, so the advice matches the cause.

Three cases cover the retryable branch (verified they go red if the check is hardcoded back).

Also tightened the `download` comment: where the URL leads is Slack's call, but whether a hop still
carries the token is fetch's, not Slack's.

PR #400, opened 2026-08-30
fix(slack): a rejected download names the host that rejected it

kid7st:
Follow-up to #399. Now that redirects are followed by the platform, the host that answers a download
is no longer always the host we asked for — but a failure only reported the status:

```
slack file download failed: 401 <first 300 chars of the body>
```

An operator reading that cannot tell whether Slack refused the file or a CDN hop refused our token,
and the answer decides what to do next. `response.url` is the last hop after fetch follows redirects
(verified), so naming it costs one line:

```
slack file download from cdn.example failed: 401 expired signature
```

This is not a guess about redirect behaviour and adds no mechanism for one — it makes the failure that
already exists say where it happened, which the repo requires of every failure path. It applies to
every rejected download, not just redirected ones.

### DIFF
--- a/src/channels/slack/slack-api.ts
+++ b/src/channels/slack/slack-api.ts
@@ -261,7 +261,10 @@ export function createSlackApi({ botToken, baseUrl = "https://slack.com/api" }: 
     }
     if (!response.ok) {
       const detail = codePointPrefix(await response.text().catch(() => ""), 300) || "file bytes were rejected";
-      throw new SlackApiError("file download", response.status, detail);
+      // response.url is the LAST hop, so a download that redirected before failing names the host that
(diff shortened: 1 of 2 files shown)


kid7st:
Closing: this is the same reflex as the allowlist it followed, one layer further out.

Using Slack's URL with Slack's token means a rejection is Slack's answer to us, whichever host
produced it — so the operator's next step is the same either way. The response body is already in the
log, and it identifies the source on its own: `{"ok":false,"error":"invalid_auth"}` reads as the API,
an S3 XML error reads as a CDN. Only an empty-body 401 would have gained anything.

The premise was also wrong. I argued the missing host was something #399 amplified; it was not — the
log line is byte-identical before and after, since the allowlist never recorded the host either.

Issue #421, opened 2026-09-01
Tunnel readiness polling is self-defeating: --tunnel registration and `add slack` both stall on a fresh hostname


kid7st:
## What happens

Behind an HTTP proxy, polling a freshly created Cloudflare Quick Tunnel URL from the moment it is printed makes that URL permanently unreachable **for the rest of that Node process**. Waiting quietly and asking once succeeds.

This is the readiness step `--tunnel` depends on: `startCloudflareTunnel` → `waitForHealth(<tunnel URL>/health)` → register the channel's webhook. `waitForHealth` polls every 1–3s starting immediately, which is exactly the pattern that triggers it, so on a proxied machine `fastagent dev --tunnel` and `deploy docker --tunnel` can fail to register with a timeout instead of a reason.

## Evidence

All against the same proxy on the same machine, same tunnel-creation code path:

| Client | Pattern | Result |
|---|---|---|
| `fetch` (global dispatcher from `installProxyFetch`) | poll from t+0 | 65s of `Client network socket disconnected before secure TLS connection was established`, never recovers |
| `undici.request` with a **fresh** `EnvHttpProxyAgent` per attempt | poll from t+0 | 40s+, same failure |
| `fetch` (global dispatcher) | quiet 20s, then **one** request | `200` |
| `fetch` (global dispatcher) | quiet 5s, then one request | `200` |
| `curl` (separate process each time) | poll from t+0 | `200` at t+2s |

Observed in the live probes: `test/live/tunnel.live.test.ts` fails on a proxied machine and passes on CI (direct connection); `test/live/telegram.live.test.ts` fails there too, at ~128s — `waitForHealth`'s 120s budget — making `registerTelegramWebhook` return `failed`.

## What this rules out

- **Not a stale connection pool on one Agent instance.** A brand-new `EnvHttpProxyAgent` per attempt fails identically.
- **Not "the tunnel needs more time".** A quiet 20s wait then a single request succeeds, while 40s of polling does not.
- **Not the proxy refusing the host.** `curl` through the same proxy reaches it at t+2s.

What is left: something process-wide in Node/undici, poisoned by the early failed attempts themselves, that a new dispatcher does not reset and a new process does not inherit. I did not chase it further into undici.

## Reproduce

```js
// node --experimental-strip-types, with HTTPS_PROXY set
import { createServer } from "node:http";
import { startCloudflareTunnel } from "./src/tunnel.ts";
[...]

PR #436, opened 2026-09-03
fix(tunnel): hold a quick-tunnel URL back until its hostname is in DNS

### LINKED ISSUE #435: telegram: setWebhook cannot resolve a fresh Quick Tunnel hostname within the 80s registration budget
## What happens

`--tunnel` + telegram fails to register on a freshly minted Quick Tunnel hostname. The registrar exhausts all 8 attempts (80s) and reports manual registration:

```
ERROR [fastagent] telegram: Telegram could not reach
https://staff-prior-portable-local.trycloudflare.com/telegram after retries
(last error: TelegramApiError: telegram setWebhook failed:
 400 Bad Request: bad webhook: Failed to resolve host: Name or service not known)
```

Seen on two consecutive nightly `live` runs with different hostnames:
- [33651121643](https://github.com/fastagent-sh/fastagent/actions/runs/33651121643) — `assessment-museum-matters-portraits`
- [33653445996](https://github.com/fastagent-sh/fastagent/actions/runs/33653445996) — `staff-prior-portable-local`

## What it is not

Not a regression from #428, whose retry loop is doing exactly what it was written to do — the log shows all 8 attempts, correctly classified as transient, correctly retried.

Not the tunnel being down: `tunnel.live.test.ts` passes in the same run (the URL carries a request back to the process), and in run 33653445996 **feishu verified its own Quick Tunnel URL successfully** while telegram could not resolve one minted seconds apart.

The error is DNS specifically — `Failed to resolve host`, not a connection or 530 — so Telegram's resolvers are taking longer than 80s to see a new `*.trycloudflare.com` name.

## Why it matters beyond CI

An author running `fastagent dev --tunnel` with a telegram channel hits the same wall and is told to register by hand. #428 made the wait honest (the platform's verdict is the probe); this says the budget behind that verdict may be too small for one platform/tunnel combination.

## Open questions

- What budget actually suffices? 80s is not enough; unknown whether 3 or 10 minutes is.
- Is it Telegram-specific, or Quick-Tunnel-specific, or the pair? Feishu succeeding on a sibling hostname in the same run argues for the pair.
- Should `REGISTRATION_ATTEMPTS` be per-registrar rather than shared, given the evidence that platforms differ?

Not proposing a number without data — logging the observation and the two runs that produced it.

kid7st:
Closes #435.

## The bug

A quick tunnel's hostname does not exist in DNS when cloudflared prints it. `*.trycloudflare.com` is
not a wildcard — each tunnel gets its own record, published only once the tunnel's **first connection
registers with the edge**, which is after the URL appears:

| | measured |
|---|---|
| cloudflared prints the URL | +0s |
| first connection registers | +1.5s |
| the record is published | +3–6s (0.2–4s over QUIC) |
| Telegram's first `setWebhook` | **+0.1–0.5s** (109ms / 304ms / 471ms across the three failing runs) |

Every registrar opens its mouth inside that window. Telegram then keeps answering `Failed to resolve
host` for all 8 attempts — 80s against a hostname that went live in under 4.

## Why the retry loop cannot save it

It is not the budget. Attempt 8 lands at t+80s, long after the record is live, and still gets
NXDOMAIN; no budget short of Telegram's own cache expiry would help. The variable is the **start
time**, not the length.

## The fix

`startCloudflareTunnel` does not hand the URL out until its hostname resolves — bounded at 30s, never
fatal. Every registrar is downstream of that one function, so waiting once covers Telegram, Feishu,
Lark, Slack and `add feishu`/`add slack` alike.

This is not the local-reachability gate #421 removed. That one asked whether *this machine* could
reach the origin, which a proxy or a slow resolver can make false for minutes while the platform
reaches it fine. This asks whether a **global DNS record exists** — `resolve4` goes to the configured
nameservers through c-ares, so each poll really re-queries — and on timeout it warns and serves
anyway.

The timeout also became a diagnosis rather than a shrug: because the record follows the connection, a
hostname that never appears means the tunnel never connected (a network dropping cloudflared's QUIC,
say), and nothing will reach that URL at all. The warning says that instead of blaming DNS.

## Two corrections to the issue

- **It is a regression from #428**, though not from its retry loop, which does exactly what it was
  written to do. #428 removed the `waitForHealth` gate, and on a runner with no proxy that gate was
  incidentally supplying the several seconds that kept the first platform call out of this window.
  Three consecutive live failures on commits containing #428; zero before.
- **It is not Telegram-specific.** In run 33651121643 — one of the two the issue cites — **feishu
  failed identically**, 8 attempts of `cannot verify`. Feishu survives when it happens to spend a few
  seconds on token and scope calls before its PATCH. Whoever speaks first loses, which is also why
  `REGISTRATION_ATTEMPTS` stays shared: the platforms do not differ, their warm-up time does.

## Verified

`lint` + `typecheck` + `test` (1617). Two offline checks, both red without the change: the URL is
withheld while DNS says NXDOMAIN, and the timeout path still serves *and* says why. Live, against a
real tunnel: held back 3.6s, and `resolve4` on the returned URL answers immediately — plus a run
where the tunnel never connected, which took the 30s branch and printed the warning.
`test/live/tunnel.live.test.ts` now asserts the returned hostname resolves, which is what catches a
URL escaping through that timeout.

### DIFF
--- a/src/tunnel.ts
+++ b/src/tunnel.ts
@@ -7,6 +7,7 @@
  * Process orchestration, not assembly — lives outside the engine, beside dev-supervisor.ts.
  */
 import { type ChildProcess, spawn } from "node:child_process";
+import { resolve4 } from "node:dns/promises";
 import type { RegistrationOutcome } from "./channels/registration.ts";
 import type { DeclaredChannel } from "./channels/discover.ts";
 import { registerFeishuWebhook } from "./channels/feishu/register-webhook.ts";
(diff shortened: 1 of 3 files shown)


kid7st:
Closing in favour of a redesign. The DNS gate here answers the wrong question: probing a name is what publishes the negative answer that then blocks the probe, and two rounds of patches (recursive resolver → authoritative nameservers → error-code triage) each fixed a symptom the previous one introduced. cloudflared already reports when the tunnel connects — that signal is local, free, and side-effect-free. Reopening against `Tunnel.ready`, where the consumer decides whether to wait rather than the tunnel deciding for everyone.

PR #484, opened 2026-09-07
refactor(pi): scope invocation and control execution with Effect

kid7st:
## Summary

- Keep invocation leases, sessions and subscriptions scoped through consumer settlement. Publish readiness after subscribing so waiting controls emit observable queue events. Cancel by aborting and joining actual SDK work, including late session acquisition and quiet iterator reads.
- Share typed SDK failures and scoped acquisition with control-plane writes. Keep compaction accept-fast and release resources before publishing its outcome.
- Translate callback and command failures at their protocol boundaries. Log cleanup faults without losing subsequent cleanup or replacing the operation outcome.
- Preserve public APIs, dependency-free contracts, shared assembly identity, channel shutdown and durable replay policies.

## Validation

- `npm run lint && npm run typecheck && npm test && npm run build`: passed, 1,733 tests across 121 files on Node 26.8.1.
- 178 focused tests passed on each of Node 22.19.0 and 24.20.0.
- Seven fault-injection groups and three compiler-negative checks verified.
- Real Pi tool cancellation and HTTP/control coverage; real Telegram/Pi integration verifies ACK-independent execution and completed-event context commits.
- Runtime, heap, build and implementation-size measurements are recorded in [#480](https://github.com/fastagent-sh/fastagent/issues/480#issuecomment-5574040792).

Refs #480.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -257,12 +257,15 @@ src/
     │                         # must contend on the same lease and validate against the same registry
     ├── turn-kit.ts          # the turn mechanism's pi-CLASS-neutral half: lease (single-writer
     │                         # floor), terminals (settled message/thrown error → SPEC terminal +
-    │                         # retryable), EventQueue (push→pull), prompt image prep, the SPEC
+    │                         # retryable), prompt image prep, the SPEC
     │                         # projection, and the observation seam (RunControls + SessionObserver)
     ├── invoke-session.ts    # THE L0: pi's AgentSession, one per invoke, over the same durable
     │                         # record. Events translate ONCE into the rich SessionEvent vocabulary;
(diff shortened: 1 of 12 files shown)


kid7st:
Addressed the handoff’s P2 readiness-ordering finding in 49d6d4c7. The SDK subscription is installed before the readiness Deferred succeeds, so initialization-time steer/followUp queue events reach both observers and state(). Initialization failures and interruption notify waiting commands with the original Cause; existing protocol boundaries retain their error responses.

Real-SDK regressions verify queued steer/followUp state and events. Factory/subscription failure checks cover all three waiting controls. Local lint, typecheck, all 1,733 tests, and build pass; 178 focused tests pass on each of Node 22.19.0 and 24.20.0. The informational deletion callout requires no behavioral change.

PR #496, opened 2026-09-08
fix(deploy): persist workspaces across restarts and releases

### LINKED ISSUE #472: agentcore: persona's self-improvement promise is silently ephemeral on a baked deployment
## Problem

The scaffolded `persona.md` tells the agent it can improve itself by editing `persona.md` and writing new `skills/*/SKILL.md`. On AgentCore those edits are silently lost: the definition lives in the image at `/app`, which is outside the S3-snapshotted state root (`/mnt/state`), so an edit survives at most until the microVM is reclaimed (3 min idle) and never survives a redeploy.

`docs/deploy.md` documents the intended pattern ("freshness and write-back run through git, driven by the agent", "definition updates need a redeploy"), and `deploy` prints a note about it at deploy time. But the agent never sees either: the persona it runs with still says "write it into your definition", with no mention that the write must be committed and pushed **in the same turn** to persist, and the workspace ships no write-back credential unless the author happened to declare one in `deploy.secrets`.

Observed in a fresh 0.21.0 workspace deployed with `fastagent deploy agentcore --run`.

## Why it matters

This is the one place the definition's promise and the host's behavior disagree, and the failure is invisible: the agent reports "I've updated my skill", the owner believes it, and the next turn after a reclaim runs the old definition. That contradicts the fail-visible principle the rest of deploy follows.

## Suggestions (any one closes the gap)

- `fastagent deploy agentcore` (and other baked hosts) gates or warns when the persona contains the self-improvement scaffold text and no write-back credential (`GH_TOKEN` or similar) is in `deploy.secrets`.
- Ship a host-aware line into the container's context (e.g. an injected note in the system prompt when `FASTAGENT_AGENTCORE=1`): "your definition is baked; an edit persists only if you `git commit` and `git push` it in this turn".
- Or: on AgentCore, include the definition directory in the S3 snapshot so edits survive reclaim and are re-applied over the image on restore (redeploys still win).
- At minimum, the persona scaffold could carry one sentence: "On a deployed copy, an edit persists only through git; see docs/deploy.md."

kid7st:
Closes #472.

## Behavior

- Preserve the workspace, unfinished files, state and credentials independently of the compute lifecycle. Git remains optional.
- Seed the whole workspace once. Publish subsequent releases into the nested definition directory only; same-release restarts retain its edits.
- Recover interrupted publication before opening the agent, exclude competing processes with a kernel file lock, and retry interrupted dependency installation.
- Support explicitly disposable bind-mounted directories on Docker, Fly and AgentCore. Railway keeps dependencies on its volume.
- Use an operator-owned EFS access point and VPC for AgentCore. Every entry point shares one runtime session while envelope session IDs select separate conversations.
- Bootstrap AgentCore storage on invocation and load the active workspace's FastAgent package so tools share their session context.
- Give each live AgentCore probe an independent, disposable EFS stack in the configured test network.

## Deployment setup

Deployment requires a nested definition. AgentCore requires `deploy.agentcore` with an existing EFS access point, subnets and security groups. Regenerate deployment artifacts to publish a new definition release.

The live probes require `FASTAGENT_LIVE_AGENTCORE_NETWORK` and EFS provisioning permissions in the CI AWS policy.

## Verification

- Lint, typecheck and the full local suite: 1,798 passed; two Linux-only tests skipped on macOS.
- Linux workspace/locking suite: 24 passed. Mutation checks detect a leaked lock descriptor and the descriptor-observation race.
- Node 22.19 targeted regressions: 38 passed; two Linux-only tests skipped on macOS.
- Real AgentCore/EFS: native mounting, idle recovery, stop/restart, competing-writer exclusion, definition-only redeployment, Git and unfinished-file retention, synthetic credential retention, and disposable-file removal.
- Real CloudFormation creation and deletion of the isolated EFS test stack.

Fly bind mounts have not been exercised against a live Fly account. The EFS cloud probe used no external model calls; the full provider/wake probes require the CI network configuration above.


kid7st:
## Revised storage plan for AgentCore: SessionStorage, not EFS

After tracing the cost chain this branch introduced, the EFS mount is a consequence of one new product promise rather than an AgentCore requirement:

```
persist the whole workspace
  -> durable data grows from "a few MB of session state" to "repo + .git + node_modules"
  -> the S3 snapshot no longer fits
  -> a real mount is needed -> EFS
  -> EFS needs VPC mode -> no public egress -> NAT gateway (~$33/mo standing)
  -> EFS is slow for many small files -> deploy.temporaryDirectories -> bind mounts -> SYS_ADMIN
```

Every link is a reasonable step; together they turn AgentCore from "AWS CLI + Docker, pay per use" into "bring your own VPC, EFS, mount targets, security groups and a NAT gateway". The previous design had already priced that trade (`deliberately not the default (~$33/mo)`) and this branch reverses it without recording why.

### What AWS actually offers

| Filesystem | Across stop/resume | Across deploy (version update) | VPC |
|---|---|---|---|
| Session storage (preview) | yes | **no — "resets on version update"**, plus 14-day idle expiry | not required |
| EFS access point | yes | yes | **required** |
| S3 Files access point | yes | yes | **required** (also NFS 2049 to a mount target) |

There is no cross-deploy persistence on AgentCore without a VPC, and with a VPC there is no model/channel egress without NAT. So the choice is not "which durable mount" but "does AgentCore promise cross-deploy memory at all".

### Decision

It does not. AgentCore keeps one `SessionStorage` mount at `/mnt/data`; a deploy resets the workspace, `.state` and `.secrets`. That is the platform's own semantics (the image is the definition), stated instead of engineered around.

Credential handling needs no new rule: `maybeSeedAuth` is absent-only, so a restart within a release keeps the credential the box rotated, and a deploy — which wipes the mount — re-seeds from `FASTAGENT_AUTH_SEED`. Deploying *is* re-authenticating. The one thing to document is that an OAuth refresh token is single-use and shared with the builder machine, so the box can lose model access between deploys; the fix is another deploy, and the failure must say so rather than surface a bare 401.

### What this removes

- `deploy.agentcore` (EFS ARN / subnets / security groups) and its config validation
- `NetworkMode: VPC` + `NetworkModeConfig` + the EFS IAM statements — back to `NetworkMode: PUBLIC`
- `deploy.temporaryDirectories` in full: the config key, `validateTemporaryDirectories`, `mountTemporaryDirectories`, `unmountWorkspace`, Docker's `cap_add: [SYS_ADMIN]` + `apparmor=unconfined`, Railway's gate, and the image's dependency on `mount`/`umount`

`temporaryDirectories` only ever had a motive on a network filesystem. On Docker and Fly the volume is local, so bind mounts bought nothing and cost container isolation; Railway rejected the deploy outright over a pure optimization; AgentCore — the only host with a real motive — had no handling at all. SessionStorage is a local disk that survives stop/resume, so `node_modules` needs no special treatment there either.

Kept from this branch: the release manifest, the workspace lease, and the recoverable definition replacement — those earn their place on the hosts with real volumes (Fly, Railway, Docker).

Net effect of the PR becomes "one workspace lifecycle across hosts, minus the ~700 lines of S3 snapshot/presign/checkpoint machinery", with no new infrastructure prerequisites.

### Review findings folded into the same push

- `start.ts` dropped `.catch(failStartup)` on `createPiAgentFromDir`/`mountAgentService`, so a user-fixable startup error (a broken `channels/` module, a bad config) now prints a raw stack trace instead of the one-line `Error:` the CLI promises.
- The workspace-local package is imported without checking that it exports `openPreparedStartService`; a version-skewed agent dependency fails with `open is not a function` and no version in the message.
- The AgentCore branch calls `serve()` without `onListening`, so a booted container logs nothing at all until its first envelope.
- `openStartService` overwrites an operator-set `FASTAGENT_STATE_DIR`/`FASTAGENT_SECRETS_DIR` silently.
- The boot install runs through `execFile` with the default 1 MB buffer and no streamed output.
- `writeArtifacts` matches the release manifest by a hardcoded `"/fastagent.release.json"` instead of `RELEASE_FILE`.
- `deferAgentcoreService`'s catch also covers `ready.handler(request)`, labelling a serving failure as `initialization failed`.
- `deploy` now rejects flat agents, but `quickstart.md`, `ai-start.md` and `cli.md` still recommend `init . --flat`.

PR #506, opened 2026-09-09
info/deploy: warn about env names read by tools/schedules/channels that no deploy secret carries

### LINKED ISSUE #473: info/deploy: report process.env names read by tools/schedules/channels that are missing from deploy.secrets
## Problem

Tools, schedules and channels read secrets with `process.env.NAME`, and `deploy` only ships what `config.deploy.secrets` lists (plus channel secrets it knows about). Nothing checks the two against each other. Forgetting one name means the deployed tool fails on its first real call, days later, with an error only in CloudWatch.

In this workspace `tools/x-post.ts` reads four `X_*` vars and `tools/github-api.ts` reads `GITHUB_TOKEN`; `schedules/daily-digest.ts` reads `SLACK_DIGEST_CHANNEL`. All six had to be typed into `deploy.secrets` by hand, and the only thing that would have caught an omission was remembering.

## Suggestion

Add a diagnostic (in `fastagent info` and as a `deploy` preflight warning) that scans `tools/`, `schedules/` and `channels/` for `process.env.<NAME>` / `process.env["<NAME>"]` and reports names that are neither in `deploy.secrets`, nor a channel secret the adapter declares, nor a fastagent runtime variable (`PORT`, `FASTAGENT_*`). A static regex over the source is enough; false positives can be silenced by listing the name.

Sample output:

```
WARN  tools/x-post.ts reads X_API_KEY, X_API_SECRET, X_ACCESS_TOKEN, X_ACCESS_SECRET — not in config.deploy.secrets
WARN  schedules/daily-digest.ts reads SLACK_DIGEST_CHANNEL — not in config.deploy.secrets
```

kid7st:
Closes #473.

`src/deploy/secret-scan.ts` statically scans `tools/`, `schedules/` and `channels/` sources for
`process.env.NAME` / `process.env["NAME"]` and reports names that are neither in
`config.deploy.secrets`, declared by a first-party channel adapter (both ingress modes), nor
runtime-provided (`PORT`, `FASTAGENT_*`, `HOME`/`PATH`/proxy vars). Both readers print the same
wording:

- `fastagent info` — a stderr warning, plus `undeclaredSecrets` in `--json`.
- `deploy` pre-flight — an advisory `warn` message (never a gate); it also treats the model key env
  var as carried, which `info` cannot know.

```
WARN  [fastagent] tools/x-post.ts reads X_API_KEY, X_API_SECRET — not in config.deploy.secrets, so a deployed box would read undefined. Add the names there (or ignore this if the value is local-only).
```

Nothing is imported or executed, so a tool that fails to load on the builder machine is still
scanned. Known limitations, documented in the module header: computed names
(`` process.env[`X_${k}`] ``), destructuring (`const { X } = process.env`) and aliases are invisible;
a name inside a comment or string counts as a read. A false positive is silenced by listing the name
in `deploy.secrets`.

Verified: new pre-flight test (undeclared tool/schedule names warn; declared, runtime and
adapter-declared names stay silent), plus a manual `info` / `info --json` run on a scratch agent.


kid7st:
Superseded by the declaration-based fix: the scan reported a hand-sync from outside the code; the replacement removes the sync (tools/schedules declare their own secrets, deploy carries them, serving asserts them).

PR #515, opened 2026-09-09
refactor(pi): one admission rule for every run control

kid7st:
## What changed

`createPiAgentFromSession`'s three run controls each carried their own copy of the same admission
rule (wait for this run's binding, then refuse once it has settled), including three copies of the
refusal message. `command(run)` now holds that rule, and `enqueue(prompt, kind)` collapses `steer`
and `followUp`, which differed only in which pi queue they name.

## Ordering the helper has to preserve

Admission and the enqueue must stay adjacent, with no await between them. `AgentSession.steer()`
pushes onto `_steeringMessages` for a finished run without complaining, so a command admitted while
the run is live and delivered after it settles is dropped from the model call and still reported as
`{ ok: true }` by `runAction`. Prompt preparation therefore stays outside `command()` — where it also
keeps overlapping session acquisition, so a first image-carrying `steer` does not serialize a resize
behind binding.

`abort`'s in-flight counter and success flag keep their exact positions: the increment happens after
admission and before the port call, inside the effect rather than at construction.

## Verification

`npm run lint && npm run typecheck && npm test`: 1,642 passing, 2 skipped, 127 files.

The settled-refusal rule keeps its existing coverage in `test/invoke-session.test.ts` and
`test/session-control.test.ts`; `followUp` reaches it through the same `enqueue` path as `steer`.
`refuses a command whose own prompt preparation outlived the run` covers the ordering constraint
above by settling the run from inside the steer's own preparation — it fails if preparation moves
behind the admission check.

Follow-up to the #480 Effect adoption review.

### DIFF
--- a/src/engines/pi/invoke-session.ts
+++ b/src/engines/pi/invoke-session.ts
@@ -131,39 +131,35 @@ export function createPiAgentFromSession(options: CreatePiAgentFromSessionOption
       }
     };
     const ready = Deferred.await(bound);
+    /**
+     * Every control command obeys one admission rule: wait for this run's binding, then refuse once it has settled.
+     * Kept in one place so a fourth command cannot arrive with a fifth spelling of it.
+     */
+    const command = (run: (session: AgentSession) => Effect.Effect<void, PortFailure>): Promise<void> =>
+      Effect.runPromise(
+        Effect.gen(function* () {
+          const session = yield* ready;
+          if (settled)
+            return yield* Effect.fail(
+              new PortFailure(new Error("run already settled; the command cannot take effect")),
+            );
+          yield* run(session);
+        }),
+      );
+    // Image preparation is pure and runs after admission: a settled run must not pay for a resize nobody reads.
+    const enqueue = (p: Prompt, queue: "steer" | "followUp"): Promise<void> =>
+      command((session) =>
+        Effect.gen(function* () {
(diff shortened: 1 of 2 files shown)


kid7st:
Review findings addressed in `7c3f6715`.

**P1 — admission and enqueue were separated by an await.** Moving `toPiPromptOptions` behind the
settled check opened a real window: the check passed, an image resize (dynamic import + Photon) ran
for hundreds of milliseconds, the run ended, and only then did the message reach
`AgentSession.steer()` — which pushes onto `_steeringMessages` for a finished run without
complaining, so `runAction` returned `{ ok: true, runId }` for a command nobody would ever consume.
Preparation is back outside `command()`, so the check and the enqueue are again adjacent.

**P3 — lost overlap with session acquisition.** Same fix: preparation once more runs while `ready`
is still pending, so a first image-carrying steer does not serialize resize behind binding. The
comment now states the ordering constraint instead of the (wrong) cost argument that motivated the
move.

**P3 — `queue` shadowing.** The parameter is `kind`; `queue` in that closure is only the run's event
queue again.

New regression test `refuses a command whose own prompt preparation outlived the run`
(`test/invoke-session.test.ts`) settles the run from inside the steer's own preparation hook. It
fails on the previous ordering (steer resolves, message dropped) and passes now.

`npm run lint && npm run typecheck && npm test`: 1,642 passing, 2 skipped, 127 files.

PR #522, opened 2026-09-11
refactor(deploy): read carried values from the value file only

kid7st:
Step 2b of [docs/design/configuration.md §12](../blob/main/docs/design/configuration.md). Stacked on #520 (the classification it implements lands there).

## The gap

#520 established one rule for the model: **the environment running `deploy` is not part of the environment being deployed.** Every other carried value still ignored it. `assembleSecrets` read `process.env`, which by then is `.secrets/.env` merged into the builder's shell by `loadDotEnv` — the two are indistinguishable after the merge.

So `TELEGRAM_BOT_TOKEN=x fastagent deploy fly --run` pushed the shell's value while `.secrets/.env` said something else, and preflight could not report which one travelled, because it could not tell.

## The rule

The selected value file is the only source. The environment running `deploy` is not consulted, so there is no precedence to remember and no way for a deployment to carry a value that is written down nowhere.

The industry splits on this exactly along the line of what a tool configures:

| Tool | Rule | Configuring |
|---|---|---|
| dotenv | never overrides an already-set variable | this process |
| Docker Compose | shell beats `env_file` | this process (the container) |
| Terraform | `terraform.tfvars` beats `TF_VAR_*` (checked last) | a remote object |
| Kamal 2 | reads `.kamal/secrets` only; stopped loading `.env` into the environment | a remote deployment |

A deployment is the second kind. This is also what §9 of the design already said ("CI materializes the same file before running the same command") — the implementation had not caught up.

## Changes

- `assembleSecrets` takes `values: ReadonlyMap<string, string>` instead of `env: NodeJS.ProcessEnv`.
- `preflightDeploy` reads the file once and publishes `values` + `valueFile` on its facts, so the plan side and the run side cannot disagree about what this deployment carries. The four hosts pass it through with everything else.
- Gate wording on every host: `no value for: X — the deployed environment is declared by the agent's .secrets/.env, and this deploy reads only that file (exporting the variable here does not reach the deployment). Add them there and re-run`.
- Runbook and docs updated where they told operators to export variables.

**Breaking for scripted CI** that exported variables instead of writing the file. The remedy is the same one OpenAI's own CI guide gives for `auth.json`: write the file before running the command.

## Verification

New case in `test/deploy-fly-run.test.ts`: an exported `GH_TOKEN` with an empty value file lands in `missingSecrets` and never in `secrets`. `npm run lint && npm run typecheck && npm test` pass (1656 passed).


kid7st:
Superseded by #524: merging #520 deleted this PR's base branch, which closed this one and left it unreopenable. Same branch, same content, base `main`.

PR #532, opened 2026-09-12
refactor(init): one layout — the definition always lands in a subdirectory

### LINKED ISSUE #531: init: one layout — the definition always lands in a subdirectory (drop --flat and --minimal)
## Decision

One layout, no placement knobs. `init` always scaffolds the definition into a subdirectory (`./fastagent/` by default, `--agent-dir <name>` renames it), and always scaffolds the complete kit.

This is not dropping the "this repo is the agent" adoption path. That repo stays an agent host — it is the **workspace**, and the definition lives in `./fastagent/` inside it. What goes away is `init` writing a definition *into* a directory that already holds someone else's files.

Two flags go with it:

- `--flat` / `--agent-dir .` — the only reason `init` had an "adopt this directory" mode.
- `--minimal` — a second scaffold shape nothing depends on; the complete kit is the convention.

## Why

1. **Convention over configuration** (`docs/design/configuration.md` §1): conventions govern what this agent *is*, including its directory layout. A flag that relocates the definition is a placement choice the author should not have to make, and `--flat` is a second spelling of `--agent-dir .` — one knob, two names, held together by a `conflicts` rule.
2. **Flat is already second-class.** `deploy` refuses it (preflight requires the agent to sit inside a workspace, because a release replaces the agent directory and nothing else). A supported layout that half the product cannot serve is a maintenance tax with no upside.
3. **Adoption is where the sharp edges live.** `--flat` keeps every existing file, so the agent inherits the host repo's `package.json`. Since the config file is now always `fastagent.config.ts` (#TBD / `refactor/single-config-filename`), a `.ts` config inherits the nearest `package.json` module type — where `.mjs` was unambiguous. Reproduced on `main` + that branch:

   ```console
   $ cd /tmp/flatcjs2          # package.json: {"name":"repo","type":"commonjs"}
   $ fastagent init . --flat --no-install
   ... scaffolded, exit 0
   $ fastagent info
   Error: /tmp/flatcjs2/fastagent.config.ts: Unexpected token 'export'
     (the agent dir must be ESM — set "type": "module" in package.json)
   ```

   With a `package.json` that has no `type` field at all, it loads but Node prints `[MODULE_TYPELESS_PACKAGE_JSON] ... Reparsing as ES module ... This incurs a performance overhead` on **every** config load (every command, every `dev` restart).

   A scaffolded subdirectory carries its own `type: module` manifest, so fixing the layout removes this class of failure at the source. No ESM pre-check in `init` is needed — that check would only exist to serve the mode being deleted.

## Scope

| Location | Change |
|---|---|
| `src/cli/program.ts:49, 53-58, 64, 75, 78` | Drop the `--flat` and `--minimal` flags, the `fastagent init . --flat` example, and the `f.flat` / `f.minimal` mapping (the `conflicts: ["agentDir"]` rule goes with `--flat`) |
| `src/scaffold/init.ts:29-35` | `agentDirNameError` stops accepting `"."`; its message explains that the definition lives in a subdirectory and the directory around it is the workspace |
| `src/scaffold/init.ts:63, 91, 134-140, 153, 176` | Remove the `flat` branches: the target must be empty, and `writeFile(..., "wx")`'s `EEXIST` is no longer swallowed (a collision is a failure again) |
| `src/scaffold/init.ts:38-56, 84-93` | Remove `ScaffoldOptions.minimal` and `ScaffoldResult.complete`; the file list is unconditional |
| `src/paths.ts` | **Unchanged.** `resolvePlacement`'s `holds a config -> { agentDir: dir, workspace: dir }` row is a runtime fact a deployed container depends on (it is shipped the agent directory alone); only `init`'s ability to *create* that shape goes away |
| `docs/cli.md:45, 54, 62, 65`, `docs/quickstart.md:49, 62`, `docs/ai-start.md:67, 99, 105`, `docs/design/core.md:69` | Restate as: the repo is the workspace, the definition is in `./fastagent/`; drop the flat and minimal documentation |
| `test/init.test.ts:121, 205-232, 277-315` (+ the `minimal: true` fast path used across the file) | Replace the `--flat` case with a refusal of `--agent-dir .`; drop the `--minimal` case; the remaining cases scaffold the complete kit |

## Out of scope

- `--agent-dir <name>` stays: the name never decides what is an agent, and a workspace may hold several agents (`FASTAGENT_AGENT` picks).
- Placement *resolution* is untouched, so an existing flat agent on disk keeps serving; only `init` no longer produces one.

kid7st:
Closes #531.

`init` always scaffolds the complete kit into a subdirectory (`./fastagent/`, renamed by `--agent-dir <name>`). `--flat`, `--agent-dir .`, and `--minimal` are gone.

**A standalone agent repository is still supported** — that repository is the *workspace*, and the definition lives in `./fastagent/` inside it. Placement still *resolves* an agent sitting at a directory (a deployed container is shipped the agent directory alone, so that row is a runtime fact), which is why `src/paths.ts` is untouched. Only `init`'s ability to *create* that shape goes away.

## Why the flat mode had to go rather than get a pre-check

It kept every existing file, so the agent inherited the host repo's `package.json` — and since the config file is now always `fastagent.config.ts` (#530), its module type comes from that manifest. In a CommonJS repo the scaffold succeeded and then every command failed:

```
Error: …/fastagent.config.ts: Unexpected token 'export'
```

With a `type`-less manifest it loaded but Node printed `[MODULE_TYPELESS_PACKAGE_JSON] … Reparsing as ES module` on **every** config load. A scaffolded subdirectory carries its own `type: module`, so fixing the layout removes the failure class at the source. An ESM pre-check in `init` would only have existed to serve the mode being deleted.

Verified after the change, in a repo with `{"type":"commonjs"}`: `init` writes only `fastagent/`, the agent's manifest is `module`, and `info` resolves the config. `--agent-dir .` exits 2 with the usage message; `--flat` and `--minimal` are `unknown option`.

## What the deletion took with it

- **`ScaffoldResult.kept` and every branch reading it.** With no adoption mode the target must be empty, so `writeFile(…, "wx")`'s `EEXIST` is a failure again rather than a "kept yours" report — and the four `init` notes about inherited `package.json`/`.gitignore` files describe a situation that can no longer arise.
- **`ScaffoldOptions.minimal` / `ScaffoldResult.complete`.** One scaffold shape; the file list is unconditional. Don't want the example tool? Delete it — everything is yours after `init`.
- **`FlagSpec.conflicts` and its now-unused `optionKey` helper (`src/cli/kernel.ts`).** `--flat` was the only conflicts user, so removing it left the field, its build-time validation, its commander binding, and its test with nothing to guard. Restoring conflicts is three lines over commander's native `.conflicts()` if a mutually exclusive pair ever shows up.

### DIFF
--- a/docs/ai-start.md
+++ b/docs/ai-start.md
@@ -64,7 +64,7 @@ directories or the same directory.
 | Situation | Placement |
 |---|---|
 | New agent or an agent for an existing project | `fastagent init [workspace]` creates `workspace/fastagent/`; existing workspace files stay untouched. |
-| A standalone agent repository or an existing package that is itself the agent | `fastagent init . --flat` puts the definition in the current directory and keeps existing files. Review retained package and ignore files. Not deployable as-is: `deploy` requires the agent to sit inside a workspace. |
+| A standalone agent repository or an existing package that is itself the agent | Run `fastagent init` in it: that repository is the WORKSPACE and the definition lands in `./fastagent/`. This is also the shape `deploy` requires. |
 | An existing application embeds the agent | Keep the application's layout. A nested definition is convenient; the app retains auth, routes, database, and deployment. See [embedding](#8-embed-only-what-the-application-needs). |
 
 A config file identifies an agent, not its directory name. Check the workspace itself and its direct
@@ -96,7 +96,7 @@ my-agent/                         # workspace; run FastAgent commands here
 ```
 
 `init` installs dependencies in `fastagent/`. If installation fails, run `npm --prefix fastagent install`
-before continuing. `--no-install` defers that install; `--minimal` omits the code tool and package manifest.
+before continuing. `--no-install` defers that install.
 A global CLI installation alone does not make package imports available to authored tools.
 
 **Working-directory convention for the rest of this guide:** stay in `my-agent/`, the workspace.
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -42,16 +42,16 @@ Most commands take an optional workspace directory (the agent is there, or in it
 ## `fastagent init`
 
 ```bash
-fastagent init [dir] [--minimal] [--no-install] [--agent-dir <name>|--flat]
+fastagent init [dir] [--no-install] [--agent-dir <name>]
 ```
 
 Creates a self-iterating agent — it can edit its own definition (persona.md and skills are re-read every turn). A fresh agent has `persona.md` (the agent's identity: how to improve yourself), a `writing-great-skills` example skill (from [mattpocock/skills](https://github.com/mattpocock/skills) — the guide to authoring skills), a `fetch-url` example code tool, config, `.secrets/.env.example`, and `.gitignore`. No `AGENTS.md` is scaffolded (it is project context, not identity); an existing one is kept untouched. Everything is written offline; by default it also writes `package.json` and runs `npm install`. Ignore hygiene is two scaffolded files: the agent `.gitignore` (`node_modules/`, `.state/`, a stray `.env`) and `.secrets/.gitignore` (everything but `.env.example`). fastagent writes both once, at `init` — no later command reads, rewrites or verifies them, so they are yours. They are separate on purpose: the root one is the file you will edit, and a nested `.gitignore` outranks it, so the credentials stay protected either way.
 
-**Where the files land** — no detection and no prompt. By default the WHOLE agent — definition, config, `.secrets/`, `.state/` — goes into `./fastagent/`; the directory around it gets zero writes and becomes the agent's WORKSPACE when you point fastagent there. The agent self-contains its `package.json`, so the workspace's manifest and lockfile are never touched. `init` refuses when the target already holds a `fastagent.config.ts` (already an agent) or any other content.
+**Where the files land** — no detection, no prompt, no choice. The WHOLE agent — definition, config, `.secrets/`, `.state/` — goes into `./fastagent/`; the directory around it gets zero writes and becomes the agent's WORKSPACE when you point fastagent there. The agent self-contains its `package.json`, so the workspace's manifest and lockfile are never touched. `init` refuses when the target already holds a `fastagent.config.ts` (already an agent) or any other content.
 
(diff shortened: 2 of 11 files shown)


kid7st (commented):
## Interactive review record

This records the curated outcome of an interactive `/review` session. It omits private model reasoning and the raw conversation.

## Review Scope
- 审查了相对 `4b77dc9caac59f7001cd0dc6bdcc39724e433103` 的 11 个变更文件，以及 PR #532、issue #531 和全部 review threads。
- 已运行 `npm run lint`、`npm run typecheck`、`npm test`，全部通过。

## Constraints & Preferences
- `init` 必须只生成完整的子目录布局，不恢复 `--flat`、`--agent-dir .` 或 `--minimal`。
- 现有 flat agent 的运行时解析仍须保留。

## Findings

### [P2] 更新仍描述已删除 flat 模式的设计真源 — `docs/design/core.md:98-101`

这里仍声明 `--agent-dir .` 会采用当前目录并保留已有文件，直接违背本次变更后的 CLI 和 scaffold 行为。`docs/design/core.md` 是当前架构的权威文档，维护者会由此得到错误契约。应改为“`init` 始终写入空的直接子目录；flat 布局仅保留运行时解析支持”，并同步修正 `test/env.test.ts:115` 中“`--agent-dir .` exists”的陈旧注释。

### [P3] 删除失去生产调用方的 `optionKey` — `src/cli/kernel.ts:66-77`

删除 `FlagSpec.conflicts` 后，`optionKey` 已无生产调用方，只被 `test/cli-kernel.test.ts` 用来测试自身，删除该测试不会损失任何产品行为保护。应删除该函数、JSDoc、测试和 import；同时将目前误归入 `describe("cli kernel: the option-key naming rule")` 的三个 help 测试恢复到独立的 `describe("cli kernel: help")`。

## Review Decisions
- (none)

Verdict: correct

## Human Reviewer Callouts (Non-Blocking)

- **This change introduces backwards-incompatible public schema/API/contract changes:** 删除了公开 CLI 选项 `--flat`、`--agent-dir .` 和 `--minimal`；旧调用现在会以 usage error 退出。

PR #537, opened 2026-09-13
feat(runtime): refuse a second writer over one agent's file-backed state

### LINKED ISSUE #490: schedule: make the cron slot claim exactly-once across processes
## Problem

Suggested priority: P1. Reviewed at `3d2ea7a` (0.21.1).

File-backed deployment state requires a single writer, but the default session lease is a Set allocated per agent assembly. Separate instances do not contend. `fastagent fire <name>` creates another agent over the normal session directory and uses the same stable `schedule:<name>` session as the running scheduler. Two services can likewise point at the same state while using different ports. These ordinary commands can enter a topology the documentation declares unsupported without a startup refusal.

A local mechanism check created two independent L0 agents with a held session factory. Both entered the factory for the same session before either completed. This proves independent lease ownership only; it was not a cross-process disk-corruption experiment.

Sources: [default lease](https://github.com/fastagent-sh/fastagent/blob/3d2ea7a/src/engines/pi/turn-kit.ts#L35-L49), [directory opener](https://github.com/fastagent-sh/fastagent/blob/3d2ea7a/src/engines/pi/open.ts), [fire command](https://github.com/fastagent-sh/fastagent/blob/3d2ea7a/src/cli/commands/fire.ts), [single-machine scope](https://github.com/fastagent-sh/fastagent/blob/3d2ea7a/docs/deploy.md#single-machine-tier).

## Proposal

Make ownership of file-backed writable state an enforced opening/lifetime contract. Reject a conflicting writer with the owner/path and an actionable way to use the already-running service or a separate state location. Consider actual resolved write paths, including an overridden sessions directory; guarding only a nominal state root is insufficient.

Use existing local locking/ownership facilities. Do not add distributed state backends, automatic remote-target selection, or a registry. Keep pure read-only commands usable. Global provider authentication has its own coordinated store and must not be confused with exclusive project runtime state.

## Acceptance checks

- Two processes opening the same file-backed runtime state cannot both write.
- A serving scheduler plus `fire` cannot concurrently write the same scheduled conversation.
- Independent state directories continue to run in parallel.
- Normal close and process termination allow recovery without permanently stale ownership.
- Custom shared backends and deliberately shared injected leases retain their own documented ownership contract.

Related: #482/#483 for state placement and isolated tests. The goal is to enforce the shipped single-writer tier rather than silently imply multi-instance support.

kid7st:
Closes #490.

## The gap

`docs/deploy.md` and `docs/design/core.md` said file-backed state is single-process; nothing enforced it. The session lease is an in-memory `Set` per assembly, so two processes never contend, and two ordinary commands walk into the unsupported topology: `fastagent dev` serving while `fastagent fire digest` runs the same `schedule:digest` conversation, or two `start`s on different ports over one directory. pi's session store is an append-structured journal, so concurrent writers interleave entries; channel state is whole-file rewrites, so updates are silently lost.

## The change

Opening a directory takes an exclusive **write claim** on its resolved write paths — the state root *and* the sessions directory, because `--sessions-dir` moves them apart and guarding only the nominal root would miss a second run pointed at the same journals.

The claim is a **listening unix socket** (`/tmp/fastagent-<hash of the directory>.sock`), not a lock file, because that turns the two questions this guard must answer into facts the kernel owns:

| Question | How it is answered |
|---|---|
| Is the holder still alive? | A dead holder's socket refuses connections the instant its process ends. No stale window, no heartbeat, no asking the OS about a pid — which in a container, where the agent is pid 1, answers wrongly about the asker itself. |
| Who holds it? | The holder answers with its own pid and command line, so the refusal states the holder instead of inferring it. |

Only the kernel's "nobody is listening" (`ECONNREFUSED`/`ENOENT`) permits taking over a claim. A holder too busy to answer, a foreign socket, an unreadable path, or a non-socket file in the way all **refuse** — a refused run is recoverable, a stolen claim is not. What occupies the path is decided by `lstat`, not by the connect errno, which differs between Linux and macOS. The one non-atomic step (unlink a dead name, then bind it) runs under an `O_EXCL` sentinel, so two simultaneous takers produce a refusal rather than two writers who each believe they are alone.

**Ownership is one chain, not a convention every caller remembers:** the opener holds the claim, mounting transfers it (`mountAgentService` and `mountAgentcoreService` both release on failure and on `close()`), an opener that throws after taking it releases it, and a process that exits or dies drops it with its socket. `dispose()` is idempotent.

Scope, stated rather than implied: the guard is per kernel namespace (the machine, or the container). One process per mounted volume is a different question, already answered by `leaseDeployment`. `chat` is outside it (pi's TUI owns its own session storage), and `dev --tunnel`'s supervisor writes channel onboarding state while its worker holds the claim — once per registration, documented in `src/state-lock.ts` and `docs/design/core.md`.

## Behaviour change

`fastagent invoke` / `chat`-adjacent one-shots and `fastagent fire` are **refused** while `dev`/`start` serves the same directory, and a second `createAgentService` on one directory (in this process or another) rejects. That is the shipped single-writer tier being enforced instead of implied; the refusal names the holder and the ways out. There is deliberately no flag that turns the guard off — a second writer is what it exists to prevent. `docs/cli.md`, `docs/quickstart.md`, `docs/embedding.md` and `docs/api-reference.md` carry it.

## Verification

`npm run lint && npm run typecheck && npm test` (1699 passed), Node 22/24/26 in CI.

`test/state-lock.test.ts` covers: the in-process refusal and its own wording, a second **process** being refused by name (its pid and command), a SIGKILLed holder's claim being taken immediately, a holder too busy to answer NOT being treated as dead, a takeover already in flight refusing instead of racing, a non-socket file being refused rather than deleted, per-path guarding (a shared sessions directory under two state roots), rollback of a partial multi-path claim, and the opener/dispose lifecycle. `test/service.test.ts` and `test/agentcore-service.test.ts` cover both mounts releasing on failure.

Each guard was mutation-checked: removing it fails its test. The `lstat` check was found by CI — on Linux, connecting to a regular file answers `ECONNREFUSED`, which macOS reports as `ENOTSOCK`.

## Known limits

Windows is out (unix sockets under `/tmp`; the repo claims no Windows support). Multi-user sticky-`/tmp` behaviour is reasoned from permission bits, not exercised. A shared network volume across hosts is not covered by this guard, by construction — that is the tier boundary, not a bug.


kid7st:
Closing unmerged: the premise this was built on does not hold.

Measured rather than assumed (spikes in [#490](https://github.com/fastagent-sh/fastagent/issues/490#issuecomment-latest)): two processes writing one session **do not corrupt it** — pi appends one whole line per transaction, so the file stays parseable and the conversation branches instead. What two processes actually cost is one specific decision, the cron slot claim, which `fireScheduleOnce` makes with a non-atomic read-modify-write of `fires.json`.

So this PR forbids harmless topologies (two processes serving different channels, `invoke` during `dev`, an embedder mounting twice) and cannot prevent the dangerous one (the same bot token in two places, which no directory lock sees), while its complexity is entirely in questions the framing created: is the holder alive, who holds it, may this claim be taken over. Eight rounds of review found ten defects in that machinery, every one of them inside those questions.

Replaced by a much smaller change: an atomic slot claim in the scheduler. #490 is rewritten accordingly.

Issue #553, opened 2026-09-16
no way to read a past scheduled turn: attach never backfills, and there is no offline session reader


kid7st:
## What is missing

There is no way to read what a past scheduled turn said.

#549 removed `<stateRoot>/schedule/runs.jsonl`, which had carried a copy of each fire's reply. That copy was the
problem it was deleted for (#546: model output accumulating where nothing prunes it), and the text is not lost — it
is in the session the fire ran in (`schedule:<name>`, persisted under `<stateRoot>/sessions/`). What is missing is a
way to read it back:

| Path | Reads a past turn? |
|---|---|
| `fastagent schedule history <name>` | No — the claim holds `firedAt` / `outcome` / `ms`, by design |
| `fastagent attach <session>` | **No**, and see below: the capability is there, the client does not ask for it |
| `fastagent chat` + `/resume` | No — `buildAgentSessionRuntime` uses `SessionManager.create(rootCwd)` (pi's own per-workspace location), not `<stateRoot>/sessions/` |
| `fastagent logs` | No — the reply is deliberately not logged |
| `GET /control/sessions/<id>/entries` | **Yes** — with no `since`, `entries()` returns the whole journal (`src/engines/pi/session-control.ts`) |

So an operator asking "what did last night's digest actually say?" has to open the `.jsonl` by hand.

## Two separate gaps

**1. `attach` never backfills an existing session.** `src/cli/commands/attach.ts` initialises the cursor to
`state.leafEntryId`, so the first round asks for `entries({ since: <current leaf> })` and gets nothing back — for a
schedule that already fired, the replay block is always empty and only new events are tailed. The code calls this
step "backfill (render the durable record since the cursor)", and the support already exists; only the
opening cursor stands in the way. A flag (`--from-start`, or making the first round pass no cursor) is the whole
change.

This one is arguably a bug in `attach`'s own stated behaviour, independent of schedules.

**2. There is no OFFLINE reader.** Both paths above need `sessionControl: true` (off by default) and a running
serve writing `<stateRoot>/control.json`. When the agent is stopped — which is the normal state when someone is
investigating last night's run — nothing can print a session. The minimum would be a read command that opens the
journal directly: `fastagent schedule history <name> --session`, or a more general `fastagent session show <id>`
that does not go through the control plane at all.

## Why this matters more after #549

The justification for deleting the stored reply was "it is already stored once, in the session". That is true of the
[...]

PR #556, opened 2026-09-17
refactor(cli)!: drop `fastagent attach`

kid7st:
`fastagent attach` is removed. The session control plane — the in-process hub, `/control/*`, `connectSessionControl` / `connectAgent` — is untouched, and so is every chat channel's `/stop`, which reaches the live run through that same hub.

## Why

attach never came from a user need. It entered in #258 ("Phase 3 — the HTTP+SSE transport"), whose stated purpose was *"one transport serves every remote consumer — Web panel, desktop app, CLI attach"*: it was the consumer that could be built inside this repo. Every later commit touching it followed a change in the plane (#309 navigate, #310 commands, #373 lifecycle, #542 the waitable subscription boundary) rather than anything an author or operator asked for.

What it does that nothing else does turned out to be narrow. Watching a local serve is what `dev`'s turn traces already print; stopping a runaway turn in a chat channel is that channel's own `/stop` (which needs no `sessionControl`); saying something into a Telegram conversation is Telegram. The residue was steering a session with no human channel, and remote intervention from a laptop — both behind a config flag that is off by default.

Reading a past turn, which used to be the other reason to reach for it, is now `fastagent chat --session <id>` and `fastagent schedule history` (#553).

## Why this costs no coverage of the transport

attach was not the plane's end-to-end proof, which was my assumption before checking. Of the 48 tests in `control-http.test.ts`, 6 were attach's own; the transport keeps its independent ones, including:

- `local and remote are isomorphic: capabilities/state/entries/dispatch answer identically`
- `describeSpecConformance("remote agent over /control/invoke")` — the SPEC suite over a real HTTP server
- fail-closed auth, CORS/preflight, heartbeats and the connect watchdog, seq-gap termination, SSE lifecycle

So the wire still has executable clients and a conformance suite; what goes is the interactive one.

## What a user loses

- No terminal client for a session with no human channel (a schedule's, a github issue's): no live view, no `/abort`, no typed steering.
- No `--url`/`--token` intervention against a deployed box from a laptop.

Both remain reachable over `/control/*` for anything that speaks the protocol — the Web panel and desktop app the plane was designed for. `<stateRoot>/control.json` stays, for a client that shares the filesystem.

## Removed

| | |
|---|---|
| `src/cli/commands/attach.ts` | 589 lines |
| its CommandSpec in `src/cli/program.ts` | 23 |
| `control-http.test.ts`: `activePathSlice`, `answerSlashInput`, `decideRound`, 4 × `attachRound` | 376 |
| `cli.test.ts`: the remote-attach placement case | 12 |
| `docs/cli.md`: the command section, plus references in `api-reference`, `configuration`, `troubleshooting`, `design/session-control` | |

Net −1037/+19.

## Verification

`npm run lint && npm run typecheck && npm test` — 1703 passed, 2 skipped (was 1718 with attach's 15).


kid7st (commented):
## Interactive review record

This records the curated outcome of an interactive `/review` session. It omits private model reasoning and the raw conversation.

<review-report>
## Review Scope

- Review scope: `git diff 41572e73e2edb308eab16356c08a49d8402c1a6b` (PR #556, HEAD `70d87bdb`, matching `origin/refactor/drop-attach`), 17 files total, +30/−1047. Includes deletion of `src/cli/commands/attach.ts`, the CommandSpec in `src/cli/program.ts`, 6 attach-specific cases in `test/control-http.test.ts`, one case in `test/cli.test.ts`, plus wording updates to `docs/cli.md`/`api-reference.md`/`configuration.md`/`troubleshooting.md`/`design/session-control.md`/`AGENTS.md` and several source comments.
- Working tree: `git status --porcelain` is empty, `git ls-files --others --exclude-standard` shows no untracked files, and there are no staged/unstaged changes, so this round reviews only the committed branch content.
- Verified myself (without relying on the previous round's conclusions): `npm run typecheck` passes; `npm run lint` (biome + check-doc-links + knip) passes; `npm test` 128 files / 1707 passed | 2 skipped. Re-checked the whole repo (including hidden directories such as `.github`, `scripts`) with `rg -i attach`; all remaining hits are unrelated semantics like attachment / `railway add` / `AttachRolePolicy`, with no leftovers pointing at the removed command. `handleControl`, `fakeUnreachable`, `placementOrExit`, `failStartup`, `connectAgent`, `commands()`, and `SessionEventStream.ready` all still have consumers and test coverage, so the deletion leaves no orphaned exports.
- Could not verify: manual end-to-end CLI behavior against a real running serve (without attach there is no terminal client to exercise `/control/*`), so I had to rely on the HTTP-server-level cases in `control-http.test.ts`.

## Constraints & Preferences

- The repository's maintenance policy explicitly does not consider backwards compatibility or migration, so "removing a command requires a migration path/deprecation period" is not treated as a fix item; breaking changes are only surfaced as a callout for human review.
- `<stateRoot>/control.json` and `announceControl` are product surface explicitly retained per the PR description (for discovery by desktop clients on a shared filesystem); they should not be removed later on the grounds of "no in-repo consumer".
- Documentation links are validated by `scripts/check-doc-links.mjs`; when changing anchors, that check must keep passing.

## Findings

### [P3] `docs/design/session-control.md:353` The only remaining example does not have the shape it is cited to demonstrate

The point of this passage is that "the two reads (`entries({ since })` + `state()`) must run concurrently with live iteration, not be sequenced before it." The original text used `src/cli/commands/attach.ts` as the primary evidence and the api-reference example as a secondary one; after attach was deleted, only the api-reference example remains, and in that code `await s1.entries({ since: cursor })` and `await s1.state()` appear after `await watching;`, i.e. the backfill happens only once live iteration has finished — exactly the ordering this text advises against. A reader who follows the pointer will find a counterexample. Either move those two reads inside the `watching` window of the api-reference example (even just adding a line of comment noting that they run concurrently with iteration), or drop the "has that shape" clause and keep only the recipe's four-step ordering description.

## Review Decisions

- Carrying over the previous round's judgment: the "attached" / "attach/detach" wording at `src/engines/pi/session-control.ts:157/416` is generic verbiage for subscriber connection, unrelated to the removed command; leave unchanged.
- Carrying over the previous round's judgment: the `REQUEST_TIMEOUT_MS` comment in `src/session-remote.ts` has been reworded to a neutral phrasing ("a client's reliability model counts failed rounds against the budget"). That constant's behavior is still pinned by the `no usable response in 10s` case in `control-http.test.ts`, so it is not raised again this round.
- Accepting the explicit decision in the PR description: keep the `control.json` write and `announceControl`; do not classify them as dead code merely because there is currently no in-repo consumer.
- Per the repository's maintenance policy, removing a CLI command does not require migration notes or a deprecation period; the breaking fact goes into a callout and is not counted as a finding.

Verdict: correct

## Human Reviewer Callouts (Non-Blocking)

- **This change introduces backwards-incompatible public schema/API/contract changes:** the `fastagent attach` subcommand is removed from the CLI surface (`src/cli/program.ts`, with the corresponding section deleted from `docs/cli.md`). Scripts depending on that command will exit with commander's unknown-command error; the `/control/*` protocol, `connectSessionControl` / `connectAgent`, and `control.json` are unchanged.
</review-report>

PR #560, opened 2026-09-17
fix: give the file-backed session store a file-backed lease

### LINKED ISSUE #554: the session lease does not reach across processes: `fire` beside `dev` silently branches the record
## What happens

`inProcessLease()` is a `Set` in one process (`src/engines/pi/turn-kit.ts:17`), while the records it protects are files on disk. So the single-writer rule — the one that answers a concurrent turn with `session_busy` — does not reach a second process opening the same record.

`docs/design/core.md` already states the consequence and names it as an implementation limit rather than a boundary:

> Two writers do not corrupt a session. pi appends one whole transaction per line, so the file stays parseable; the conversation *branches* (both writers' entries share a parent) and one branch falls off the path the next turn reads. Within one process this cannot happen — the session lease answers a concurrent turn with `session_busy` — so it is that rule not reaching across processes.

## The reproducible case

`fastagent fire <name>` invokes on `schedule:<name>` (`src/cli/commands/fire.ts:56`) — the same session id the scheduler inside a running `dev`/`start` fires. Run one by hand while the resident scheduler fires the same schedule and both processes append to one record, each from its own in-memory snapshot. Nothing errors; a branch is silently abandoned.

(`invoke` uses `randomUUID()` so it cannot collide, and `chat` writes pi's per-workspace location. `chat --session` opens a copy and never appends to a served record. So `fire` is the local collision that exists today, plus any topology with two serves over one state root.)

## Why the scope should be the record

`Lease` is already a port, injected at `src/engines/pi/create.ts:281` (`opts.lease ?? inProcessLease()`), and core.md lists a shared lease among what multiple instances require. The lease's persistence domain should match the store's: the in-memory store gets the in-process lease, the file-backed store gets a file-backed one.

What core.md rejects is a *directory-level* writer lock, because it "would forbid harmless topologies (two channels in two processes, a one-shot `invoke` beside a serving `dev`, an embedder mounting twice)". A record-scoped lease does not: two processes serving different sessions never touch the same lock.

## Sketch

`proper-lockfile` is already a dependency and already used for a cross-process lock with staleness reclaim (`src/engines/pi/auth.ts:81`, `stale: 30_000`), and `claimSlot` is the existing `O_EXCL` precedent for an atomic decision.

```ts
export function fileLease(sessionsDir: string): Lease {
  tryAcquire(session) {
    // lock <sessionsDir>/.locks/<encoded id>, NOT the record file: records are created lazily
    // (proper-lockfile needs the target to exist), and a lock file must not land in the directory
    // list() scans.
    try { return lockfile.lockSync(lockPath, { stale: 30_000, update: 10_000, onCompromised }); }
    catch (e) { if (e.code === "ELOCKED") return null; throw e; }
  }
}
```

Four points that decide whether it is correct:

- `ELOCKED` → `null` (which becomes `SessionBusy`, control flow). Every other error must throw: a read-only filesystem or a permission fault turning into "busy" is a silent degradation.
- The `update` heartbeat is what lets a lock be held for a whole model turn; a killed process's lock is reclaimed after `stale`.
- `tryAcquire` is synchronous and is exactly a no-retry probe, so `lockSync` fits the existing port signature — `acquireSessionLease` does not change.
- Verification needs no subprocess: `proper-lockfile` is mkdir-based, so a second `lockSync` on the same path in the same process also gets `ELOCKED`. Swapping the implementation back to `inProcessLease` must turn that test red.

## Known limit

mtime-based staleness is unreliable on network filesystems (clock skew, attribute caching), so this covers multiple processes on one machine — which is the case above. Multiple instances across machines still need the shared backend core.md already names.

kid7st:
Closes #554.

## Behavior

A session's single-writer rule now reaches across processes on one machine. `fastagent fire <name>` beside a running `dev`/`start` — both invoke `schedule:<name>` — gets `session_busy` instead of silently appending a second branch to the record. Same for any second serve over one state root.

The lease's persistence domain follows the store's: `piInMemorySessionRecordStore` keeps `inProcessLease()`, and `piSessionRecordStore` (the only one `createPiAgentFromDir` builds) gets `fileLease()`. `Lease` was already an injected port, so nothing above it changed.

## How

`fileLease({ dir, cwd })` locks `<sessionsDir>/.locks/<piSessionId(session)>` through `proper-lockfile`, already a dependency and already the cross-process lock in `auth.ts`.

- Record-scoped, never directory-scoped. Two processes serving *different* sessions over one state root never contend, so the topologies core.md deliberately allows (a one-shot `invoke` beside a serving `dev`, two channels in two processes, an embedder mounting twice) stay allowed.
- `ELOCKED` → `null` → `SessionBusy`, which is control flow. Every other error throws: a read-only filesystem or a permission fault reported as "busy" would be a silent degradation.
- The lock names a path rather than locking the record file: records are created lazily, and `realpath: false` is what lets proper-lockfile lock a name with no file behind it.
- `stale: 30_000` with proper-lockfile's default heartbeat lets one lock be held for a whole model turn while a killed process's lock is reclaimed.
- `lockSync` fits the existing synchronous, no-retry `tryAcquire` signature, so `acquireSessionLease` is untouched.

## Why a lock, when the writes are appends

Append-only prevents *corruption*, not *conflict*. pi writes one whole JSON line per entry, so two writers leave a parseable file with every line in it — but `parentId` is taken from the writer's in-memory leaf, so the two turns become siblings and the next turn's context walks back from one leaf only. The other turn is still on disk and no longer in the conversation. `test/session-store.test.ts` pins exactly that: 7 lines on disk, one whole turn missing from `buildContextEntries()`.

pi does not lock sessions because it never shares one file: `newSession()` mints `<timestamp>_<id>.jsonl` per run. fastagent maps a stable caller session id onto one record and re-opens it, which is where the shared-file case comes from. (pi does lock its auth, settings and trust files — all three shared by construction.)

Two secondary reasons the store is not purely append-only either: `SessionManager._rewriteFile()` truncates and rewrites on a version migration, and `_persist` flushes with `wx` before the first assistant message.

## Verification

proper-lockfile is `mkdir`-based, so a second `lockSync` on the same path in the same process is the second process's answer — the test needs no subprocess. Two mutations were run to prove the assertions can fail:

| Mutation | Result |
|---|---|
| `fileLease` → `inProcessLease()` | contention + fault-vs-busy tests red |
| lock directory → the records directory | fault test red |

A third case (a lock must not appear in `list()`) stayed green under both mutations — `recordFiles` filters on `.jsonl`, so placement is not what enforces it — and was dropped rather than kept as a tautology.

## Known limit

mtime-based staleness is unreliable on network filesystems, so this covers multiple processes on one machine. Multiple instances across machines still need the shared lease backend `docs/design/core.md` names; that section is updated to say which half is now closed.


kid7st:
Closing without merging. The justification did not survive checking.

`claimSlot` (`src/schedule/state.ts:235`, `O_EXCL` per schedule+slot) already prevents two resident schedulers over one state root from firing the same slot, so the multi-process schedule collision I claimed does not exist. What is left:

- **`fire` beside a resident scheduler.** Needs the cron instant to land inside a manual turn. `fire` exists to avoid waiting for that instant, so the overlap is thin.
- **Two parallel `fire`s of one schedule.** The operator asked for two turns on one conversation, in two terminals they are watching. A lost branch there is visible and self-inflicted, not silent.
- **A long-connection channel delivering one conversation to two processes.** Unverified that Feishu even does this, and nobody has hit it.

Against that, the lock adds a failure mode of its own: a SIGKILLed process leaves a lock that refuses turns on a healthy session until the 30s staleness window expires. Trading a rare silent failure for a rare visible one is not obviously a win at this frequency.

`docs/design/core.md` on `main` already states the limit accurately, so nothing there needs changing. The branch stays (`fix/cross-process-session-lease`) with a working `fileLease` and a test pinning the branch-drop, if this ever shows up in practice.

